PR #769, opened 2021-12-20
Remove 'windows' folder with giveio.sys driver

mariusgreuel:
I really need to do this, the giveio.sys driver is of no use any more.


dl8dtl:
I don't understand the changes to avrdude.texi.
Everything else is fine by me.

mariusgreuel:
> I don't understand the changes to avrdude.texi.
Anything specific? Or do you want to keep the documentation?

dl8dtl:
> > I don't understand the changes to avrdude.texi.
> 
> Anything specific? Or do you want to keep the documentation?

I was a bit confused about *what* changed.

Now, upon re-reading it, most is probably fine, but I wouldn't drop those "Credits". Even though their work is going to be removed, at some point in time, these people contributed something that was useful to the project.

mariusgreuel:
> but I wouldn't drop those "Credits"

Right. As the entire section on giveio.sys is removed, perhaps AUTHORS is a good place?

dl8dtl:
> > but I wouldn't drop those "Credits"
> 
> Right. As the entire section on giveio.sys is removed, perhaps AUTHORS is a good place?

Seems reasonable to me. We could add a section like "Contributors to code no longer present" or such.

PR #778, opened 2021-12-23
CMake: rework avrdude.conf.in

### LINKED ISSUE #776: CMake doesn't correctly handle conditionals in avrdude.conf.in
When running CMake, `avrdude.conf.in` is configured into a customized version `avrdude.conf`.
While replacing the `@DEFAULT_PAR_PORT@` and `@DEFAULT_SER_PORT@` macros works, handling conditionals like `@HAVE_PARPORT_BEGIN@` / `@HAVE_PARPORT_END@` doesn't work: all these sections are always included into the result.

Editing `ac_cfg.h` works according to the selection options.

yegorich:
Use #cmakedefine statements to include or exclude features in
avrdude.conf.in.

Fixes #776

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -1318,7 +1318,9 @@ programmer
   connection_type = usb;
 ;
 
-@HAVE_PARPORT_BEGIN@
+#cmakedefine HAVE_PARPORT
+#if defined( HAVE_PARPORT )
+
 # Parallel port programmers.
 
 programmer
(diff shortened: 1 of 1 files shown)


mariusgreuel:
Thank you for the PR. I think what you are proposing is not possible. AFAIK, CMake cannot interpret C pre-processor statements, so the code is not removed as it should.

Also, the ac script is still in use, so the avrdude.conf.in file has to be somewhat compatible with both cmake and ac.

BTW, I have a fix for this issue in the queue, doing a simple regex replace as the ac script does.

yegorich:
@mariusgreuel OK. I see. I'll close my PR.

Issue #770, opened 2021-12-21
Avrdude always needs the absolute path to avrdude.conf


kristofmulier:
Whenever one launches `avrdude`, it seems like the full absolute path to the `avrdude.conf` file must be given with the `-C` argument. I tried with a relative path (relative to the `avrdude` executable, something like `../etc/avrdude.conf`) but that doesn't work. Avrdude absolutely wants the *absolute* path.

In the makefile that we created for Embeetle IDE, we therefore end up with monstruous expressions, like:

```makefile
FLASHTOOL_PATH = $(if $(filter $(dir .),$(dir $(FLASHTOOL))),$(firstword $(wildcard $(patsubst %,%/$(FLASHTOOL),$(subst $(PATHSEP), ,$(PATH))))),$(FLASHTOOL))

.PHONY: flash
flash: $(BINARIES) print_flash
	"$(FLASHTOOL)" -C "$(subst /bin/../,/,$(subst /avrdude\$,,$(FLASHTOOL_PATH)\$)/../)etc/avrdude.conf" \
                 -v \
                 -patmega328p \
                 -carduino \
                 -P$(FLASH_PORT) \
                 -b115200 \
                 -D \
                 -Uflash:w:$(ELF_FILE:.elf=.hex):i
```

It would be so much easier if one could pass a *relative* path in the `-C` argument. Or even if the `-C` argument could be left out alltogether and `avrdude` looks itself for the file at `../etc/avrdude.conf`.

dl8dtl:
No, it's not necessarily an absolute path.

There is a system default directory where it is being searched for, ususally `${prefix}/etc/` (where prefix could be like `/usr`, or `/usr/local`). Only if it's not found there, the `-C` option is required. The location of this directory can be overridden at ./configure using the `--sysconf-dir` option. The idea behind all this is, for a self-consistent standard installation, there is never a reason to provide a `-C` option at all.

Deriving the location of the binary itself (in order to walk from there to the system configuration directory) is a non-trivial task. If the program was started by the surrounding shell by searching along `${PATH}`, `argv[0]` does not contain a directory component but just "avrdude" only. Under unixoid systems, it's not even guaranteed the binary itself still has a directory entry while it is running: if someone deleted the file right after starting it, the directory entry will be gone yet the physical file keeps existing as long as it is being executed (deletion will proceed once it exited).

mariusgreuel:
[...]

PR #780, opened 2021-12-24
Find 'avrdude.conf' based on absolute path to executable

kristofmulier:
<h2>1. General idea</h2>

This pull request is all about improving the way `avrdude.conf` is found. As suggested by @dl8dtl (see https://github.com/avrdudes/avrdude/issues/770), the file will now be searched in this order:

1. `<dirpath of executable>/../etc/avrdude.conf`<br>This is the most likely location. Every AVRDUDE installation I have ever seen puts the `avrdude.conf` file here.

2. `<dirpath of executable>/avrdude.conf`<br>Having `avrdude.conf` right next to the executable is somewhat unlikely (especially for Unix systems), but can sometimes happen on Windows.

3. `CONFIG_DIR/avrdude.conf`<br>If `avrdude.conf` was not found near the executable, the 'old' system kicks in and looks for the file in a location like `/usr/local/etc/avrdude.conf`. The code to do this is unaffected. I just moved it a bit and put it in an if-block, such that it only kicks in if nr. 1 and nr. 2 both failed to locate the file.

&nbsp;<br>
<h2>2. Executable location</h2>

Of course, this new procedure to look for `avrdude.conf` requires the location of the AVRDUDE executable in the first place. That's far from easy. I got it solved with the **whereami** library:
https://github.com/gpakosz/whereami/tree/master/src

I only needed to drop in these two files:

 - `whereami.c`
 - `whereami.h`

After that, I could use the `wai_getExecutablePath()` function to obtain the executable abspath, the length of that path and the length of its dirpath (the path to the directory of the executable). With all of this, it's a piece of cake to create the following local variables:

 - `executable_abspath` (eg. `/home/kristof/avrdude/src/avrdude`)
 - `executable_abspath_len` (eg. `33`)
 - `executable_dirpath` (eg. `/home/kristof/avrdude/src`)
 - `executable_dirpath_len` (eg. `25`)

I turn all backslashes into forward slashes. This way, it should all work on Windows too. Unfortunately, I can't test it on Windows myself. I currently work on Ubuntu and would love to cross-compile it here to create a Windows distribution of AVRDUDE. Is that possible?

Thanks a lot @dl8dtl and @mariusgreuel for your help!


dl8dtl:
Sorry, I should have written that first: thank you very much for your work! Much appreciated.

Overall comment: the documentation needs to be adjusted to reflect the new algorithm to find the system config file.
I'm sorry it is by now still in `texinfo` format which might look a bit archaic, but 20 years ago, it's been the tool to eventually get a nice PDF documentation. We might change that to a more modern system as soon as somebody volunteers to transfer the existing texinfo stuff over … (There's also a man page, `avrdude.1`, but if you provide the documentation in texinfo, I can easily copy & paste it over into the man page.)

kristofmulier:
Hi @dl8dtl ,
Thank you for your help!

I added a few more commits to my pull request to achieve the following:

1. **Clean coding style:**<br>I put spaces around operators (eg. `a - 1` instead of `a-1`) and spaces after keywords like `if`, `while` and so forth.

2. **Update docs:**<br>I updated the documentation file `src/doc/avrdude.texi` to reflect the new way AVRDUDE searches the system config file `avrdude.conf`. The effects of my modifications are visible on pages 11 and 41 in `avrdude.pdf`.

Kind regards,
Kristof

Issue #481, opened 2021-12-11
[bug #54603] Error in `avrdude': corrupted size vs. prev_size


avrdude-admin:
Max <max@winsoft.hu>
Fri 31 Aug 2018 12:33:21 PM UTC
Programmer hardware: USBasp
Device type: t10

After programming from Pi3 via USBasp to an ATTiny10 with avrdude6.3, this error occurs:

- - - Error in `avrdude': corrupted size vs. prev_size: 0x00dbc688 ***Aborted

Tried the avrdude package from Pi, and also make/compile/install the avrdude6.3 latest sources. Same issue.

Full details:

pi@raspberrypi:~ $ cat /proc/meminfo|grep Mem
MemTotal:         766744 kB
MemFree:          670376 kB
MemAvailable:     677816 kB
pi@raspberrypi:~ $ tail /boot/config.txt -n1
gpu_mem=256
pi@raspberrypi:~ $ sudo avrdude -p t10 -c usbasp -v -v -v -v -F

avrdude: Version 6.3, compiled on Aug 31 2018 at 11:58:40
Copyright (c) 2000-2005 Brian Dean, [http://www.bdmicro.com/](https://savannah.nongnu.orghttp://www.bdmicro.com/)
Copyright (c) 2007-2014 Joerg Wunsch

System wide configuration file is "/usr/local/etc/avrdude.conf"
User configuration file is "/root/.avrduderc"
User configuration file does not exist or is not a regular file, skipping

Using Port                    : usb
Using Programmer              : usbasp
avrdude: usbasp_open("usb")
avrdude: seen device from vendor ->www.fischl.de<-
avrdude: seen product ->USBasp<-
AVR Part                      : ATtiny10
Chip Erase delay              : 0 us
PAGEL                         : P00
BS2                           : P00
RESET disposition             : dedicated
RETRY pulse                   : SCK
serial program mode           : yes
parallel program mode         : yes
Timeout                       : 0
StabDelay                     : 0
CmdexeDelay                   : 0
SyncLoops                     : 0
ByteDelay                     : 0
PollIndex                     : 0
PollValue                     : 0x00
Memory Detail                 :

Block Poll               Page                       Polled
Memory Type Mode Delay Size  Indx Paged  Size   Size #Pages MinW  MaxW   ReadBack
----------- ---- ----- ----- ---- ------ ------ ---- ------ ----- ----- ---------
signature      0     0     0    0 no          3   16      0     0     0 0x00 0x00
[...]

PR #784, opened 2021-12-26
Fix avr_read() for page reads

mariusgreuel:
Here comes a bugfix for #481. There is a buffer overrun because avr_read() attempts to do page reads for the signature buffer (which only has a size of 3). We may want to round up every memory buffer allocation to the next page size multiple so that page reads can be used, but that is another issue.


dl8dtl:
> We may want to round up every memory buffer allocation to the next page size multiple

The problem is, there is no notion of a per-device uniform "page size", and most signature memories are usually handled as one-byte pages.

PR #786, opened 2021-12-28
Add support for Micronucleus bootloader

mariusgreuel:
This PR adds support for the [Micronucleus bootloader V1 and V2](https://github.com/micronucleus/micronucleus), so you do no longer need the Micronucleus command-line utility.

This bootloader is typically used on small ATtiny boards, such as Digispark (ATtiny85), Digispark Pro (ATtiny167), and the respective clones. By default, it bootloader uses the VID/PID 16d0:0753 (MCS Digistump).

As the micronucleus bootloader is optimized for size, it implements writing to flash memory only. Since it does not support reading, use the **-V** option to prevent avrdude from verifing the flash memory. To have AVRDUDE wait for the device to be connected, use the extended option **-x wait**.

Example:
```
avrdude -c micronucleus -p t85 -x wait -V -U flash:w:main.hex
```

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -467,6 +467,8 @@ add_library(libavrdude STATIC
     linuxspi.h
     linux_ppdev.h
     lists.c
+    micronucleus.c
+    micronucleus.h
     my_ddk_hidsdi.h
     par.c
     par.h
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -136,6 +136,8 @@ libavrdude_a_SOURCES = \
 	linuxspi.h \
 	linux_ppdev.h \
 	lists.c \
+	micronucleus.c \
+	micronucleus.h \
 	my_ddk_hidsdi.h \
 	par.c \
 	par.h \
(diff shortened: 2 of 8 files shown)


MCUdude:
BTW I'm getting this warning when compiling Avrdude on my mac:

```
micronucleus.c:398:9: warning: passing 'uint8_t *' (aka 'unsigned char *') to parameter of type
      'char *' converts between pointers to integer types with different sign [-Wpointer-sign]
        buffer, size,
        ^~~~~~
```

dl8dtl:
For very historical reasons, it seems `usb_control_msg()` is declared to take a `char *` argument for the buffer.
The warning is benign, but could be silenced by a typecast if we want.

mariusgreuel:
> BTW I'm getting this warning when compiling Avrdude on my mac:

I am embarrassed. That was the only -Wall warning. Fixed it.

Perhaps we should enable -Wall by default.

dl8dtl:
> Perhaps we should enable -Wall by default.

I don't mind. Next step: -Wextra ;-)

mariusgreuel:
> > Perhaps we should enable -Wall by default.
> 
> I don't mind. Next step: -Wextra ;-)

Added. There are a few warnings for linux/i686, but otherwise acceptable. -Wextra is a little noisy yet.

PR #791, opened 2022-01-01
Add PicKit4 and SNAP programmers

MCUdude:
The Pickit4 and SNAP programmers can also be used as PDI and ISP programmers. This PR adds ISP and PDI support to the Pickit4 and SNAP programmers.

I can confirm that the ISP and PDI options on both programmers work; I've tested it on actual hardware.


MCUdude:
BTW does anyone know of a simpler way to get the Pickit4 and SNAP into AVR mode without having to do it through MPLAB?

dl8dtl:
> BTW does anyone know of a simpler way to get the Pickit4 and SNAP into AVR mode without having to do it through MPLAB?

Well, you can either pester Microchip to release their communications protocol documentation (as Atmel did), or you're going to sniff the USB traffic between the MPLAB and Pickit.

PR #792, opened 2022-01-02
avrftdi: don't use the deprecated ftdi_usb_purge_buffers routine

yegorich:
Since commit ed46f09c1ccd1351e003a200ba50e3e4778ac478 (Implement
tc[io]flush methods & deprecate broken purge_buffers methods.)
`ftdi_usb_purge_buffers()` routine is deprecated. Use `HAVE_FTDI_TCIOFLUSH`
macro to invoke the newly introduced `ftdi_tcioflush()` routine.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -36,6 +36,7 @@ option(DEBUG_CMAKE "Enable debugging output for this CMake project" OFF)
 
 include(CheckIncludeFile)
 include(CheckFunctionExists)
+include(CheckSymbolExists)
 include(GNUInstallDirs)
 
 set(CONFIG_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}")
@@ -177,6 +178,8 @@ find_library(HAVE_LIBFTDI1 NAMES ${PREFERRED_LIBFTDI1})
 if(HAVE_LIBFTDI1)
     set(LIB_LIBFTDI1 ${HAVE_LIBFTDI1})
     set(HAVE_LIBFTDI_TYPE_232H 1)
+    set(CMAKE_REQUIRED_LIBRARIES ${LIB_LIBFTDI1} ${LIB_LIBUSB} ${LIB_LIBUSB_1_0})
+    check_symbol_exists(ftdi_tcioflush "libftdi1/ftdi.h" HAVE_FTDI_TCIOFLUSH)
 endif()
 
 # -------------------------------------
--- a/src/avrftdi.c
+++ b/src/avrftdi.c
@@ -720,7 +720,11 @@ static int avrftdi_open(PROGRAMMER * pgm, char *port)
 	/* set SPI mode */
 	E(ftdi_set_bitmode(pdata->ftdic, 0, BITMODE_RESET) < 0, pdata->ftdic);
 	E(ftdi_set_bitmode(pdata->ftdic, pdata->pin_direction & 0xff, BITMODE_MPSSE) < 0, pdata->ftdic);
+#ifdef HAVE_FTDI_TCIOFLUSH
+	E(ftdi_tcioflush(pdata->ftdic), pdata->ftdic);
+#else
 	E(ftdi_usb_purge_buffers(pdata->ftdic), pdata->ftdic);
+#endif
 
 	write_flush(pdata);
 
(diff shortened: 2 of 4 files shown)


yegorich:
The first `libftdi1` version implementing `ftdi_tcioflush()` is 1.5.0. According to the GitHub actions, only MacOS provides this version.

This PR is compile-tested only.

mariusgreuel:
Your patch does not seem to work for MinGW:
https://github.com/avrdudes/avrdude/runs/4685849550?check_suite_focus=true

To be honest, I would favor a `#define _FTDI_DISABLE_DEPRECATED 1` in avrftdi.c.

P.S.: I am not really a Un*x guy, but I am pretty sure the T in TCIOFLUSH stands for terminal. Not sure why they picked such a name.

dl8dtl:
> I am pretty sure the T in TCIOFLUSH stands for terminal.

Correct.

> Not sure why they picked such a name.

TC = "terminal control"

This is the historic name passed as a constant to an `ioctl()` syscall.

(As these ioctls have never been agreed on between differen Unix flavours, Posix instead standardized a set of library functions, so the respective function is `tcflush()`.)

yegorich:
According to https://cmake.org/cmake/help/latest/module/CheckFunctionExists.html, it is better to use `CheckSymbolExists`. I'll resend the new version and we will see whether it works for mingw.

yegorich:
`CheckSymbolExists` doesn't help. @mariusgreuel can you look at the `CMakeError.log`? Why does it have problems with detecting the routine?

dl8dtl:
Btw., so far, the idea is that we maintain both, CMake and autoconf/automake in parallel.
Right now, the PR only handles CMake though.

yegorich:
So, mingw issue is fixed. As we use static libs, I had to also add `libusb` dependencies.

@dl8dtl now I can try to also integrate this check into `configure.ac`.

dl8dtl:
Btw., I can test that stuff once it's ready.

yegorich:
@dl8dtl autoconf seems to work. Can you look at it? Thanks

dl8dtl:
> @dl8dtl autoconf seems to work. Can you look at it? Thanks

For reasons I don't quite understand yet, it fails here (on FreeBSD). I get link errors since several libraries are no longer being linked.

Code from your repository, main branch links like:
```
libtool: link: gcc -Wall -Wno-pointer-sign -g -O2 -o avrdude avrdude-main.o avrdude-whereami.o avrdude-term.o  ./libavrdude.a -lusb -lftdi1 -lelf -lpthread -lm -lhidapi -lreadline -lncurses -ltermcap
```

Code from your repository, ftdi-fix-deprecated-call branch links like:
```
libtool: link: gcc -Wall -Wno-pointer-sign -g -O2 -o avrdude avrdude-main.o avrdude-term.o  ./libavrdude.a -lusb -lelf -lpthread -lm -lftdi1
```

As you can see, a lot of libraries are missing.

Will look into that tonight.

dl8dtl:
> Will look into that tonight.

Found it. You clobbered LIBS, rather than extending it.

yegorich:
Yes, now I see it too.

yegorich:
It's a pity that GitHub doesn't support FreeBSD in the Actions. Whether this [project](https://github.com/marketplace/actions/freebsd-vm) could  help?

dl8dtl:
> It's a pity that GitHub doesn't support FreeBSD in the Actions. Whether this [project](https://github.com/marketplace/actions/freebsd-vm) could help?

I'll let @mariusgreuel answer that, he's more knowledgable than me about the Github infrastructure.

dl8dtl:
Just as a confirmation, seems to work fine here.

mariusgreuel:
> > It's a pity that GitHub doesn't support FreeBSD in the Actions. Whether this [project](https://github.com/marketplace/actions/freebsd-vm) could help?
> 
> I'll let @mariusgreuel answer that, he's more knowledgable than me about the Github infrastructure.

I tried the FreeBSD action, with mixed success: As it runs in a VM, it is terribly slow, which is annoying when tinkering around. It also got stuck multiple times (which may be caused by GitHub actions which are also not perfect). For now, I have it running in my personal repo as a separate workflow, so that it does not mess up my status badge. I will monitor it to see how it performs over time.

dl8dtl:
Well, as I use FreeBSD as my primary development platform, I think that gives it enough testing coverage.

PR #793, opened 2022-01-03
Use yacc/byacc as an alternative to bison, closes #785

### LINKED ISSUE #785: CMake: Recognize more than just bison
Right now, the CMake project files require `bison` as a parser generator.
However, the grammar of `avrdude.conf` is generic enough to also work with the standard `yacc` or `byacc` (Berkeley yacc) generators.
In order for CMake to become a fully functional replacement of for autoconf/automake, this should be fixed.

mariusgreuel:


dl8dtl:
Cool, thanks!

PR #794, opened 2022-01-03
Derive program version string from last commit

mariusgreuel:
This PR automatically sets the version code based on the last commit, i.e. the date and hash is added to the version string. For instance, when the CMake project version is set to "6.99", the displayed version code will be "6.99-20201231 (0123abcdef)".

When the commit is tagged, which is assumed to be a release, the date and hash is omitted, so the displayed version code will be just "6.99".


dl8dtl:
Nice ;-)

kristofmulier:
Hi @mariusgreuel ,
Why would you omit the date when building a tagged commit?

The reason I'm asking is because Embeetle IDE parses the version output and then displays that in the **Toolbox** tab. For example:

image

What you see in the screenshot above results from parsing this string:

```bash
avrdude version 6.3-20210108, URL: <http://savannah.nongnu.org/projects/avrdude/>
```

Embeetle first derives the version nr, which is `6.3`. Because AVRDUDE had in the past a three digit version once, we add a trailing `.0` digit if there is no third digit present. Hence we obtain the version nr: `6.3.0`

Then Embeetle derives the date from the next number `20210108`, which is (I suppose) written as `yyyy-mm-dd`.

It would be a pity if the date is omitted.

mariusgreuel:
> Why would you omit the date when building a tagged commit?
Because it was never included in the releases. The CI builds version strings differ from the release builds.

PR #796, opened 2022-01-04
buspirate: fix -Wtautological-constant-out-of-range-compare

yegorich:
When compiling with clang 11.0.1 the following warning appears:

buspirate.c:198:14: warning: result of comparison of constant -1 with expression of type 'char' is always false [-Wtautological-constant-out-of-range-compare]
                if (*buf_p == EOF) {
                    ~~~~~~ ^  ~~~
Cast `char` to `int` to match the EOF type.

### DIFF
--- a/src/buspirate.c
+++ b/src/buspirate.c
@@ -195,7 +195,7 @@ static char *buspirate_readline_noexit(struct programmer_t *pgm, char *buf, size
 			continue;
 		if (*buf_p == '\n')
 			break;
-		if (*buf_p == EOF) {
+		if ((int)*buf_p == EOF) {
 			*buf_p = '\0';
 			break;
 		}


yegorich:
This PR is compile-tested only.

davidsainty:
I don't know this bit of code, but this change looks wrong.  *buf_p is a char, so it has already lost the distinction between (char)-1 and (int)-1 that was available in the value returned from buspirate_getc().  The int from buspirate_getc() should be stored into a temporary int and compared to EOF before it is converted into a byte value and the distinction is lost.

dl8dtl:
Basically, that entire code is fishy.
`buspirate_getc()` returns an `int` (which possibly could be `EOF`), but it then gets assigned to a `char*`, in order to afterwards compare it (among others) against `EOF` again.
The value returned from `buspirate_getc()` ought to be stored (in parallel to `*buf_p`) in an `int` local variable, and that one then used for the comparison. That will also fix the warning.

mariusgreuel:
@dl8dtl: We should think about how we want to handle warnings. While I personally have a zero warning strategy, I would rather not fix single warnings via PRs. We can easily produce hundreds of warnings if we crank up the warning level or switch compilers. I started off with a CMake project using -Wall -Wextra -pedantic, and turned that off real quick. Same for clang and MSVC, depending on the warning level.

mariusgreuel:
@davidsainty: I agree, the PR makes it worse, as it only removes the warnings - it does not fix the underlying problem flagged by clang.

dl8dtl:
> While I personally have a zero warning strategy, I would rather not fix single warnings via PRs.

Well, warnings have been fixed via issues (bugs or patches) in the past as well.
The pickyness of compilers improved though over time, so we've got more of those warnings again. ;-)

It certainly makes sense to fix in particular those warnings that hide programming errors, like in this case here. First, `EOF` is truncated to `\xFF` (thus hiding that possible char), and then, depending on the signedness of the default char, it's quite possible the comparison agains `EOF` will never become true. So this is a genuine bug, and should be fixed.

Other warnings the compilers recently came up with might not be that important to fix.

dl8dtl:
I just pushed a commit that implements the codeflow sketched above.

PR #798, opened 2022-01-04
Add MSVC builds and better WinUSB/FTDI support

mariusgreuel:
This PR adds native MSVC builds to AVRDUDE, as well as better support for WinUSB and native Windows FTDI drivers.

For various reasons, some of the existing AVRDUDE code is incompatible with MSVC. I added a compatibility shim to hide some issues, and made some changes to other C files, for instance where GCC extensions are used. I tried to make as few changes to the existing code.

Due to the lack of proper vcpkg or nuget packages, I added the required libraries as Git repos. Libusb and libfdti are rewritten to work around the libusb issues on Windows, but there are a few changes to libelf and libhidapi, too, such as adding support for CMake. The use of the GitHub libraries is enabled via the option USE_EXTERNAL=1.

I will submit a separate PR later to clean up some Windows related issues.

Feel free to suggest any changes you may have.


mariusgreuel:
I attached another clean-up commit. Its mostly Windows legacy code, the weird WIN32NATIVE define, but also a little < C99 clean-up: As the code-base does not compile without <stdint.h>, <stdlib.h>, etc., I removed detection for the following #defines.

I am not entirely sure about getaddrinfo(), but the build-matrix succeeds for all of them.

/* Define to 1 if you have the <stdint.h> header file. */
#cmakedefine HAVE_STDINT_H 1

/* Define to 1 if you have the <stdlib.h> header file. */
#cmakedefine HAVE_STDLIB_H 1

/* Define to 1 if you have the <inttypes.h> header file. */
#cmakedefine HAVE_INTTYPES_H 1

/* Define to 1 if you have the `usleep' function. */
#cmakedefine HAVE_USLEEP 1

/* Define to 1 if you have the `getaddrinfo' function. */
#cmakedefine HAVE_GETADDRINFO 1

/* Define to 1 if you have the `gettimeofday' function. */
#cmakedefine HAVE_GETTIMEOFDAY 1

/* Define to 1 if you have the `strcasecmp' function. */
#cmakedefine HAVE_STRCASECMP 1

dl8dtl:
> getaddrinfo

```
STANDARDS
     The getaddrinfo() function is defined by the IEEE Std 1003.1-2004
     (“POSIX.1”) specification and documented in RFC 3493, “Basic Socket
     Interface Extensions for IPv6”.
```
So I'd say, if MSVC can handle it, it's safe to assume it's just there.
We started relying in C99+ some time ago, with <stdint.h> and all its types, so removing these checks is fine. Again, MSVC is the late-follower for C standard versions anyway, so everything that works there is OK.

PR #801, opened 2022-01-04
Conffile clean up

dl8dtl:
Cleanup the config file handling a bit.


kristofmulier (approved):
Thanks a lot @dl8dtl :-)

PR #805, opened 2022-01-05
serbb_win32: fix pointer-to-int-cast warning

yegorich:
`HANDLE` is defined as `void*` hence it should be printed as a pointer
i.e. `%p`.


mariusgreuel:
This problem will be fixed in #798.

PR #809, opened 2022-01-07
Add missing ATmega__5 and ATmega__9 targets to avrdude.conf

### LINKED ISSUE #110: [bug #21499] Add support for the ATmega165* family
Joerg Wunsch <None>
Fri 02 Nov 2007 07:37:53 AM UTC

The ATmega165* family is currently not supported by avrdude.
Mostly same as ATmega169* except of the signatures.

[file #23834: atmega165a_and_derivative.conf](https://savannah.nongnu.org/bugs/download.php?file_id=23834)
[file #23845: atmega165a_corrected.conf](https://savannah.nongnu.org/bugs/download.php?file_id=23845)

This issue was migrated from https://savannah.nongnu.org/bugs/?21499

MCUdude:
It looks like a few ATmega targets are missing:

* ATmega164A
* ATmega165A
* ATmega165P
* ATmega169P
* ATmega325P
* ATmega3250P
* ATmega645P
* ATmega6450P
* ATmega6490P

Closes #110

There are other issues as well that should be considered too.
* Formatting is all over the place, and targets are not placed in a chronological order
* A few device names shouldn't even be there. For instance, there is no such thing as an ATmega169. There only exists ATmega169A's, ATmega169P's and ATmega169PA's.
* Parts with different suffixes are not consistent. There exist an ATmega8A target (identical to ATmega8, even has the same signature), but there doesn't exist an ATmega32A target.

I'm not sure I'd have patience enough to fix the formatting, but would you approve a PR if I added support for _all_ suffixes (A/P/PA/PB/V/PB) and structured the parts chronologically.

If yes, would you like to structure the parts based on 
* Family (ATmega48,88,168,328)
* Their leading number(s) (ATmega103,128,128RFA1,1281,1284,16,162,163,164,165 etc.)
* size of their number(s) (ATmega8,16,32,64,103,128,128RFA1, 162, 163,164,165 etc.)
?


dl8dtl:
STK502
Sure, there has been an ATmega169 once. :-)
Basically, all those "A" devices are die-shrink versions of their former non-A counterparts; they are normally supposed to be drop-in replacements for the older chips, so you could alias them in `avrdude.conf`. Alas, for a few parts, Atmel decided to add (or modify) some features during the transition, so for those, they also changed the signature.
I agree some structure would be useful, the file went way beyond what has once been anticipated. Structuring by family might be useful (and has been done for quite some recent additions). However, reorganizing that file is going to be a tremendous amount of work, with a high risk potential. In theory, from a QA point of view, afterwards you'd need to verify each and every single MCU to see whether you didnd't confuse something.
It would probably be possible to read and parse the file using the grammar parser, and then dump out a new one from the in-memory contents, but then, you'd completely lose any common ancestors and such, as well as all comments.

dl8dtl:
Basically, it awaits fixing of that signature, yes.

screenshot-2022-01-12T213051

Just copied over the signature table from the (new) datasheet. It's a strange back and forth with the last digit … the ATmega169PA uses the same signature as the original ATmega169, while the ATmega169A got an updated number in the signature row, but still the old JTAG part number – the signature row is a fuse cell that can be programmed during chip test, but the JTAG ID is hard-wired in the metal mask. For the larger devices though, each "A" device uses the same signature as their historic counterparts (0x03 / 0x04 in the end, matching the JTAG ID), but the P/PA counterparts got the new signature byte.

A mess. Please cross-check that the resulting `avrdude.conf.in` has all the signatures that way.

While you are at it, I just noticed the historic ATmega165 is also missing. It got 0x1e 0x94 0x07, but is otherwise identical to the ATmega165A.
screenshot-2022-01-12T213943

MCUdude:
Thanks for looking into it.

All signatures seem to be correct, I forgot to add the `atmega649p` target.
While we're at it, do you want me to add the corresponding *A and *PA as well, similar to what's been done to ATmega8?

```
part parent "m8"
    id               = "m8a";
    desc             = "ATmega8A";
    ;
```

If yes, the complete list for these two families will be:
* ATmega165,A,P,PA
* ATmega325,A,P,PA
* ATmega645,A,P,PA
* ATmega169,A,P,PA
* ATmega329,A,P,PA
* ATmega3290,A,P,PA
* ATmega649,A,P,PA
* ATmega6490,A,P,PA

MCUdude:
BTW, does `ocdrev` really has to be included on parts that has a "parent" where `ocdrev` is already defined?

```
part parent "m328"
    id			= "m328p";
    desc		= "ATmega328P";
    signature		= 0x1e 0x95 0x0F;

    ocdrev              = 1; # Do we really need this to be here?
;
```

dl8dtl:
All this is confusing.

I'm trying a table, including the 3rd signature byte (all bytes in hex):
```
Device      Signature <>   A    P    PA
------------------------------------------
ATmega165   1E 94     07   10   07   07
ATmega325   1E 95     X    05   X    0D
ATmega3250  1E 95     X    06   X    0E
ATmega645   1E 96     X    05   0D   X
ATmega6450  1E 96     X    06   0E   X
ATmega169   1E 94     05   11   X    05
ATmega329   1E 95     03   03   X    0B
ATmega3290  1E 95     04   04   X    0C
ATmega649   1E 96     03   03   0B   X
ATmega6490  1E 96     04   04   0C   X
```
Terrible, isn't it?
Of course, I'm not fully sure devices marked "X" really do not exist, but at least, that's what the datasheets say I have found (or the old ones I've still got around locally).

dl8dtl:
> BTW, does `ocdrev` really has to be included on parts that has a "parent" where `ocdrev` is already defined?

No, I think it can be inherited.

MCUdude:
After searching on Octopart.com, it looks like ATmega645PA and ATmega6450PA don't exist. All the others do.

Do you want me to add all parts that actually exist to avrdude.conf.in (excluding the *V suffix)?

dl8dtl:
> Do you want me to add all parts that actually exist to avrdude.conf.in (excluding the *V suffix)?

Yes, please.

MCUdude:
>> Do you want me to add all parts that actually exist to avrdude.conf.in (excluding the *V suffix)?

> Yes, please.

Done, see the latest commit.

This is something I'd like to do on more AVR families too. Maybe one PR per family to keep the commit history nice and traceable.

So what would be the rules to follow for a target that probably doesn't have a "non-suffix name", like the ATmega164? Should we still provide an "ATmega164" target even though this exact part probably doesn't exist, or should we "only" provide for the A,P, and PA parts?

dl8dtl:
> Should we still provide an "ATmega164" target even though this exact part probably doesn't exist

It makes sense to only provide entries for parts that do exist. For the ATmega*4 line, I actually do have an "ATmega644/V" datasheet, so this part did exist at least in form of engineering samples. The different size counterparts (164, 324, 1284) do only exist with at least a "P" suffix. They evolved around the time when Atmel developed that "Picopower" technology, so the 644 is the only one predating that.

MCUdude:
>  The different size counterparts (164, 324, 1284) do only exist with at least a "P" suffix.

Sorry, but this is not correct. I personally own an ATmega164A, an Amega324A and an ATmega1284 (non-A doesn't exist, only 1284 and 1284P). It's very confusing...

dl8dtl:
> I personally own an ATmega164A, an Amega324A and an ATmega1284 (non-A doesn't exist, only 1284 and 1284P)

That's what I meant with "at least".
The "P" (picopower) devices were first, and all parts except the ATmega644 started with the "P" suffix.
The "A" suffix denotes a die-shrink device, they came much later. In some cases, they have been named "PA", in others just "A".
I've never seen an ATmega1284 without any suffix; my very first one (date code 0733) is already an ATmega1284P.

MCUdude:
> I've never seen an ATmega1284 without any suffix; my very first one (date code 0733) is already an ATmega1284P.

I'm sure they are rare. But I promise they do exist!

BTW in avrdude.conf, what is the `reset` sting used for, other than printing it out in verbose mode?
```
reset            = dedicated;
...
reset            = io;
```
I'll assume `dedicated` means dedicated reset pin, while `io` means multiplexed with an io. If my assumptions are correct, there _alot_ of parts that have `dedicated` instead of `io`.

dl8dtl:
> > I've never seen an ATmega1284 without any suffix; my very first one (date code 0733) is already an ATmega1284P.

> I'm sure they are rare. But I promise they do exist!

I researched a little more, yes, they exist.

Historically, the device has been developed without the "P", but during development, they got the picopower feature added, so all production silicon of that time had the "P" appended.

It seems, later on, they added a non-P version, very likely just as a kind of marketing campaign (maybe they got a large customer not wanting to pay for "P"). You can see in the datasheet, the (hardwired) JTAG ID still ends up in 0x05, but the signature has been bumped to 0x06. (I wouldn't be surprised if that were the only actual difference between both devices.)

So yes, we need both these then.

> BTW in avrdude.conf, what is the `reset` sting used for, other than printing it out in verbose mode?

It's used in the STK500v1 code. Have a look at appnote AVR061 (it's in the `atmel-docs` subdirectory), it describes whether the device has a `RSTDSBL` fuse. I have no clue what the old STK500 fimware might have done with that information. :-/

As STK500v1 is mostly dead anyway, to the best of my knowledge, it is only used in a few bootloaders anymore, which probably ignore that parameter anyway. But OK, if you think we have entries for devices which are marked `io` but actually don't have a `RSTDSBL` fuse, we might fix that to be technically correct.

MCUdude:
> As STK500v1 is mostly dead anyway, to the best of my knowledge, it is only used in a few bootloaders anymore, which probably ignore that parameter anyway. But OK, if you think we have entries for devices which are marked io but actually don't have a RSTDSBL fuse, we might fix that to be technically correct.

I'll go through the file soon and fix it so everything is _technically_ correct too.

So is this PR ready to be merged?

dl8dtl:
> So is this PR ready to be merged?

I thought you want to update it for the other missing parts?
The wrong signature for the (non-A) ATmega169 needs to be fixed before merging it.

MCUdude:
> I thought you want to update it for the other missing parts?

Yes, of course. Sorry, but I'm so used to that the A and the "non-A" parts share the same device signature. I'll fix it.

> I thought you want to update it for the other missing parts?

I'd love to add the test as well. There aren't many left really. It's the entire ATmega__4, the ATmega__8 series and the ATmega16/32/64/128. And a few ATtinys too. I'll continue as soon as the kids are in bed!

MCUdude:
I've fixed the incorrect ATmega169 signature.

I'll add the rest of the missing targets in a new PR, to prevent this from being way too big.
Ready to be merged then!

dl8dtl:
> I've fixed the incorrect ATmega169 signature.

Though now, the '169A is wrong … (it needs to get 0x11).
I'm merging it anyway, and we can fix that afterwards.

dl8dtl:
> we can fix that afterwards.

I fixed it in the same commit as the `NEWS` update.

PR #817, opened 2022-01-09
Print expected part

### LINKED ISSUE #812: ATmega328P reported as lgt8fx328p
When connecting to popular boards such as the Arduino Uno, avrdude 6.4 reports an ATmega328P signature as "lgt8fx328p":

```
avrdude.exe -c arduino -p m328p -P com5
...
avrdude.exe: Device signature = 0x1e950f (probably lgt8fx328p)
```

I think it would be nice if avrdude was to print the _expected_ part.

Further, when providing the wrong signature, I think it would be nice if avrdude was to print the _most likely_ part.

```
avrdude.exe -c arduino -p m328 -P com5
...
avrdude.exe: Device signature = 0x1e950f (probably lgt8fx328p)
avrdude.exe: Expected signature for ATmega328 is 1E 95 14
```

I guess all this would not bother me that much if it weren't for the popular ATmega328P part.

I suggest two things:
- If the correct signature is specified, print the part that was specified, and not the one from the lookup table.
- For parts with ambiguous signatures, mark the more popular part as "preferred" and print that one.

mariusgreuel:
Fixes #812

### DIFF
--- a/src/config_gram.y
+++ b/src/config_gram.y
@@ -294,7 +294,7 @@ prog_def :
         lrmv_d(programmers, existing_prog);
         pgm_free(existing_prog);
       }
-      PUSH(programmers, current_prog);
+      LISTADD(programmers, current_prog);
 //      pgm_fill_old_pins(current_prog); // TODO to be removed if old pin data no longer needed
 //      pgm_display_generic(current_prog, id);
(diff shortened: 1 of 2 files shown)


dl8dtl:
Looks good to me.

PR #818, opened 2022-01-09
pindefs: conform to the function declaration

yegorich:
The last parameter in the `pins_check()` routine is declared as
`const bool`. Add the missing `const` specifier.


yegorich:
Original warning from Visual Studio 2019:

`avrdude\src\pindefs.c(220,132): warning C4028: formal parameter 4 different from declaration`

PR #826, opened 2022-01-11
CMake: enable dynamic-link library for libavrdude

yegorich:
Add on option `BUILD_SHARED_LIBS` to also build a DLL variant
of the `libavrdude` library. Turn it off by default to preserve
current behavior.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -586,6 +586,8 @@ add_library(libavrdude
     ${BISON_Parser_OUTPUTS}
     )
     
+set_target_properties(libavrdude PROPERTIES PREFIX "")
+
 target_include_directories(libavrdude
     PUBLIC
     "${PROJECT_SOURCE_DIR}"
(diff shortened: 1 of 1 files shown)


dl8dtl:
Speaking of that, I noticed that the static library built is named `liblibavrdude.a`.
There seems to be one "lib" too much. ;-)
Could we perhaps fix that at the same time?

yegorich:
This could be a problem. CMake doesn't like the fact that two targets have the same name. I'd suggest to rename the library into, for example, either `avrdudesdk` or `avrdudecore`. But this is a matter of another PR.

dl8dtl:
Strange behaviour of CMake. As they append the "lib" theirselves, they should also consider that a different target.

Actually, I just verified the FreeBSD port, we did ship `libavrdude.a` and `libavrdude.so` already in the past (with the automake system).

Googled a bit around,
`SET_TARGET_PROPERTIES(libavrdude PROPERTIES PREFIX "")`
should do the trick.

yegorich:
It really does. Learned something new.

dl8dtl:
> It really does. Learned something new.

👍

PR #829, opened 2022-01-11
CMake: add initial support for installing lib and include files

### LINKED ISSUE #827: shared library not installed by CMake infrastructure
When creating a shared library, the `avrdude` executable is linked against it, but `make install` doesn't install the newly built shared library.

yegorich:
Fixes #827


mariusgreuel:
Looks good!

yegorich:
@dl8dtl any comments?

yegorich:
I have added versioning info.

PR #832, opened 2022-01-13
Add missing ATmega and ATtiny targets

MCUdude:
As promised, here's the PR where I've gone through most of the "classic" AVR parts and added the various missing variants, mainly with suffix *A and *PA.


MCUdude:
The only "real" change in this PR is the ATmega1284P target. Previously it was an exact duplicate of the ATmega1284 target, but now it instead inherits all its properties, like it should have been done in the first place.

dl8dtl:
Curious, do you have a personal summary of all the new devices added by that change? I need it to update the `NEWS` file. If you don't have it, I'm going to pick it up out of the diff output.

MCUdude:
> Curious, do you have a personal summary of all the new devices added by that change?

Sure! Do you want me to edit the first post and put the summary there?

dl8dtl:
You can just put it here as a comment. I'll copy&paste it from there into `NEWS` then.

MCUdude:
I'll try to be as short as possible. Feel free to reword if that fits the NEWS format better!

> Add missing targets with *A and and *PA suffixes.
This includes ATmega48A/48PA/88A/88PA/168A/168PA, ATmega164A/644A/644PA, ATmega16A/32A/64A/128A.
A few missing ATtinys has also been added: ATtiny13A, ATtiny24A/44A/84A, ATtiny261A/461A/861A, ATtiny828R, ATtiny2313A and ATtiny1634R.

PR #834, opened 2022-01-13
Add missing tinyAVR-2 series targets

MCUdude:
I just found out that Avrdude currently doesn't support the latest and greatest tinyAVR-2 series with 32 kiB flash, the ATtiny3224, ATtiny3226, and ATtiny3227.

Borrowed from the megaTinyCore repo:
https://github.com/SpenceKonde/megaTinyCore/blob/8b9b2f1546fd6c67079268092cd700999021ed6c/megaavr/avrdude.conf#L16429-L16499

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -17519,6 +17519,78 @@ part parent    ".avr8x_tiny"
 ;
 
 #------------------------------------------------------------
+# ATtiny3224
+#------------------------------------------------------------
+
+part parent    ".avr8x_tiny"
+    id        = "t3224";
+    desc      = "ATtiny3224";
+    signature = 0x1E 0x95 0x28;
+
+    memory "flash"
+        size      = 0x8000;
+        offset    = 0x8000;
+        page_size = 0x80;
+        readsize  = 0x100;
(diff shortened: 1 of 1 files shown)


MCUdude:
BTW, do you want me to add the soon-to-be-released AVR-DD series to avrdude-conf as well? Datasheets are available, so all necessary information is available. 

64kiB flash and 8kiB RAM in a 14-pin package? Wow!

[AVR64DD28 product page](https://www.microchip.com/en-us/product/AVR64DD28)

[AVR-DD Preliminary datasheet](https://ww1.microchip.com/downloads/aemDocuments/documents/MCU08/ProductDocuments/DataSheets/AVR64DD32-28-Prelim-DataSheet-DS40002315A.pdf)

dl8dtl:
> BTW, do you want me to add the soon-to-be-released AVR-DD series to avrdude-conf as well?

Makes sense to me, yes, please.
Could you just update that PR here? Then we have it in one rush.

MCUdude:
> Could you just update that PR here? Then we have it in one rush.

Sure! Will do it a little later today

PR #836, opened 2022-01-14
Add missing tinyAVR-2, AVR DD and AVR EA targets

MCUdude:
This PR adds the following targets:

* ATtiny3224, ATtiny3226 and ATtiny3227. 
    - These are considered "safe", because I borrowed these from [megaTinyCore](https://github.com/SpenceKonde/megaTinyCore), and they have been tested on actual hardware.
* AVR16DD14/20/28/32, AVR32DD14/20/28/32 and AVR64DD14/20/28/32
* AVR8EA28/32, AVR16EA28/32/48, AVR32EA28/32/48 and AVR64EA28/32/64

For the AVR DD and EA series, I've used the source files from the [`pymcuprog`](https://github.com/microchip-pic-avr-tools/pymcuprog) project to get the necessary details. Please review closely. I've made sure the signature, memory sizes, and page sizes are correct, but please do the same to verify that it is correct. I don't own any DD or EA hardware, and due to the silicon crisis, neither of us won't for a long time I think.

https://github.com/microchip-pic-avr-tools/pymcuprog/tree/main/pymcuprog/deviceinfo/devices

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -18410,6 +18410,294 @@ part parent    ".avrdx"
 ;
 
 #------------------------------------------------------------
+# AVR16DD14
+#------------------------------------------------------------
+
+part parent    ".avrdx"
+    id        = "avr16dd14";
+    desc      = "AVR16DD14";
+    signature = 0x1E 0x94 0x34;
+
+    memory "flash"
+        size      = 0x4000;
+        offset    = 0x800000;
+        page_size = 0x200;
+        readsize  = 0x200;
+    ;
+
+    memory "eeprom"
+        size      = 0x100;
+        offset    = 0x1400;
+        page_size = 0x1;
+        readsize  = 0x100;
+    ;
+;
+
+#------------------------------------------------------------
+# AVR16DD20
+#------------------------------------------------------------
+
+part parent    ".avrdx"
+    id        = "avr16dd20";
+    desc      = "AVR16DD20";
+    signature = 0x1E 0x94 0x33;
+
+    memory "flash"
+        size      = 0x4000;
+        offset    = 0x800000;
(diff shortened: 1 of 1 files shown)


MCUdude:
Hold on, I'll run the `atdf-to-avrdude.xslt` script to verify that the data I've gathered is correct.

MCUdude:
One thing I don't get; why isn't all memory readsizes set to 1? AFAIK a single byte can be read from all memories

dl8dtl:
> One thing I don't get; why isn't all memory readsizes set to 1? AFAIK a single byte can be read from all memories

To be honest, I can't tell you, not without studying what the code is doing with that parameter.

MCUdude:
I found the following code in the pymcuprog sources:

```py
    # Read size is always 1 byte except for flash that can only read complete words
    readsize = 1
    if name in ['progmem']:
        readsize = 2
```

https://github.com/microchip-pic-avr-tools/pymcuprog/blob/a9411a8e4a5db8b54517c51da0bae96bf8385a65/pymcuprog/deviceinfo/harvest.py#L100-L103

In jtag3.c, this is the only real use I can find:
https://github.com/avrdudes/avrdude/blob/e5edecf95ea3a8dafc45309914ebb1a2127047f5/src/jtag3.c#L1157-L1160

This is the usage in @dbuchwald's SerualUPDI implementation:
https://github.com/avrdudes/avrdude/blob/e5edecf95ea3a8dafc45309914ebb1a2127047f5/src/serialupdi.c#L726-L751

dl8dtl:
> In jtag3.c, this is the only real use I can find:

stk500v2.c and jtagmkI.c/jtagmkII.c always use it instead of the pagesize.

PR #843, opened 2022-01-17
Fix receive buffer implementation in ftdi_syncbb programmer

mariusgreuel:
This PR fixes a bug in the ft245r implementation.

The ft245r.c code requests a number of bytes from ftdi_read_data(), but fails to check how many bytes actually have been returned, which resulted in a rx ring-buffer underflow. The code just assumes that the number of requested bytes have been returned and processes the (potentially not written) buffer.

The problem manifested itself with my Windows implementation of the libftdi library: For some reason, the implementation of the original libftdi1 library always returns the requested number of bytes, and not just parts of it. That may break depending on timing, I do not know. My Windows implementation based on the FTDI Windows drivers seems to be faster and typically return portions of the requested bytes, even if I do multiple reads as the original libftdi1 library does.

Now, one could suspect that some sort of timeout could help (and it does), however, I did not want to introduce a timeout because a) what should the timeout value be? and b) it slows down our second FTDI implementation in avrftdi.c, which actually requests more bytes than it really needs (and thus hits the timeout multiple times), while correctly checking whether the number of requested bytes have been returned.

So rather than fiddling with some arbitrary timeouts that are easy to break, I am just fixing the ft245r.c code. Partially, I should say, because there are other issues with not checked return values, potential ring-buffer overflows, etc.

### DIFF
--- a/src/ft245r.c
+++ b/src/ft245r.c
@@ -147,6 +147,7 @@ static struct {
 static struct {
     int discard;	// # of bytes to discard during read
     int pending;	// # of bytes that have been written since last read
+    int len;	// # of bytes in receive buffer
     int wr;		// write pointer
     int rd;		// read pointer
     uint8_t buf[FT245R_BUFSIZE];	// receive ring buffer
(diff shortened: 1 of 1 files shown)


MCUdude:
I don't own an ft245r based programmer, but I see there's a few of old ft232r/ft245r related bugs and patches that's still "open" here on Github. Some of them are has probably been fixed in the past. Du you have any plans going though these?
Thanks!

https://github.com/avrdudes/avrdude/issues?q=is%3Aissue+is%3Aopen+ft245r

mariusgreuel:
@MCUdude : Thanks for the hint, I knocked a couple of issues from the list. As with others, it is really hard to tell what is going on, so I leave them open for now.

Issue #848, opened 2022-01-20
USBasp returns ERANGE for unknown error


dl8dtl:
When encountering an unknown error code, the USBasp code returns ERANGE, which translates into the text "Result too large".
That does not make any much sense.

MCUdude:
I've never had any issues with USBasp. How can the issue be reproduced?

dl8dtl:
A friend of mine just caught that one. The reason is not yet clear, maybe a bug in FreeBSD 13. libusb eventually returns `LIBUSB_ERROR_UNKNOWN` which the USBasp code then translates into "Result too large".
Whatever it is, just throwing `ERANGE` for any undecided error is utter nonsense.

dl8dtl:
Closed by PR #850

PR #850, opened 2022-01-21
Fix libusb-1.0 error strings

dl8dtl:
Previously, all libusb-1.0 error strings have been translated
by libusb_to_errno() into an errno-like integer, only in order
to pass that one straight into strerror().

For unhandled libusb-1.0 error codes, ERANGE was returned, which
is just nonsense ("Result too large", bug #848).

Instead, use a single function errstr() now that either can
use strerror() on some well-defined errno-like code, or otherwise
just return a string that mentions the numeric value.

### DIFF
--- a/src/usbasp.c
+++ b/src/usbasp.c
@@ -66,42 +66,63 @@
 
 static libusb_context *ctx = NULL;
 
-static int libusb_to_errno(int result)
+static const char *errstr(int result)
 {
+	static msg[30];
+	int n = 0;
+
 	switch (result) {
 	case LIBUSB_SUCCESS:
-		return 0;
+		return "No error";
 	case LIBUSB_ERROR_IO:
-		return EIO;
+		n = EIO;
+		break;
 	case LIBUSB_ERROR_INVALID_PARAM:
-		return EINVAL;
+		n = EINVAL;
+		break;
 	case LIBUSB_ERROR_ACCESS:
-		return EACCES;
+		n = EACCES;
+		break;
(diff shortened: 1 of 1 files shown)


MCUdude:
I'm now getting build warnings after this PR was merged:

```
 CCLD     avrdude
ld: warning: directory not found for option '-FCoreFoundation'
  CC       libavrdude_la-ft245r.lo
  CC       libavrdude_la-stk500.lo
  CC       libavrdude_la-usbasp.lo
usbasp.c:71:9: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
        static msg[30];
        ~~~~~~ ^
usbasp.c:122:12: warning: incompatible pointer types passing 'int [30]' to parameter of type 'char *'
      [-Wincompatible-pointer-types]
                snprintf(msg, sizeof msg, "Unknown libusb error: %d", result);
                         ^~~
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/secure/_stdio.h:57:29: note: expanded from macro 'snprintf'
  __builtin___snprintf_chk (str, len, 0, __darwin_obsz(str), __VA_ARGS__)
                            ^~~
usbasp.c:123:10: warning: incompatible pointer types returning 'int [30]' from a function with result type 'const char *'
      [-Wincompatible-pointer-types]
                return msg;
                       ^~~
3 warnings generated.

```

dl8dtl:
> I'm now getting build warnings after this PR was merged:

Please, file this as a bug report so it won't be forgotten.

PR #852, opened 2022-01-23
CMake: split configuration in two files

### LINKED ISSUE #830: CMake: split CMakeLists.txt into common and source related
@mariusgreuel what about splitting the `CMakeLists.txt` and moving the common one to the project's root directory? This would simplify integration into different Linux distros and IDEs as, by default, they all expect `CMakeLists.txt` to be in the root folder.

yegorich:
The main `CMakeLists.txt` file in the project's root directory takes
care of the main project settings like project name and version,
handling the options, finding dependencies, etc.

The `src/CMakeLists.txt` handles options that are necessary to build
libavrdude library and avrdude binary.

Fixes #830

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,9 +29,6 @@ env:
 jobs:
   linux-x86_64:
     runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: ./src
     steps:
       - uses: actions/checkout@v2
       - name: Install prerequisites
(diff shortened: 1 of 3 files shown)


yegorich:
@mariusgreuel this splitting is working now. Though I was tempted to add `CMakeLists.txt` to `src/msvc` too i.e. making it an object library.

mariusgreuel (approved):
Just tried it with VS, looks good!

PR #858, opened 2022-01-31
CMake: use CMAKE_CURRENT_BINARY_DIR to locate avrdude.conf

yegorich:
With the split `CMakeLists.txt` infrastructure avrdude.conf
will be created in the `build/src` and not `build` folder. Hence,
fix its location in the install command.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -268,5 +268,5 @@ install(TARGETS libavrdude
     ARCHIVE DESTINATION lib
     PUBLIC_HEADER DESTINATION include COMPONENT dev
     )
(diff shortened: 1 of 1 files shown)


mariusgreuel:
Looks good. Perhaps one comment: When I have projects that include sub-projects, I tend to use the CURRENT_* variables instead of PROJECT_* variables. For instance, it might be better so that the output of the src/CMake project goes into the build/src folder, too. It does not make much of a difference for such a small project, but perhaps it is a little cleaner to use CMAKE_CURRENT_*.

dl8dtl:
> Perhaps one comment:

Merge anyway? (You can also merge it if you want.)

PR #859, opened 2022-01-31
Remove the "safemode" feature.

dl8dtl:
This feature has been designed with the sometimes quite flakey direct
(parallel or serial port attached) bitbang programming adapters in
mind that were quite common about two decades ago.

With parallel ports vanishing from modern PCs almost completely, and
the advent of various USB-attached low-cost programming devices,
this class of programmers disappeared almost completely.

Furthermore, the fuse combinations that were covered by the feature
are no longer around on all recent AVR devices, so for an ever
increasing number of devices, safemode already became meaningless and
was turned off anyway.

With the prospective version 7.x release, it's a good point in time to
introduce a major change like this one.


MCUdude:
I tried this on an Attiny13 and an ATmega1284P, and it seems to work fine, at least for me.

dbuchwald:
This is awesome, I will test it tomorrow against some of the other SerialUPDI/USBASP targets.

dl8dtl:
OK, I wiped of the dust from a lot of ancient equipment, digging out AVRDUDE's history. ;-)
I tried everything on an old computer with real parallel and serial ports.

1. programmer "bsd" (Brian Dean's original design, on parallel port), using an AT90S8515
2. programmer "ponyser" (Ponyprog bitbanger, on serial port), against an ATmega16

Apart from bug #860 all worked well. That bug is unrelated to the safemode feature, but it even prevented safemode from working correctly:

```
avrdude: safemode: Sorry, reading back fuses was unreliable. I have given up and exited programming mode
```

dl8dtl:
Fun pictures for those interested in history. ;-)
bsd
ponyser

MCUdude:
I love the pictures, thanks for sharing!

Should we perhaps just merge this? Once merged, many more people will be "testing this PR" when continue working on the Avrdude project.

dl8dtl:
> Should we perhaps just merge this?

I agree.

mariusgreuel:
I just tried to "reverse engineer" the RS232 programmer. Seems to be the premium version with two additional Zeners :-). I did not recognize the transistor, though. I just pulled out an old Fischer/Schlegel that I bought once in East Berlin, assuming it is a RGW part, however, could not find the equivalent. So I guess you will have to do all the testing with the old pony...

dl8dtl:
https://www-user.tu-chemnitz.de/~heha/basteln/Konsumg%C3%BCter/DDR-Halbleiter/ss216.gif

I guess any npn will do – it's just an inverter for that signal.

http://www.electroons.com/electroons/programmers.html

" This one is most reliable programmer."

:-))

dl8dtl:
I never really built it for actually using it, only for occasional testing. I guess I didn't even test it for about a decade now, and frankly, I don't know how well it would work behind an USB-RS232 bridge. Remember, it's been designed for true RS232 ports that have no additional USB packetizing overhead.

mariusgreuel:
> I never really built it for actually using it, only for occasional testing

OK. As soon as I have some time, I will link the picture in the wiki and label it the official jw programmer.

ericwazhung:
Having used an In-System-[AVR]-Programmer as an essential "daily [screw]-driver" tool for the vast majority of my piddly income since, literally, "the year 2000," and, *specifically* a USBtinyISP for *long*-over a decade (the same unit) up until about 3 months ago, and seeing that the kits are still available; AND being that the comparatively-recent linux distro I'd been using near the end of the USBtinyISP's life came with an avrdude version that still had "safemode"...

I feel it's somewhat relevant to point out my experiences: In that I received fuse-readback mismatches--as indicated by "safemode" errors--*numerous* times over, say, the last three years, with, again, a *USB* programmer (*not* serial nor parallel bitbanged, as this thread suggests safemode was only useful for) and, again, within a relatively-recent linux-distro's "apt-get install avrdude".

I'd further like to add to the discussion's consideration that, upon the spectacular death of my long-trusty/trusted USBtinyISP, I was in immediate need of a programmer *that night* and sought-out my 20+ year-old hand-assembled parallel-programmer (PonyProg, if I recall?) as a solution to that night's dilemma, and was pleased to find that avrdude did, still, in fact support it. 

So that when I was *unable* to locate the parallel-port-based unit I'd stopped using well over a decade prior, I almost set-out to solder one up *that night* to keep my *original* project going *that night*. 

Though, ultimately, it turned-out the FT2232-based breakout-board I'd used for JTAGging PIC32 chips some decade+ ago--before deciding that my long-trusty AVRs were still fine-and-dandy for my needs--could be repurposed for ISPing AVRs. And, thankfully, that seems to work for now. 

I am, however, still planning to purchase another USBtinyISP, as this FTDI-programmer seems fragile; all it would take is a strand of fine-gauged stranded wire, stripped slightly too vigorously, or a ball of solder carelessly-flicked from the iron, to fall between those tightly-spaced TQFP pins to possibly destroy the FTDI ISP programmer, or even, possibly, the USB port its tiny-spaced surfacemount connector is wired to. And, well, I've already had to resolder the fine-pitched surfacemount pins of its tiny USB connector, who knows if next time I'll be so lucky that it doesn't rip up the traces. Nevermind its being 3.3V... I know it says its pins are 5V tolerant, but having 5V feeding into a 3.3V chip's inputs 24/7 is a bit disconcerting.
That USBtinyISP, OTOH, lasted for darn-near twenty years and only needed its 74AHC125 replaced once... so I think I could get another 15 years out of another, without all these piddly-concerns from the FTDI...

TL;DR:

Until tonight's discovery that my recently-upgraded avrdude no longer has the "safemode" messages (discovered due to a bash script which looked for that message), I really had no idea *what* "safemode" was for, I just knew that sometimes my prototype projects didn't communicate properly with avrdude... and that it was often indicated by this thing called "safemode" that came as a [new-to-me!] feature in avrdude a few years back, during a prior linux-distro upgrade.

And, yet, I have been using a *USB* based programmer that *entire* time I'd *ever* received "safemode" failures--which this forum suggests shouldn't have happened.

I don't know why it happened; looking over the changes, as best I could understand of what github presented me, it seems the only thing "safemode" did was read the fuses three times before avrdude's doing anything else, then verifying that they remained the same *after* (or giving the option to rewrite them if they changed).

From what I see of the code-changes, none of that seems to have anything to do with parallel or serial programmers. 

It seems, instead, that maybe "safemode" may've usefully detected any number of communication-errors that may have less to do with the programmer's technology than to do with, say, stray 60Hz noise, ground-loops, or powersupply/regulator "brown-outs", or maybe even switching noise from once-respected brands' having switched to essentially relabelled dollarstore USB "chargers," or any number of other factors which could cause a bit to be flipped during an SPI exchange over a long cable in a protocol that does not implement checksums. 

(OH! I remember, now! A USB power-"wart" that put out 5.2V... when I replaced that, the rash of safemode messages disappeared! That was... 2-3 years ago. Though, I'd definitely gotten them since).

(In fact, since I've switched to the FTDI-programmer and the new linux distro, and thus the new avrdude which no longer has safemode, I think I do recall having had a weird problem that was fixed by rewriting the fuses, even though they'd long-ago been written to the correct values. Not certain about my ageing-memory on that, though.)

Now, who knows what this all means? Not I. 

If I knew avrdude's sourcecode and internal functionality to such great extents, then I'd surely lose the equal-amount of brainspace and time now used for paying emergency bills...

(which is to say "Thank You" to everyone who has made this tool available to us mortals, especially those of us who rely on its being there!)

Where was I? Oh mid-rant about kids on the lawn moving my tools around...

Nah, it's fine. This is a joke, aight?

BUT, *something* in "safemode" *was* useful at times (or at least seemed to be)... even with a USB programmer... OTOH, who knows, maybe all those extra transactions *caused* it to be more-likely to run into extraneously-flipped bits...?

Dunno, just thought someone might like my 2cents.

(BTW, Great pictures, up in the thread!)

stefanrueger:
@ericwazhung Thanks for a heart-warming and considered comment! It makes a few good points, including how unreliable programmers, the electromagnetic environment, setup, cables, the workbench etc can be. All of which can brick a part by writing an unintentionally modified fuse value. Safemode was a neat idea intended to *reduce* that risk. It would need to be completely re-designed as literally hundreds of parts have come out since, with new and different fuse memories. 

These days, I recommend to thoroughly [test a programmer](https://github.com/avrdudes/avrdude/discussions/2129) and setup *before* using both in anger. I certainly have bricked two ATmega328P boards with my [FTDI USB connector](https://github.com/avrdudes/avrdude/assets/14282046/e101d6cd-3085-4b7d-95b8-a0e552541579), [see note (2) here](https://github.com/avrdudes/avrdude/discussions/2129#discussioncomment-17358077). So, I wouldn't use that setup for any commercial activity.

PR #872, opened 2022-02-11
Tweak programmer info formatting strings

MCUdude:
Now all colons are on a straight line, just like #853 did to all jtag3 compatible programmers


MCUdude:
Something went terribly wrong here. I'll see if I can fix the issue

PR #876, opened 2022-02-12
Add serial_drain() to updi_physical_send_double_break().

dl8dtl:
This is needed here on FreeBSD.

### DIFF
--- a/src/updi_link.c
+++ b/src/updi_link.c
@@ -161,6 +161,11 @@ static int updi_physical_send_double_break(PROGRAMMER * pgm)
 
   updi_set_rtsdtr_mode(pgm);
 
+  /*
+   * drain any extraneous input
+   */
+  serial_drain(&pgm->fd, 0);
+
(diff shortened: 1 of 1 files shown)


dbuchwald:
Do you realize that this is the final fix for SerialUPDI I completed today? Hahahaha :)

dl8dtl:
So, you say, I can just drop this PR, and use your one instead?

Will give that a try tonight.

dbuchwald:
Yeah, I think you can use my PR instead, it has two more changes which are required for PL2303HX operation (increased timeout and replacement of "double break" by single fast break at the beginning).

You have to tell me one day how did you come up with the idea of missing `serial_drain` there - it was actually pretty nasty bug!

dl8dtl:
> You have to tell me one day how did you come up with the idea of missing `serial_drain` there - it was actually pretty nasty bug!

I can't remember anymore. Probably trial & error.

Issue #875, opened 2022-02-12
UPDI via EDBG: fails when no target connected


dl8dtl:
So well, OK, obviously no target communication can happen in that situation. ;-)
However, parameter setup on the EDBG itself ought to work when using the `-F` option – that way it could e.g. be done on an STK500 without a target connection.
I was trying to adjust Vtarget on an Curiosity Nano where the actual target is disconnected.
```
% ./build_freebsd/src/avrdude -c pkobn_updi -p avr128db48 -tF

         Vtarget         : 3.31 V
         PDI/UPDI clock Xmega/megaAVR : 100 kHz

avrdude: Short read, read only 0 out of 64 bytes
avrdude: jtag3_edbg_recv(): Unexpected response 0x12
avrdude: retrying with external reset applied
avrdude: jtag3_edbg_send(): Unexpected response 0x81, 0x11
avrdude: jtag3_edbg_recv(): Unexpected response 0x80
avrdude: retrying with external reset applied
avrdude: JTAGEN fuse disabled?
avrdude: initialization failed, rc=-1
avrdude: AVR device initialized and ready to accept instructions
avrdude: Device signature = 0x000000 (retrying)
avrdude: Device signature = 0x000000 (retrying)
avrdude: Device signature = 0x000000
avrdude: Yikes!  Invalid device signature.
avrdude: Expected signature for AVR128DB48 is 1E 97 0C
avrdude> vtarg 4
>>> vtarg 4 
avrdude: jtag3_edbg_send(): Unexpected response 0x81, 0x11
avrdude: jtag3_edbg_recv(): Unexpected response 0x80
avrdude: jtag3_set_vtarget(): cannot obtain V[aref]
avrdude (vtarg): failed to set V[target] (rc = -1)
avrdude> q
>>> q 
avrdude: jtag3_edbg_send(): Unexpected response 0x81, 0x11
avrdude: jtag3_edbg_recv(): Unexpected response 0x80
avrdude: jtag3_edbg_send(): Unexpected response 0x81, 0x11
avrdude: jtag3_edbg_recv(): Unexpected response 0x80
avrdude: jtag3_edbg_signoff(): unexpected response 0x81, 0x11
avrdude: jtag3_edbg_signoff(): unexpected response 0x01, 0x00

avrdude done.  Thank you.
```

MCUdude:
[...]

PR #877, opened 2022-02-13
Reduce jtag3 output verbosity

MCUdude:
+ some other minor tweaks.

#875 related


MCUdude:
@dl8dtl I guess this can be merged as it is? The only changes besides verbosity level is to not print the "JTAGEN message" if not in JTAG mode, and format the Vtarget string so that the colons lines up regardless of verbosity level.

dl8dtl:
yep, looks good

Issue #38, opened 2021-12-10
[bug #14374] Feature Request : terminal mode : improving EEPROM writes


avrs-admin:
Vincent Trouilliez <vincent.trouilliez@modulonet.fr>
Fri 02 Sep 2005 04:08:14 PM UTC

I love to use the terminal mode to look at the content of the EEPROM, the data is very niceley presented.
It is so convenient to modify the EEPROM quickly/directly without having to make a hex file and download it !

So, in order to extend the usefulness of this command, it would be great to make its parser more clever/flexible.

For example we could type in one single command :

write eeprom 0x0000 "A string",0,"another string",0,2.456,-456.76,16036,-113

A bit like when declaring constants with the DB DW etc, in an assembler. The more data formats the parser could understand, the more marvelous/useful/practical/magic it would be ! :o)

This issue was migrated from https://savannah.nongnu.org/bugs/?14374

avrdude-admin:
Joerg Wunsch <joerg_wunsch>
Sat 17 Sep 2005 07:45:36 PM UTC

Without a patch, that's going to be rather low-piority, as
the requested feature while being useful, might require a
bit of thought before.

MCUdude:
Implemented in #880

PR #880, opened 2022-02-17
Avrdude terminal write improvements

### LINKED ISSUE #879: Make the terminal 'write' functionality more useful
The terminal mode is really handy if you have a target where you want to just check something or do minor changes. However, I feel like the 'write' functionality has more potential.

The current syntax is this:
`write <memtype> <addr> <byte1> <byte2> ... <byteN>`

Features that would be neat to have:

```
# Fill user defined memory section with a known value (three dots as part of the syntax)
# write <memtype> <start addr> <length> <byte to write to all memory locations> <...> 

# example
write eeprom 0x40 0x10 0x00 ...
```

```
# Write strings with \0 at the end
# write <memtype> <start addr> <string in quotes>

# Example (writes Hello World!\0 to EEPROM)
write eeprom 0x80 "Hello World!"
```

Any thoughts?

MCUdude:
I've been working one some improvements regarding the Avrdude terminal write functionality. I have not yet updated any docks; I'd like to get some feedback first.

With this PR it's now possible to fill a memory section, write 8, 16 and 32-bit integers, characters and floats to memory. Negative numbers works too, and everything is stored in little endian.

Any thoughts? I'm all ears!

I have yet to add support for strings in double quotes, but this is quite a bit of work I believe.

#38 related
Closes #879 

```
avrdude: Device signature = 0x1e9650 (probably m4808)
avrdude> write eeprom 0x00 0x100 0xff ...
>>> write eeprom 0x00 0x100 0xff ... 

avrdude> read eeprom 0 256
>>> read eeprom 0 256
0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0010  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0040  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0050  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0060  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0070  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0080  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0090  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write eeprom 0x00 'H' 'e' 'l' 'l' 'o' 1 2 3
>>> write eeprom 0x00 'H' 'e' 'l' 'l' 'o' 1 2 3 

avrdude> write eeprom 0x10 0x10 'a' 'b' 'c' 'd' ...
>>> write eeprom 0x10 0x10 'a' 'b' 'c' 'd' ... 

avrdude> write eeprom 0x20 0xdeadbeef
>>> write eeprom 0x20 0xdeadbeef 

avrdude> write eeprom 0x30 3.141592
>>> write eeprom 0x30 3.141592 

avrdude> write eeprom 0x40 0x10 'A' 'V' 'R' 0xdeadbeef ...
>>> write eeprom 0x40 0x10 'A' 'V' 'R' 0xdeadbeef ... 

avrdude> read eeprom 0x00 0x100
>>> read eeprom 0x00 0x100 
0000  48 65 6c 6c 6f 01 02 03  ff ff ff ff ff ff ff ff  |Hello...........|
0010  61 62 63 64 64 64 64 64  64 64 64 64 64 64 64 64  |abcddddddddddddd|
0020  ef be ad de ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0030  d8 0e 49 40 ff ff ff ff  ff ff ff ff ff ff ff ff  |..I@............|
0040  41 56 52 ef be ad de ef  be ad de ef be ad de ef  |AVR.............|
0050  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0060  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0070  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0080  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0090  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
```

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -23,7 +23,7 @@
 #include <ctype.h>
 #include <string.h>
 #include <stdio.h>
-//#include <stdint.h>
+#include <stdint.h>
 #include <stdlib.h>
 #include <limits.h>
 
@@ -329,18 +329,6 @@ static int cmd_dump(PROGRAMMER * pgm, struct avrpart * p,
 static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
 		     int argc, char * argv[])
 {
-  char * e;
-  int len, maxsize;
-  char * memtype;
-  long addr;
-  long i;
-  unsigned char * buf;
-  unsigned char b;
-  int rc;
-  int werror;
-  int write_mode, start_offset;
-  AVRMEM * mem;
-
   if (argc < 4) {
     avrdude_message(MSG_INFO,
       "Usage: write <memtype> <start addr> <data1> <data2> <dataN>\n"
@@ -350,19 +338,22 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
     return -1;
   }
 
-  memtype = argv[1];
-
-  mem = avr_locate_mem(p, memtype);
+  int32_t i;
+  uint8_t write_mode;       // Operation mode, "standard" or "fill"
+  uint8_t start_offset;     // Which argc argument
+  int32_t len;              // Number of bytes to write to memory
+  char * memtype = argv[1]; // Memory name string
+  AVRMEM * mem = avr_locate_mem(p, memtype);
   if (mem == NULL) {
     avrdude_message(MSG_INFO, "\"%s\" memory type not defined for part \"%s\"\n",
             memtype, p->desc);
     return -1;
   }
+  uint32_t maxsize = mem->size;
 
-  maxsize = mem->size;
-
-  addr = strtoul(argv[2], &e, 0);
-  if (*e || (e == argv[2])) {
+  char * end_ptr;
+  int32_t addr = strtoul(argv[2], &end_ptr, 0);
+  if (*end_ptr || (end_ptr == argv[2])) {
     avrdude_message(MSG_INFO, "%s (write): can't parse address \"%s\"\n",
             progname, argv[2]);
     return -1;
@@ -377,8 +368,8 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
   // Figure out how many bytes there is to write to memory
   if (strcmp(argv[argc - 1], "...") == 0) {
     write_mode = WRITE_MODE_FILL;
-    len = strtoul(argv[3], &e, 0);
-    if (*e || (e == argv[3])) {
+    len = strtoul(argv[3], &end_ptr, 0);
+    if (*end_ptr || (end_ptr == argv[3])) {
       avrdude_message(MSG_INFO, "%s (write ...): can't parse address \"%s\"\n",
             progname, argv[3]);
       return -1;
@@ -387,7 +378,7 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
     write_mode = WRITE_MODE_STANDARD;
   }
 
-  buf = malloc(mem->size);
+  uint8_t * buf = malloc(mem->size);
   if (buf == NULL) {
     avrdude_message(MSG_INFO, "%s (write): out of memory\n", progname);
     return -1;
@@ -442,13 +433,14 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
       } else if (suffix == '\'') {
         data.size = 1;
       }
+
       // Try integers
-      data.ll = strtoll(argv[i], &e, 0);
-      if (*e || (e == argv[i])) {
+      data.ll = strtoll(argv[i], &end_ptr, 0);
+      if (*end_ptr || (end_ptr == argv[i])) {
         // Try float
-        data.f = strtof(argv[i], &e);
+        data.f = strtof(argv[i], &end_ptr);
         data.is_float = true;
-        if (*e || (e == argv[i])) {
+        if (*end_ptr || (end_ptr == argv[i])) {
           data.is_float = false;
           // Try single character
           if (argv[i][0] == '\'') {
@@ -466,7 +458,7 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
         (((strncmp(argv[i], "0x", 2) == 0) && strlen(argv[i]) > 3) || \
         (data.ll > 0xFF && strlen(argv[i]) > 2))) {
         avrdude_message(MSG_INFO, "Warning: no size suffix specified for \"%s\". "
-                                  "Writing %d bytes\n",
+                                  "Writing %d byte(s)\n",
                                   argv[i],
                                   llabs(data.ll) > 0xFFFFFFFF ? 8 :
                                   llabs(data.ll) > 0x0000FFFF || data.is_float ? 4 : \
@@ -500,22 +492,24 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
   }
 
   pgm->err_led(pgm, OFF);
-  for (werror=0, i=0; i < (len + data.bytes_grown); i++) {
-    rc = avr_write_byte(pgm, p, mem, addr+i, buf[i]);
+  bool werror = false;
+  for (i = 0; i < (len + data.bytes_grown); i++) {
+    int32_t rc = avr_write_byte(pgm, p, mem, addr+i, buf[i]);
     if (rc) {
       avrdude_message(MSG_INFO, "%s (write): error writing 0x%02x at 0x%05lx, rc=%d\n",
               progname, buf[i], addr+i, rc);
       if (rc == -1)
         avrdude_message(MSG_INFO, "write operation not supported on memory type \"%s\"\n",
                         mem->desc);
-      werror = 1;
+      werror = true;
     }
 
+    uint8_t b;
     rc = pgm->read_byte(pgm, p, mem, addr+i, &b);
     if (b != buf[i]) {
       avrdude_message(MSG_INFO, "%s (write): error writing 0x%02x at 0x%05lx cell=0x%02x\n",
                       progname, buf[i], addr+i, b);
-      werror = 1;
+      werror = true;
     }
 
     if (werror) {
--- a/src/term.c
+++ b/src/term.c
@@ -416,7 +416,7 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
       int64_t ll;
       uint8_t a[8];
     };
-  } data;
+  } data = {.bytes_grown = 0, .size = 0, .is_float = false, .ll = 0};
 
   for (i = start_offset; i < len + start_offset - data.bytes_grown; i++) {
     data.is_float = false;
--- a/src/term.c
+++ b/src/term.c
@@ -371,7 +371,7 @@ static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
   }
 
   // Figure out how many bytes there is to write to memory
-  if(strcmp(argv[argc - 1], "...") == 0) {
+  if (strcmp(argv[argc - 1], "...") == 0) {
     write_mode = WRITE_MODE_FILL;
     len = strtoul(argv[3], &e, 0);
     if (*e || (e == argv[3])) {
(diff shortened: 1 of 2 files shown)


mariusgreuel:
From the logs, there is not really something obvious:
https://github.com/avrdudes/avrdude/runs/5239599936?check_suite_focus=true

A re-run did not help either. Perhaps there has been some MSYS2 package upgrade that broke it. Also, Github recently updated their Windows distribution.

I will look into that some other time.

MCUdude:
Thanks @mariusgreuel, now all builds are passing.

@dl8dtl is this something you will accept as it is, or does major/minor changes has to be made?

One thing I wasn't able to get working in another was was the conversion from float to long.

```
write_val = ((char)*(ptr+3)<<24) + ((char)*(ptr+2)<<16) + ((char)*(ptr+1)<<8) + (char)*ptr;
```

If there is a better way that doesn't screw up the result, let me know!

(For reference, I've made sure the data is written exactly like the Arduino EEPROM library does it.
This means that you can write data to and address, and retrieve it directly. Nothing special, just pointers doing their thing)

```
#Avrdude terminal
write eeprom 0x00 3.141592

#Microcontroller code
float f;
EEPROM.get(0x00, f); // (start addr, variable to store data to)
```

dl8dtl:
> @dl8dtl is this something you will accept as it is, or does major/minor changes has to be made?

See my review comments above.

> `write_val = ((char)*(ptr+3)<<24) + ((char)*(ptr+2)<<16) + ((char)*(ptr+1)<<8) + (char)*ptr;`

I'd never use "char" in that context. If at all, use "uint8_t". In particular, the unsignedness is important for the shift operations.
The portable way would be to take the address of the target, and use memcpy() or a loop, copying byte-wise.
The hacky (but widely used) way is to use a union.
But why convert a double to a uint32_t at all? After all, the EEPROM is byte-wise, so your target would better be a byte array to copy to.

MCUdude:
> See my review comments above.

For some reason I can't see none of the comments... Can you try to write single line comments under the "Files changes" tab?

> I'd never use "char" in that context. If at all, use "uint8_t". In particular, the unsignedness is important for the shift operations.

I use intX_t/uintX_t if I can, but I haven't done it here because I try to stick with the same data types as used elsewhere in the file.

> The portable way would be to take the address of the target, and use memcpy() or a loop, copying byte-wise.
The hacky (but widely used) way is to use a union.
But why convert a double to a uint32_t at all? After all, the EEPROM is byte-wise, so your target would better be a byte array to copy to.

It's converting it to a 32-bit signed, not unsigned. Yes, a 4-byte array and memcpy would probably be better, but I'll have to make sure endianness is correct. 
I'll see what I can do later today.

dl8dtl:
I did write it under "Files changed", but single-line is not always appropriate …

Feel free to overhaul more of that terminal code, I think it's pretty ancient. If it uses "char" where it'd better be "uint8_t" these days, then that's a reason to change it. Most of that program code is 20 years old when C99 was just new, and one could not rely on compilers implementing it.

MCUdude:
@dl8dtl Now I've improved the implementation quite a bit. I still have 64-bit doubles to handle, but I'll save that for tomorrow eventing. But besides for this and a missing suffix to force 8-bit data, does the code look all right?

I'm not very proud of the way reading suffixes is implemented, but I searched around a couldn't find a better way that would either reduce the amounts of code lines or make the code more readable. I'm all ears if you have suggestions on what can be improved!

dl8dtl (commented):
Do you want to improve this by using uint<N>_t before merging it?

Something that just occurred to me: "write_mode" could be improved by using an enum instead of #defines.

MCUdude:
Please don't merge yet. I'd like to do some more improvements before it's ready.

> Something that just occurred to me: "write_mode" could be improved by using an enum instead of #defines.

Definitely, I personal don't prefer #define constants, but almost always use enums instead. The reason why I used #define here is to "match" the existing code style.

MCUdude:
@dl8dtl one this that still puzzles me is when the warning message regarding size of data to write is printed. How "smart" should I make it? After all, it is a bit of a hazzle to implement this. This is how the current implementation works:

```
Hex:
0x0F		No warning
0xFF		No warning
0x0FF		Warning
0xFFFF		Warning
0x0FFFF		Warning
0xFFFFFF	Warning
0xFFFFFFFF	Warning
Any negative #	No warning

Dec:
0		No warning
99		No warning
255		No warning
256		Warning
1000		Warning
1000000		Warning
Any negative #	No warning
```

Is this "good enough"? If not, do you know a better way of checking all this than what I do (in the code below)?

```c
      // Print warning if data size might be ambiguous
      if(!data.size && \
        (((strncmp(argv[i], "0x", 2) == 0) && strlen(argv[i]) > 3) || \
        (data.ll > 0xFF && strlen(argv[i]) > 2))) {
        avrdude_message(MSG_INFO, "Warning: no size suffix specified for \"%s\". "
                                  "Writing %d byte(s)\n",
                                  argv[i],
                                  llabs(data.ll) > 0xFFFFFFFF ? 8 :
                                  llabs(data.ll) > 0x0000FFFF || data.is_float ? 4 : \
                                  llabs(data.ll) > 0x000000FF ? 2 : 1);
      }
```

dl8dtl:
> Is this "good enough"?

I think it is.

The important thing is to document it properly, and hint users to use suffixes to disambiguate the size guesses.

How would you want to handle documentation? Can you update either the texinfo file or the man page, or is their markup too much "black magic" for you? In the latter case, just email me your documentation addition as prosa, and I'll handle the formatting.

MCUdude:
>> Is this "good enough"?

> I think it is.

I haven't really been very pleased with this solution since it will spit out a warning even when the user is consistent and write "valid" data where the number of digits matches 1, 2, 4 or an 8-byte data type. However, I think I've found a solution that makes sense.

Will cause a warning:
* Base 10 number greater than 255
* All negative hex numbers regardless of size (-0xFF for instance)
* Positive hex numbers with leading zero and the number of digits (behind 0x) are 3, 5, 6 or 7
  - 0x0FF, 0x0FFFF, 0x0FFFFF will all cause warnings
  - 0xF00, 0xF000, 0xF0000, 0xF00000, 0xF000000, 0xF0000000 will not cause warnings

Please give it a try!

> How would you want to handle documentation? Can you update either the texinfo file or the man page, or is their markup too much "black magic" for you? In the latter case, just email me your documentation addition as prosa, and I'll handle the formatting.

I was hoping to get a Latex editor working on my machine, I've installed a VSCode Latex plugin that should make editing and previewing a breeze, but I have yet to figure out how to get started. I'll keep you updated on the progress! I'll provide all the necessary documentation if I end up giving up.

dl8dtl:
> Please give it a try!

Will do, tomorrow.



> I was hoping to get a Latex editor working on my machine

texinfo is not LaTeX!

While I could recommend LaTeX for nicely printed texts :-), it won't help you with the texinfo docs.

You should be able to install texinfo from either brew or Mac ports.

MCUdude:
> texinfo is not LaTeX!

That explains why I couldn't get the Latex plugin to work! Thanks for the hint!

BTW there is one bug I just found that I haven't addressed yet. It doesn't handle negative base10 numbers very well. It only write one byte to memory when passing -255 for instance, two bytes when writing -65535 and four bytes when writing -65536.

dl8dtl:
> It doesn't handle negative base10 numbers very well.

As long as single-byte numbers (i.e. everything between -127 and -1) is handled correctly, I wouldn't bother too much.

MCUdude:
>> It doesn't handle negative base10 numbers very well.

> As long as single-byte numbers (i.e. everything between -127 and -1) is handled correctly, I wouldn't bother too much.

It does work for everything between -127 and -1, but since I've already put quite a lot of effort into this I'd like to get everything "right".

I'll take a look at this later this evening.

dl8dtl:
I'm impressed already with the current effort, kudos @MCUdude !

MCUdude:
> I'm impressed already with the current effort, kudos @MCUdude !

Thanks for the recognition! It has definitely been a handful. Doing string manipulation using Python after this would feel like cheating; way too easy!

I think this PR is finished now. There are other things that should be implemented like "proper support for doubles and support for strings in quotes. But that's for another time! I'm not even sure how to handle strings in a well structured way. The command `write eeprom 0x00 "Hello world! This is a string test."` would have to search for start and end quotes, and handle the fact that a single string is spread out over several arguments. The current implementation would probably have be completely rewritten in order to support something like this.

Regarding the documentation, it _would_ be fun to give texinfo a try, but I have literally no clue how I even begin...
I installed texinfo, and now I can execute the following binaries:
```
$ ls /usr/local/Cellar/texinfo/6.8/bin/
info		install-info	makeinfo	pdftexi2dvi	pod2texi	texi2any	texi2dvi	texi2pdf	texindex
```
None of these reveals how I as a complete newbie can edit a texi file in a "different" way than what a plain text editor can.

dl8dtl:
> Regarding the documentation, it _would_ be fun to give texinfo a try, but I have literally no clue how I even begin...

If you can use autoconf/automake, you can do

`./configure --enable-doc`

Then, a normal "make" will descend into the `doc/` subdirectory, and build the documentation.

I don't think something similar has already been added to the CMake infrastructure.

Basically, it's the `makeinfo` command, but `Makefile.am` has some hacks to auto-generate some include files first.

MCUdude:
I'm currently using autoconf/automake for building Avrdude, becuase the commands are easy to remember and works well on my computer with my current setup. I'll see what I can do! If I give up, I'll at least write the proper documentation so you perhaps can add it for me.

One question I forgot to ask earlier: As a guy that has almost exclusively used C on microcontrollers where dynamic memory allocation is a no-no, why use malloc to create `buf` when a "regular" array would do just fine? The size of the allocated memory doesn't need to be known at compile time, and since the buffer only reflects an AVR memory, it's not a large chunk to throw on the stack either...

dl8dtl:
I'll try building the docs.

MCUdude:
Well, One error has disappeared, but there are still several others, and I can't make ut build, sorry:

```
Making all in doc
../avrdude -C ../avrdude.conf -c \? 2>&1 \
	| gawk '$2 ~ /^=$/ {printf("@item @code{%s} @tab %s\n",$1,gensub("[^=]+=[ \t]*","",1))}' \
	| sed "s# *,\? *<\?\(http://[^ \t>]*\)>\?#,@*\n@url{\1}#g" \
	>programmers.texi
../avrdude -C ../avrdude.conf -p \? 2>&1 \
	| gawk '$2 ~ /^=$/ {printf("@item @code{%s} @tab %s\n",$1,$3)}' \
	| sed -e "`sed 's:\([^ \t]*\)[ \t]*\(.*\):s/\1$/\1 \2/g:g' <parts_comments.txt`" \
	>parts.texi
../avrdude -C ../avrdude.conf -c \?type 2>&1 \
	| gawk '$2 ~ /^=$/ {printf("@item @code{%s} @tab %s\n",$1,gensub("[^=]+=[ \t]*","",1))}' \
	| sed "s#<\?\(http://[^ \t,>]*\)>\?#@url{\1}#g" \
	>programmer_types.texi
  MAKEINFO avrdude.info
avrdude.texi:320: warning: `.' or `,' must follow @xref, not `f'.
avrdude.texi:645: warning: `.' or `,' must follow @xref, not `e'.
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2426: Next field of node `Documentation' not pointed to (perhaps incorrect sectioning?).
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2259: This node (Windows) has the bad Prev.
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2426: `Documentation' has no Up field (perhaps incorrect sectioning?).
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2373: Next field of node `Windows Port Names' not pointed to (perhaps incorrect sectioning?).
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2311: This node (Windows Configuration Files) has the bad Prev.
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2373: `Windows Port Names' has no Up field (perhaps incorrect sectioning?).
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2311: Next field of node `Windows Configuration Files' not pointed to (perhaps incorrect sectioning?).
/Users/hans/Downloads/avrdude/src/doc//avrdude.texi:2373: This node (Windows Port Names) has the bad Prev.
makeinfo: Removing output file `avrdude.info' due to errors; use --force to preserve.
make[2]: *** [avrdude.info] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
```

PR #882, opened 2022-02-18
Remove libhid support in ser_avrdoper.c in favor of libhidapi

mariusgreuel:
This PR removes libhid support from ser_avrdoper.c in favor of libhidapi.

libhid is the native HID library in Windows. As libhidapi is available for all Windows builds, a fallback to the native HID library is no longer needed.

### DIFF
--- a/src/ser_avrdoper.c
+++ b/src/ser_avrdoper.c
@@ -26,11 +26,12 @@
 
 #include "ac_cfg.h"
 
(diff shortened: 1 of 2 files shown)


mariusgreuel:
+5/-185 lines of code.

Also, see #883

dl8dtl (approved):
Well, you''re Mr. Windows here anyway. ;-)
If you are happy with that, for unixoid systems nothing changes by it.

PR #887, opened 2022-02-21
Shorten timing of arduino open and serial drain functions

stefanrueger:
This changes enables bootloaders to exit after a short time (256 ms) rather
than the current minium waiting time of 512 ms. After Reset, control can be
handed to the sketch faster, which is important in some applications.

### DIFF
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -93,7 +93,8 @@ static int arduino_open(PROGRAMMER * pgm, char * port)
   /* Clear DTR and RTS to unload the RESET capacitor 
    * (for example in Arduino) */
   serial_set_dtr_rts(&pgm->fd, 0);
-  usleep(250*1000);
+  /* 25 ms is ample for dis/charging the capacitor between reset and DTR/RTS */
+  usleep(25*1000);
   /* Set DTR and RTS back to high */
   serial_set_dtr_rts(&pgm->fd, 1);
   usleep(50*1000);
--- a/src/ser_posix.c
+++ b/src/ser_posix.c
@@ -505,7 +505,7 @@ static int ser_drain(union filedescriptor *fd, int display)
   unsigned char buf;
 
   timeout.tv_sec = 0;
-  timeout.tv_usec = 250000;
+  timeout.tv_usec = 100*1000;
 
   if (display) {
     avrdude_message(MSG_INFO, "drain>");


mariusgreuel:
I also often noticed a delay (depending on the programmer) that is way beyond what you would expect, and also the serial code on Windows has a longer delay as on Linux. I thought about changing that, but did not really feel comfortable doing that: After all, we are not on a real-time OS, so whenever you do something like "A => wait => B" you cannot be sure when A or B is actually executed, or when A or B is sent down the (USB) bus. Without knowing the history, it is hard to tell why the delays are often higher than they have to be on first glance, but I would expect them to have a pretty good safety margin.

That said, I am not saying we should not shorten the delays, I am just raising a finger and suggesting thorough testing.

MCUdude:
> That said, I am not saying we should not shorten the delays, I am just raising a finger and suggesting thorough testing.

I second this. I'm sure there are hardware out there that relies on undocumented behavior like this. 

> After Reset, control can be handed to the sketch faster, which is important in some applications.

When exactly would this be important for an application? In the "field", a target is only programmed once, and this PR won't affect the boot time on AVR after a power-on reset or the time it takes for the bootloader to time out after a reset pulse.

dl8dtl:
There are so many ideas about changing timeout values lately … there's also that WiFi programmer (wAVR) that needs longer response timeouts, as the responses have to travel across the network.

We already have several "default_*" values in the config grammar. I wonder whether we should also provide a "default_timeout" there, thus allowing easy overrides by  the user's `~/.avrduderc`. We could then also have a per-programmer default (which should make wAVR happy).

That won't help with the first part of this patch. I'd be very interested in hearing opinions from the Arduino folks – I know they used to ship their own hacked-up version of AVRDUDE.

stefanrueger:
Application examples:
  - System uses reset to get µP out of sleep state to do something and then goes to sleep again (imagine a board where all other µP pins are already in use)
  - Less waiting during development when frequently uploading sketches (my bootloaders have shorter WDT settings, indeed, than are currently possible with avrdude's arduino programmer)

Other examples exist for which one might wish to reduce unnecessary reset start-up times.

I agree with Jörg, this is better solved by delay constants per programmer (or by command line parameters).

mariusgreuel:
> Other examples exist

- A snappy application, i.e. faster than one can type, is just a joy to work with

stefanrueger:
Summarising for the ser_drain() timeout: this is better solved utilising a _default constant in the grammar (though, I am afraid I can be of little help with that).

The usleep() time in arduino_open() sandwiched between setting RTS/DTR on and off can stay as a constant, but I argue this constant should be much smaller: This usleep() is to allow time for the dtr-reset cap to discharge, and the board sees a reset pulse **after** this usleep() when setting RTS/DTR high.  The lower bound of the usleep() time should be in the order of RC of the reset circuit:
 - A typical board has C=100 nF and R=10k, so 1 ms is a good value for those boards. In fact, on the 0.1 µF/10k board I tested with, 0.35 ms sleep still works (just about). 
 - R is limited by the internal pull-up, so unlikely to be higher than 50k. In the absence of an external pull-up, I would consider any reset cap larger than 470 nF ill-designed. 

Hence, the suggested 25 ms sleep (470 nF times 50k) is contains a large margin for a large gamut of boards. I hold that this time does not need to be parameterised. There is no theoretical upper limit for this usleep() time. 10 seconds (or 10 minutes) also work, but nothing useful happens. 250 ms encroaches on user experience and *is* quite noticable, and should be reduced as it serve no useful purpose.

Summarising, 25 ms works safely and is better.

Changing focus while analysing timings: Actually, the sleep *after* resetting the board is more interesting: The board is contacted only after this time *plus* the ser_drain() time. The sum of these two times should be larger than the time it takes for the board to come out from reset into the bootloader. That time can be vvv long (think small F_CPU and adverse fuse settings). It makes sense to be able to change *that* sleep time after reset in arduino_open() with a command line option and not a per-programmer constant, because this depends  on the *board* and not the programmer. 

In theory it shouldn't matter whether avrdude talks to the board before it comes out of reset, because the stk500_getsync() routine should rescue this situation. However, ser_posix.c's long serial_recv_timeout is set to 5000 ms, so avrdude waits for 5 seconds for the board's first response, by which time the bootloader is likely to have timed out (unless set to 8 s, which I'd rather not). So, stk500_getsync() doesn't work either for me and has, for precisely this reason never worked for me. It's always either hunky-dory or the dreaded 10 times 5 second timeout message loop. 

I think it's a good idea for stk500_getsync() to set the serial_recv_timeout to 100 ms (as is done in other programmers), but I don't know the ramifications for other programmers using set500_getsync(). I am sure, however, that the arduino programmers is OK with 100 ms. At least not-in-sync is over in a few seconds (there is a ser-drain() timeout in play too) rather than a minute.

dl8dtl:
> Hence, the suggested 25 ms sleep (470 nF times 50k) is contains a large margin for a large gamut of boards. 

I'm always a bit hesitant with that kind of statements. But I agree that it very unlikely needs a larger value.

> In theory it shouldn't matter whether avrdude talks to the board before it comes out of reset, because the stk500_getsync() routine should rescue this situation. However, ser_posix.c's long serial_recv_timeout is set to 5000 ms, so avrdude waits for 5 seconds for the board's first response,

The idea behind the long timeout value is that you have a quick wait-retry sequence initially, to get in sync with a mis-synchronized board. Once you have got a first response, you are synchronized, and then you can wait a bit longer for any responses, as normally, the expected amount of data arrives much before that timeout.

mariusgreuel:
> Summarising, 25 ms works safely

My statistics knowledge tells me that a sample size of one is not sufficient to calculate a confidence interval :-)

My gut feeling tells me that the cap is not going to be the defining problem. I am thinking asynchronous bit-bang from a non-real-time OS, driver issues, hardware variants, and other stuff of that kind.

Anyway, as soon as I find some time I will try your 25ms and see if I can break it.

stefanrueger:
On Wed, Feb 23, 2022 at 21:06, Marius Greuel wrote:

| I am thinking asynchronous bit-bang from a non-real-time OS, driver issues
Nope, the 25 ms are really about the Physics of the dtr-reset circuit; the sleep happens *before* the board is reset.

| Anyway, as soon as I find some time I will try your 25ms and see if I can break it.
You can break it by unsoldering R and adding a capacitance of 2+ µF in parallel to the existing C.

stefanrueger:
| > Hence, the suggested 25 ms sleep (470 nF times 50k) is contains a large margin for a large gamut of boards. 
| I'm always a bit hesitant with that kind of statements. But I agree that it very unlikely needs a larger value.

Have you ever seen a reset circuit with more than 100 nF? Or with more than 22k? I've been designing prototypes for sensor boards with Atmel µP for a deade or so, produced mini-series for scientific projects, and studied quite a few schematics (uno, ProMini, moteino, jeenode, anarduino, Mega, ...) and cannot rememer a single one that had a non-standard reset cicuit. They may exist, but I haven't come across them. 

I have been using 5 ms for a long time in my own locally compiled copy of avrdude.

| The idea behind the long timeout value is that you have a quick wait-retry sequence initially, to get in sync with a mis-synchronized board. Once you have got a first response, you are synchronized, and then you can wait a bit longer for any responses, as normally, the expected amount of data arrives much before that timeout.
That may be, but have you ever seen stk500_getsync() catch? In my experience it's either not triggered or, if it is, it cannot recover, at least when WDT timeout of the bootloder is < 5 s. serial_recv_timeout is heavily manipulated in other programmers but not in arduino.

PR #894, opened 2022-02-28
Improve terminal read functionality

MCUdude:
This one is quite a bit simpler than the terminal write command.

Default read size has been increased to 256 bytes.
It's now possible to use the ... "operator" as a reference to the end address:

Read from address 0x10 to the end address:
`read eeprom 0x10 ...`

Dump the entire memory:
`read eeprom ...`

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -149,12 +149,10 @@ static int nexttok(char * buf, char ** tok, char ** next)
 static int hexdump_line(char * buffer, unsigned char * p, int n, int pad)
 {
   char * hexdata = "0123456789abcdef";
-  char * b;
-  int i, j;
+  char * b = buffer;
+  int32_t i = 0;
+  int32_t j = 0;
 
-  b = buffer;
-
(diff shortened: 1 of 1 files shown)


MCUdude:
@dl8dtl I know this is a fairly simple PR that doesn't bring much new functionality (other than support for [...]).
I was initially thinking about being able to read back the same data that's possible to write, but this opens a new can of worms we at least would have to discuss how to deal with.

Let's say we've ran the following command:
```
avrdude> write eeprom 0 0x11223344 0x55667788 0xaa 3.141592
```

One option would be to have some sort of "printf" formatting string that specifies how we want our data to be displayed, but I'm not sure how this could be solved in an elegant way.
```
read eeprom 0 "0x%08x 0x%08x 0x%02x %.6f"
```
 
Another option would be to use the suffixes. However, this rules out formatting and if one would want the numbers as hex og dec.
```
read eeprom 0 L L HH f
```

Issue #892, opened 2022-02-24
Pickit4 UPDI is writing at offset 0x4000 into flash instead of 0x0000.


merlinblack:
Compiled from latest source,on Linux.
Have a small test program for a ATmega4809 that I've exported the hex file from Arduino IDE.
Writing it using avrdude and a pickit4 in UPDI mode does not work as expected. The program is written but in the wrong place in flash.

Here's my command line and output:
```
#> ./avrdude -v -cpickit4_updi -pm4809 -Uflash:w:blinky.hex:i

avrdude: Version 6.99-20220221 (64cc54a)
         Copyright (c) Brian Dean, http://www.bdmicro.com/
         Copyright (c) Joerg Wunsch

         System wide configuration file is "/home/nigel/prog/avr/avrdude/build_linux/src/avrdude.conf"
         User configuration file is "/home/nigel/.avrduderc"
         User configuration file does not exist or is not a regular file, skipping

         Using Port                    : usb
         Using Programmer              : pickit4_updi
avrdude: Found CMSIS-DAP compliant device, using EDBG protocol
         AVR Part                      : ATmega4809
         RESET disposition             : dedicated
         RETRY pulse                   : SCK
         Serial program mode           : yes
         Parallel program mode         : yes
         Memory Detail                 :

                                           Block Poll               Page                       Polled
           Memory Type Alias    Mode Delay Size  Indx Paged  Size   Size #Pages MinW  MaxW   ReadBack
           ----------- -------- ---- ----- ----- ---- ------ ------ ---- ------ ----- ----- ---------
           signature               0     0     0    0 no          3    1      0     0     0 0x00 0x00
           prodsig                 0     0     0    0 no         61   61      0     0     0 0x00 0x00
           sernum                  0     0     0    0 no         10    1      0     0     0 0x00 0x00
           osccal16                0     0     0    0 no          2    1      0     0     0 0x00 0x00
           osccal20                0     0     0    0 no          2    1      0     0     0 0x00 0x00
           tempsense               0     0     0    0 no          2    1      0     0     0 0x00 0x00
           osc16err                0     0     0    0 no          2    1      0     0     0 0x00 0x00
           osc20err                0     0     0    0 no          2    1      0     0     0 0x00 0x00
           fuses                   0     0     0    0 no          9   10      0     0     0 0x00 0x00
           fuse0       wdtcfg      0     0     0    0 no          1    1      0     0     0 0x00 0x00
[...]

PR #895, opened 2022-03-01
For UPDI devices do not add offset when accessing flash.

janegilruud:
This will fix the issue with reading flash on newer AVR devices using Microchip tools and kits.


MCUdude:
Tusen takk! I'll give it a try when I'm home

dl8dtl:
should close #892

PR #903, opened 2022-03-18
Term docs

dl8dtl:
Update documentation for Hans' recent terminal mode changes.
Also add some CSS to make docs look less dull.


MCUdude:
Looks almost OK to me, but the documentation is lacking an example on how to use the [...] operator to fill the memory.

An example where [...] is used, and then a dump command right after:

```
avrdude> write eeprom 0x00 0x20 'a' 'b' 'c' 0x11 0xcafe 0x55 ...
>>> write eeprom 0x00 0x20 'a' 'b' 'c' 0x11 0xcafe 0x55 ... 
Info: Writing 32 bytes starting from address 0x00. Remaining space filled with 0x55

avrdude> dump eeprom 0 0x30
>>> dump eeprom 0 0x30 
0000  61 62 63 11 fe ca 55 55  55 55 55 55 55 55 55 55  |abc...UUUUUUUUUU|
0010  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|
```

And the examples would be even more self explanatory if the information messages added in #902 was included in the examples too.

PR #908, opened 2022-03-25
Add MacOS serial/parallel port note

### LINKED ISSUE #510: [bug #59801] macOS documentation improvement
Jeroen <jeroendevries@runbox.eu>
Sat 02 Jan 2021 02:05:38 PM UTC
Programmer hardware: Arduino Uno

Dear avrdude team,

As a macOS user, I did find it difficult to find out what the port number needs to be when flashing an arduino uno device.

I would like to suggest a improvement in the documentation.

- A header specific for macOS on the page [https://www.nongnu.org/avrdude/user-manual/avrdude_18.html#Unix-Port-Names](https://savannah.nongnu.orghttps://www.nongnu.org/avrdude/user-manual/avrdude_18.html#Unix-Port-Names)

- Along with a instruction how to find the current port name. For example
ls /dev/tty.usb*

If it is too much of a hussle I would like to edit the page myself if the addition is wanted.

Thank you for the project,

Jeroen

This issue was migrated from https://savannah.nongnu.org/bugs/?59801

MCUdude:
Closes #510

### DIFF
--- a/src/doc/avrdude.texi
+++ b/src/doc/avrdude.texi
@@ -2328,6 +2328,8 @@ configuration file will be always be @code{/etc/avrdude.conf}.
 
 @noindent
 The parallel and serial port device file names are system specific.
+MacOS has no default serial or parallel port names, but available 
+ports can be found under @code{/dev/cu.*}.
 The following table lists the default names for a given system.
 
 @multitable @columnfractions .30 .30 .30


MCUdude:
BTW the formatting in the [current PDF](https://avrdudes.github.io/avrdude/current/avrdude.pdf) manual is a bit all over the place where the text looks to be randomly indented. Is this a common way to format text rather than using newlines?

The latest HTML version also requires a lot more "user interaction" to browse through the pages not that every paragraph has its own page.

Here's an example. This is how it looked previously (with annoyingly many `[<<] [<] [Up] [>] [>>]` buttons):
https://www.nongnu.org/avrdude/user-manual/avrdude_18.html#Unix

Here's how it looks now:
https://avrdudes.github.io/avrdude/current/avrdude_22.html#Unix-Installation

Would it be possible to have the entire `A.1 Unix` chapter like previously, but without all these navigation buttons?

dl8dtl:
Seriously, I have no idea about why the formatting of the texi2html output might differ that much. After all, it's the same version of texi2html used, and about the only change between the old days and now is that I added a .css file to the translation, to make the layout optically a little more pleasant.

PR #914, opened 2022-04-01
Add terminal write string functionality

### LINKED ISSUE #913: Terminal write mode doesn't support string input (yet)
This is something I'd wanted to implement for a while, but haven't really figured out how without completely rewriting the existing implementation. The main problem is that the current implementation is designed to handle one argument at a time, but a string (with spaces in between words) will be "parsed as multiple arguments.

But maybe @dl8dtl can help bring some light into this.

This simple C example does exactly want I want it to. The space-separated string is interpreted as a single argc argument:

```c
#include <stdio.h>
#include <string.h>

int main(int argc, const char * argv[])
{
  for(int i = 0; i < argc; i++)
    printf("Arg%2d, Len %2lu, Str: %s\n", i, strlen(argv[i]), argv[i]);
}
```

Output:
```
$ ./arg_test 1234 "This is a space separated string"
Arg 0, Len 10, Str: ./arg_test
Arg 1, Len  4, Str: 1234
Arg 2, Len 32, Str: This is a space separated string
```

However, the terminal write implementation is somehow different. Here, the string (in quotes) is divided into six separate arguments.

```c
static int cmd_write(PROGRAMMER * pgm, struct avrpart * p,
		     int argc, char * argv[])
{
  for(int i = 0; i < argc; i++)
    avrdude_message(MSG_INFO,"Arg%2d, Len %2lu, Str: %s\n", i, strlen(argv[i]), argv[i]);

//...
}
```

Output:
```
avrdude> write eeprom 0 0x10 1234 "This is a space separated string"
>>> write eeprom 0 0x10 1234 "This is a space separated string" 
Arg 0, Len  5, Str: write
Arg 1, Len  6, Str: eeprom
Arg 2, Len  1, Str: 0
Arg 3, Len  4, Str: 0x10
Arg 4, Len  4, Str: 1234
Arg 5, Len  5, Str: "This
Arg 6, Len  2, Str: is
Arg 7, Len  1, Str: a
Arg 8, Len  5, Str: space
Arg 9, Len  9, Str: separated
Arg10, Len  7, Str: string"
```

If the string ended up as a single argument like the C code example shown above, implementing this would be trivial. So how can we get the same behavior in Avrdude terminal write mode as with the C code example?
Thanks!

MCUdude:
I hope it isn't too late to have this merged before the 7.0 release.
After tweaking the `nexttok` function, adding string write functionality was quite straightforward.
@dl8dtl It would be great if you could have a look.

The docs would have to be updated if this PR gets merged.

Closes #913

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -138,7 +138,6 @@ static int nexttok(char * buf, char ** tok, char ** next)
 
   if (*n) {
     *n = 0;
-    avrdude_message(MSG_INFO, "q: %s\n", q);
     n++;
   }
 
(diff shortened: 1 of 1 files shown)


MCUdude:
I've added some information about string handling in updated my [Gist documentation](https://gist.github.com/MCUdude/48141ffcb4482c3e29fb5d88fb8a6388).

dl8dtl:
Well, I'm impressed what you could squeeze out of that ancient piece of code. ;-)
Nevertheless, I feel this might still impose a small risk that something might be broken which wasn't noticed so far. Do you think we should have that in v7.0? Or rather roll a release, and then include it immediately?

MCUdude:
I'll do some extensive testing tomorrow to make sure that it doesn't affect other terminal functionality. I was hoping to get this merged before the 7.0 release if there is time.

MCUdude:
Just to make sure that the `nexttok` changes don't affect anything, I've tested all commands with my STK500 board, and everything works as expected, with no parsing issues whatsoever.

It would be great if you would give this PR a try on actual hardware just to figure out how it works, but I'm pretty confident that this is ready to be merged, as long as the documentation gets a tiny update as well.

MCUdude:
FYI I've just added string write to the terminal write example in the texi document.

Issue #386, opened 2021-12-10
[bug #45256] Raspberry Pi sysfs bug workaround for linuxgpio driver


avrs-admin:
Nick Sayer <None>
Fri 05 Jun 2015 05:10:31 AM UTC

avrdude 6.1 includes the 'inuxgpio' driver for Raspberry Pi GPIO.

There appears to be some sort of race condition bug in sysfs that results in the permissions for the gpio direction file being incorrect when "sufficient" time is not allowed for the creation process to complete. As a workaround, a small delay can be introduced between the export operations and the remaining setup.

[file #34162: patch.txt](https://savannah.nongnu.org/bugs/download.php?file_id=34162)

This issue was migrated from https://savannah.nongnu.org/bugs/?45256

mcuee:
#917 has been merged. I will close this as well.

Please re-open if you encounter similar issues. Thanks.

Issue #831, opened 2022-01-13
Port Linux sysfs GPIO feature to libgpiod


yegorich:
According to [libgpiod](https://git.kernel.org/pub/scm/libs/libgpiod/libgpiod.git/about/) description, sysfs interface is deprecated since 4.8. Hence, to enable compatibility with future kernel versions porting to `libgpiod` is required.

wutje:
Note that the linux_spi code already has the GPIO v2 infrastructure implemented.
See init code starting here
https://github.com/avrdudes/avrdude/blob/1c0b70da899628f1cad90e21cba8463a59a4744f/src/linuxspi.c#L176
and actual usage starting here:
https://github.com/avrdudes/avrdude/blob/1c0b70da899628f1cad90e21cba8463a59a4744f/src/linuxspi.c#L110

So instead of a new dependency one could extract this code (init and set GPIO) and reuse it for linux_gpio.

Just my 2 cents.

mcuee:
From here.
* https://github.com/avrdudes/avrdude/pull/917#issuecomment-1179769709

> > Moving to libgpiod may be the ultimate solution. #831
> 
> Absolutely! Now that libgpiod is a part of the linux kernel, it looks like libgpiod is available under /linux/gpio.h, which is how `linuxspi` interacts with the GPIO it uses as reset pin.
> 
> But I have no idea how to port the existing linuxgpio code over to libgpiod. Maybe the person who write linuxgpio in the first place would be interested to look into it?

Hopefully a developer will look into this.

OpenOCD libgpiod codes can be a potential reference. They have both libgpiod and sysfs based codes in the repo.
https://github.com/openocd-org/openocd/blob/master/src/jtag/drivers/linuxgpiod.c
https://github.com/openocd-org/openocd/blob/master/src/jtag/drivers/sysfsgpio.c

Issue #472, opened 2021-12-11
[bug #53279] Permission denied accessing /sys/class/gpio/gpioX/direction would leave pin exported


avrs-admin:
Michael Cain <loud.mouth.braggart@gmail.com>
Sat 03 Mar 2018 10:10:16 PM UTC
Programmer hardware: linuxgpio
Device type: RaspberryPi

When using a Raspberry Pi to program an atmega2560 using linuxgpio and you do not have the permission to access /sys/class/gpio/gpioX/direction it would leave the gpio pin exported.  The only way to unexport the pin would be to reboot the Rasbperry Pi.

The problem lies in the function linuxgpio_open, change line 280 from:
if (r < 0)
return r;

TO:
if (r < 0) {
linuxgpio_unexport(pin);
return r;
}

This issue was migrated from https://savannah.nongnu.org/bugs/?53279

mcuee:
Yes this is indeed similar to #917 and #372.
* #917 
* #372

mcuee:
I will mark this as duplicate of #386 and the fix is in #917.

Issue #372, opened 2021-12-10
[bug #44218] linuxgpio not working with raspberryPi 2. "Can't open gpioX/direction" : Permission denied. Fixed by little delay


avrs-admin:
Rizlik <rizlik@inventati.org>
Mon 09 Feb 2015 11:31:25 PM UTC

Hi,

When trying to use avrdude with my raspberryPi2 using linuxgpio programmer I got:

"Can't open gpioX/direction" : Permission denied

When stepping the code in gdb the problem is not reproduced, so I thought that the problem is that avrdude try to change direction too fast after he exported the gpio pin.

Indeed adding a little delay before trying to switch direction fixes the problem. I attach a little patch that solve the problem for me.

Same problem is shown by

$ echo 2 > /sys/class/gpio/export; echo in > /sys/class/gpio2/direction

So isn't a problem of avrdude.

Hope it helps,
rizlik

[file #33041: gpio_fix.patch](https://savannah.nongnu.org/bugs/download.php?file_id=33041)

This issue was migrated from https://savannah.nongnu.org/bugs/?44218

mcuee:
Yes #917 seems to be the same as #917 and #472.
* #472
* #917

mcuee:
I will mark this as duplicate of #386 and the fix is in #917.

PR #917, opened 2022-04-03
Fix permission denied error in linuxgpio programmer

asmirnou:
On Raspberry Pi using linuxgpio programmer without `sudo` fails with 
> Can't open gpioX/direction: Permission denied

The next attempt fails, because gpio pin remains exported and not removed. 

Open issue: #386

## Solution

After exporting gpio pin wait, until GPIO directory appears. Retry in case of EACCESS error while delayed udev permission rule applies. Call `unexport`, if `export` failed for whatever reason.

Verified that it fixes the problem.

### DIFF
--- a/src/linuxgpio.c
+++ b/src/linuxgpio.c
@@ -27,6 +27,7 @@
 #include <fcntl.h>
 #include <unistd.h>
 #include <errno.h>
+#include <sys/stat.h>
 
 #include "avrdude.h"
 #include "libavrdude.h"
@@ -108,7 +109,8 @@ static int linuxgpio_dir(unsigned int gpio, unsigned int dir)
 
   fd = open(buf, O_WRONLY);
   if (fd < 0) {
-    perror("Can't open gpioX/direction");
+    snprintf(buf, sizeof(buf), "Can't open gpio%u/direction", gpio);
+    perror(buf);
     return fd;
   }
 
@@ -137,6 +139,11 @@ static int linuxgpio_dir_in(unsigned int gpio)
  */
 
 #define N_GPIO (PIN_MAX + 1)
+
+/* Delay between checks for successful GPIO export (100ms) */
+#define GPIO_SYSFS_OPEN_DELAY      100000
+/* Number of retries to check for successful GPIO exports */
+#define GPIO_SYSFS_OPEN_RETRIES    10
 
 /*
 * an array which holds open FDs to /sys/class/gpio/gpioXX/value for all needed pins
@@ -245,6 +252,8 @@ static void linuxgpio_powerdown(PROGRAMMER *pgm)
 static int linuxgpio_open(PROGRAMMER *pgm, char *port)
 {
   int r, i, pin;
+  char gpio_path[60];
+  struct stat stat_buf;
 
   if (bitbang_check_prerequisites(pgm) < 0)
     return -1;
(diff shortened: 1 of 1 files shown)


MCUdude:
Thanks for the PR! I'm a pretty recent contributor, so @dl8dtl would have to have a look at this one.

[When searching for linuxgpio specific Avrdude issues](https://github.com/avrdudes/avrdude/issues?q=is%3Aissue+is%3Aopen+linuxgpio) it looks like there are released issues we well. You mentioned #386, but it looks to me like #472 and perhaps #372

mcuee:
This PR seems to help to avoid the issue.

```
pi@raspberrypi:~/Desktop/build/avr/avrdude_pr917/build_linux/src $ ./avrdude -C ./avrdude.conf -c linuxgpio -P GPIO -p m328p -U flash:w:./blinky_m328p.hex:i

avrdude: Device signature = 0x1e950f (probably m328p)
avrdude: erasing chip
avrdude: reading input file "./blinky_m328p.hex"
avrdude: writing flash (32768 bytes):

Writing | ################################################## | 100% 7.75s

avrdude: 32768 bytes of flash written
avrdude: verifying flash memory against ./blinky_m328p.hex:

Reading | ################################################## | 100% 6.49s

avrdude: 32768 bytes of flash verified

avrdude done.  Thank you.


```

mcuee:
But if the issue already happened (say using master without the fix), then it will not help. 
```
pi@raspberrypi:~/Desktop/build/avr/avrdude/build_linux/src $ ./avrdude -C ./avrdude.conf -c linuxgpio 
-P GPIO -p m328p -U flash:w:./blinky_m328p.hex:i
Can't open gpioX/direction: Permission denied
avrdude: opening programmer "linuxgpio" on port "GPIO" failed

avrdude done.  Thank you.

pi@raspberrypi:~/Desktop/build/avr/avrdude/build_linux/src $ sudo ./avrdude -C ./avrdude.conf 
-c linuxgpio -P GPIO -p m328p -U flash:w:./blinky_m328p.hex:i
Can't export GPIO 25, already exported/busy?: Device or resource busyavrdude: opening programmer "linuxgpio" on port "GPIO" failed

avrdude done.  Thank you.


```

mcuee:
All in all, I think this PR is good to be merged. It at least helps to avoid the issue in the first place.

Moving to libgpiod may be the ultimate solution.
* https://github.com/avrdudes/avrdude/issues/831

MCUdude:
> Moving to libgpiod may be the ultimate solution. https://github.com/avrdudes/avrdude/issues/831

Absolutely! Now that libgpiod is a part of the linux kernel, it looks like libgpiod is available under /linux/gpio.h, which is how `linuxspi` interacts with the GPIO it uses as reset pin.

But I have no idea how to port the existing linuxgpio code over to libgpiod. Maybe the person who write linuxgpio in the first place would be interested to look into it?

mcuee:
@stefanrueger 
Maybe you can take a look at this one as well to see if this can be merged as it is. It seems to improve the situation of quite a few outstanding issues.

stefanrueger (changes requested):
This is competently coded, well done.

Small comments:
```
         if (ret == 0) {
            break;
         } else if(...)
```
Whilst the code is not wrong, the `else` is not warranted, as `break` changes the flow of control. Below is arguably better readable:
```
  if(ret==0)
    break;

  if(...)
```

Using printf() a no-no. It is better to use the `avrdude_message()` system as it neatly considers the wanted verbosity and it prints to stderr, where all error and debug messages belong. I would change
```
if (retry_count > 1) {
  printf("Retrying...\n");
}
```
to
```
  if(retry_count > 1)
    avrdude_message(MSG_NOTICE2, "%s: retrying linuxgpio_dir_%s()\n", progname, i == PIN_AVR_MISO? "in": "out");
```
or something to that effect (untested).

stefanrueger:
I would be happy to merge the PR as @mcuee found it a useful improvement. The one request I have is to replace the `printf()` with an appropriate `avrdude_message()`.

stefanrueger (approved):

Issue #928, opened 2022-04-10
KT-LINK interface support


cederom:
Hello world :-)

Because ATtiny416XNANO (missing TPI protocol) nor ATtiny104XNANO (only uses DebugWIRE) that I have at hand can not be used to flash ATtiny10 that I need, I have created a definition for KT-LINK interface that is based on FT2232H chip that I also have at hand :-)

I will provide a PR in a moment :-)

Note: In order to use TPI you need to connect both `TDI` and `TDO` over a resistor to `TPIDATA`.

```
TDO --[470R]-- TPIDATA --[470R]-- TDI
```

Just not to spam the PR I will provide example runs below.

```
hexagon% uname -a
FreeBSD hexagon 13.1-STABLE FreeBSD 13.1-STABLE #0 stable/13-n250096-4f69c575996: Fri Mar 25 03:50:58 CET 2022     root@hexagon:/usr/obj/usr/src/amd64.amd64/sys/GENERIC  amd64
```

```
hexagon% avrdude -c ktlink -p t10 -B 100khz


avrdude: AVR device initialized and ready to accept instructions

Reading | ################################################## | 100% 0.00s

avrdude: Device signature = 0x1e9003 (probably t10)

avrdude done.  Thank you.
```

```
hexagon% avrdude -c ktlink -p t10 -B 100khz -vvv


avrdude: Version 6.4
         Copyright (c) Brian Dean, http://www.bdmicro.com/
         Copyright (c) Joerg Wunsch

         System wide configuration file is "/usr/local/etc/avrdude.conf"
         User configuration file is "/XXX/.avrduderc"

         Using Port                    : usb
         Using Programmer              : ktlink
         Setting bit clk period        : 10.0
I avrftdi_open(713): Using device VID:PID 0403:bbe2 and SN '(null)' on interface A.
D write_flush(478): Setting pin direction (0x0000) and value (0x0000)
I set_frequency(210): Using frequency: 100000
I set_frequency(211): Clock divisor: 0x003b
D avrftdi_check_pins_mpsse(570): Using valid mask mpsse: 0x0000fff8
D avrftdi_check_pins_bb(534): Using valid mask bibanging: 0x0000ffff
I avrftdi_pin_setup(644): Pin direction mask: e523
I avrftdi_pin_setup(645): Pin value mask: 6400
D set_pin(243): Setting pin 15 (ACBUS15) as PGMLED: high (high active)
[...]

PR #930, opened 2022-04-10
Added KT-LINK FT2232H interface with IO switching and voltage buffers.

cederom:
Example TPI run: https://github.com/avrdudes/avrdude/issues/928

Signed-off-by: Tomasz 'CeDeROM' CEDRO <tomek@cedro.info>


cederom:
Thank you @dl8dtl :-)

Issue #915, opened 2022-04-02
Some parts have wrong or missing ISP commands


stefanrueger:
`ATtiny43U`: the EEPROM memory read command has only 31 bits. It should be
```
   read = "1  0  1  0   0  0  0  0   0 0 0 x  x x x x",
          "0  0 a5 a4  a3 a2 a1 a0   o o o o  o o o o";
```
as the EEPROM has 64 bytes; `a5` is currently missing.

The lock memory misses the read instruction altogether; it should be
```
        read            = "0 1 0 1  1 0 0 0   0 0 0 0  0 0 0 0",
                          "x x x x  x x x x   o o o o  o o o o";
```

I suggest adding the following code just before the return statement in the function `parse_cmdbits(OPCODE *op)` of the grammar file `config_gram.y` so an error (or warning if one removes the rv assignment) is thrown when there are too few bits specified in a command:
```
  if(bitno > 0) {
    yyerror("Too few opcode bits in instruction");
    rv = -1;
  }
```
------------------------------------------------------------
`ATtiny87` and `ATtiny167`:

The EEPROM write commands for these parts contains two `a8` bits each; the second ones should be `a7`, respectively.

The addresses in the ISP commands actually behave more systematic than the highly flexible encoding scheme suggests, which leads to quite a few opportunities for specification errors in the `.conf` file. ISP commands only ever deal with 16-bit addresses (word addresses for flash, byte addreses for EEPROM) which appear in Bytes 2 and 3 of the 4-byte command sequence in the same sequence of bits. In fact, adding the warning
```
          // address bits in ISP commands are offset by 8 if a0 .. a15: warn if otherwise
          if(op->bit[bitno].bitno < 16 && op->bit[bitno].bitno != bitno-8)
            yyerror("Unusual bit position %d for a%d?", bitno-8, op->bit[bitno].bitno);
```
below line 1609 in `config_gram.y` uncovers these types of specification errors in `.conf`.

Speaking of the function `parse_cmdbits(OPCODE *op)` in the grammar file: line 1588 holds the assignment
```
   op->bit[bitno].bitno = 8*(bitno/8) + bitno % 8;
```
The rhs can be shortened to `bitno`, because `n%m` *is defined* as `n-m*(n/m)`. This code is supposed to provide a default address line for the lone bit signifier `a`, ie, when not accompanied by a number (as in `a11`).  However, it would be *much, much, much neater* if the default assignment was instead
```
[...]

PR #934, opened 2022-04-11
Fix Issue #915 wrong or missing SPI commands in .conf

stefanrueger:
Makes avrdude warn about SPI programming command definitions in `.conf` with too few bits or with incorrect address bits; two levels of warning exist:
 - Standard (no -v needed) if address bits miss or are wrong, and thus would make the SPI command fail
 - When -v is given also warns about superfluous address bits in the command that are unlikely to affect functionality

Ensures that a lone `a` in SPI programming commands uses the correct address bit number (depending on its position in the command). This includes `load_ext_addr` and considers the use of word addresses in flash memories and byte addresses everywhere else. The previous initialisation for lone `a` was not helpful, which may be the reason why it never could have been, and indeed never was, used in `.conf` files.

Also repairs the `avrdude.conf.in` file for the serious errors mentioned in Issue #915, most of which avrdude now warns about. 

This PR addresses two types of *human error* that have kept appearing in the `.conf` files:
  1. Wrong assignment of address lines in `a` bits or missing address bits
  2. Opening a memory section to modify parameters in a `parent` inherited part description (in fact it zaps all parameters first)

The reason why `avrdude -v` warns about superfluous address bit errors in SPI commands is that these indicate copy/paste errors that warrant a further look at the description.

Limitations of the PR:  The first type of human error is now thoroughly handled in that avrdude warns when it sees this behaviour, and that the current `.conf` file has been scrubbed of these errors. The second problem is only partially handled by this PR in that only some obvious errors affecting some 5 parts have been corrected in `.conf`.

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -9820,6 +9820,8 @@ part parent "m328"
 
     memory "efuse"
         size = 1;
+        min_write_delay = 4500;
+        max_write_delay = 4500;
         read            = "0 1 0 1 0 0 0 0 0 0 0 0 1 0 0 0",
                           "x x x x x x x x o o o o o o o o";
         write           = "1 0 1 0 1 1 0 0 1 0 1 0 0 1 0 0",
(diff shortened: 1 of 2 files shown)


MCUdude:
The formatting in avrdude.conf is a bit off in several places. Would be great if you could make sure the indentation level and spaces between numbers match the existing content or the content it replaces?

stefanrueger:
> Would be great if you could make sure the indentation level and spaces between numbers match ...

As first step, I think the PR should be reviewed. After all, it fixes serious SPI command problems of some 25% of SPI programming capable parts. Once that's done and people are happy with it *then* I am most happy to rewrite the formatting of the SPI programming commands.

BTW, what *is* the expected formatting? The current formatting of the 1371 commands in `.conf` is
  - Two lines with two strings (814 times 146 of which use `a` bits)
  - Four lines with four strings (517 times all of which using `a` bits)
  - Some of the 4-line descriptions indent the last 3 lines differently than the first line
  - Most are bit aligned/quite a few are not bit not aligned at all
  - Some come with nibble spacing/some without nibble spacing
  - Most start the command at position 9, but not all
  - Most of which align the = sign, but not all
  - Mixed use of styles happens within a description of a single part
  - Some use initial spaces in the strings, some don't
  - Some use tabs, probably intended to be 8-spaces stops, some use 4-space tabs, some spaces

Don't take my word for it, look at:

  https://github.com/avrdudes/avrdude/blob/60960ba590cdb70277f40a2a05b374d99d4e9b5c/src/avrdude.conf.in#L8648-L8661

or 

  https://github.com/avrdudes/avrdude/blob/60960ba590cdb70277f40a2a05b374d99d4e9b5c/src/avrdude.conf.in#L8848-L8854

The venerable `ATmega328` uses this style

  https://github.com/avrdudes/avrdude/blob/60960ba590cdb70277f40a2a05b374d99d4e9b5c/src/avrdude.conf.in#L9596-L9604

stefanrueger:
The vast majority (5382) of assignments happen at column position 25:
```
$ for i in {1..30}; do echo $(expand -t8 src/avrdude.conf.in | cut -c$i | grep = | wc -l) $i; done |\
  sort -nr | head -6
5382 25
886 19
719 22
577 26
410 9
292 15
```
How about I 
  - expand all tabs in `avrdude.conf.in` to 8-space stops
  - rewrite ***all*** memory SPI commands into one of the following two formats

```
        write           = "1 0 1 0 1 1 0 0   1 0 1 0 0 0 0 0",
                          "x x x x x x x x   i i i i i i i i";
```
when no address bit is used in the command, and something like
```
        loadpage_hi     = "0 1 0 0 1 0 0 0",
                          "0 0 0 x x x x x x x a5 a4 a3 a2 a1 a0",
		          "i i i i i i i i";

        writepage       = "0 1 0 0 1 1 0 0",
                          "0 0 a13 a12 a11 a10 a9 a8 a7 a6 x x x x x x",
                          "x x x x x x x x";
```
when at least one address bit is used? 

I'll make sure that after parsing the `avrdude.conf` file the *internal* parsed representation of the beautified `avrdude.conf` is identical to that of the previously used one.

Answers on a postcard, please.

stefanrueger:
Unless there are more comments I consider this PR now finished

MCUdude:
Thank you for spending the time doing all this! I'm just a contributor that just hangs around here. @dl8dtl has to review this in detail.

Issue #942, opened 2022-04-28
Treat `x` bits in SPI commands as if they were '0'


stefanrueger:
While the `x` don't care bit description in `.conf` SPI programming instructions by its very definition means that this bit *does not need* to be initialised, I suggest clearing `x` bits in `avr_set_bits()` nevertheless.

Rationale: There have been quite a few errors in `.conf` by users making mistakes in the SPI command descriptions. This intervention makes `x` bits behave like `0` and removes a potential pitfall. In case the part cares about the value of `x` bits (and, hence, the command description in `.conf` was wrong) the suggested change makes avrdude behave predictably and therefore eases debugging. Some of the 48 calls of `avr_set_bits()` set bits on uninitialised memory leaving `x` bits in a truly unknown state (mostly in `stk500.c` and `stk500v2.c`), thereby giving utility to clear the `x` bits of SPI commands in `avr_set_bits()`.

mcuee:
Will setting the other bits as zero fix this issue?
https://github.com/avrdudes/avrdude/blob/main/src/avrpart.c#L79

stefanrueger:
closed with PR #943

PR #943, opened 2022-04-28
Fix Issue #942: Treat x bits in .conf SPI commands as 0

stefanrueger:
Treats don't care bits in SPI programming commands as if they were 0. Avoids uninitialised bits if SPI command had wrongly used an x bit, and makes AVRDUDE behaviour deterministic in this case.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -31,6 +31,7 @@ Changes since version 7.0:
     - adding support for all Linux baud rates v.2 #993
     - Replace internal knowledge in jtag3.c by a public API #996
     - JTAG3 UPDI EEPROM fix #1013
+    - Treat x bits in .conf SPI commands as 0 #943
 
   * Internals:
 
--- a/src/avrpart.c
+++ b/src/avrpart.c
@@ -82,11 +82,11 @@ int avr_set_bits(OPCODE * op, unsigned char * cmd)
   unsigned char mask;
 
   for (i=0; i<32; i++) {
-    if (op->bit[i].type == AVR_CMDBIT_VALUE) {
+    if (op->bit[i].type == AVR_CMDBIT_VALUE || op->bit[i].type == AVR_CMDBIT_IGNORE) {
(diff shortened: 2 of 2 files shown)


dl8dtl:
Now that the long-wanted v7.0 is ought, I'd very much appreciate if someone put some work into a more general rework of all that. As you already mentioned, the details that were considered useful in `avrdude.conf` some 20 years ago aren't actually needed, because address bits are (almost?) always contiguous and ascending in the opcodes. So we could just use `a` in the opcode. If we then also fix the current behaviour that in an inherited device definition, a memory with the same name replaces the original one instead of extending/modifying it, we could probably simplify `avrdude.conf` quite a lot, by inheriting many memory definitions from a common ancestor.

Any takers?

stefanrueger:
Can do.

Half of the work needed is already in PR #934 (all the commits that change config_gram.y), which already allows just using `a` in the opcode whilst being backward compatible with old syntax of making the address line explicit. PR #934 now conflicts with avrdude.conf.in b/c the latter was changed independently afterwards. Making memory inherit is not all that difficult, even though it requires attention to detail and style re freeing/allocating memory, so what you suggest can be done. One needs to implement NULL assignment to opcodes in config_gram.y, so that inheriting a 128k flash memory from 256k flash memory can zap the inherited load_ext_addr opcode. 

There is a longish PR queue. To me it make sense to decide on the items in there so there is less likelihood of any work to conflict? I recommend reviewing/accepting PR #943 (this one), reviewing/accepting PR#936 and rejecting/removing #PR934 (I can recreate that).

stefanrueger:
@dl8dtl This particular PR is a small change for hardening the code against a particular user error in avrdude.conf. I suggest that I merge it in a couple of days unless there is dissent (speak now, or hold your peace forever :).

I also volunteer to simplify avrdude.conf along the ideas mentioned.

Issue #377, opened 2021-12-10
[bug #44540] Broken reset mode in jtagmkII_program_enable


avrs-admin:
Michael Lipp <mnl@mnl.de>
Sat 14 Mar 2015 01:54:50 PM UTC

I think the intended behaviour of trying to enter programming mode first without external reset and then with external reset is broken. The attached patch should fix this.

[file #33341: avrdude-progmode.patch](https://savannah.nongnu.org/bugs/download.php?file_id=33341)

This issue was migrated from https://savannah.nongnu.org/bugs/?44540

MCUdude:
I'm not able to reproduce the issue. It may be because my Dragon programmer is a bit flakey at times. I'm able to program the device over JTAG, but I'm always getting the following error at the end, and the program starts immediately.

```
avrdude: jtagmkII_program_disable(): bad response to leave progmode command: RSP_FAILED
```

mcuee:
I have not tried jtag node, but dragon_isp mode works fine with my AVR Dragon (tested with ATtiny2313, ATmega328P and ATmega32U4)
Ref: https://github.com/avrdudes/avrdude/issues/955#issuecomment-1123676668

mcuee:
The patch is copied here for easier checking.
```
*** avrdude-6.1/jtagmkII.c.orig	2015-03-14 11:28:05.991037948 +0100
--- avrdude-6.1/jtagmkII.c	2015-03-14 12:46:03.227272372 +0100
***************
*** 1194,1204 ****
--- 1194,1208 ----
    int status;
    unsigned char buf[1], *resp, c;
    int use_ext_reset;
+   unsigned char parm[1];
  
    if (PDATA(pgm)->prog_enabled)
      return 0;
  
    for (use_ext_reset = 0; use_ext_reset <= 1; use_ext_reset++) {
+     parm[0] = use_ext_reset;
+     (void)jtagmkII_setparm(pgm, PAR_EXTERNAL_RESET, parm);
+ 
      buf[0] = CMND_ENTER_PROGMODE;
      if (verbose >= 2)
        fprintf(stderr, "%s: jtagmkII_program_enable(): "
***************
*** 1223,1249 ****
        fprintf(stderr, "0x%02x (%d bytes msg)\n", resp[0], status);
      c = resp[0];
      free(resp);
!     if (c != RSP_OK) {
!       fprintf(stderr,
! 	      "%s: jtagmkII_program_enable(): "
[...]

PR #979, opened 2022-05-29
Dragon JTAG fix

### LINKED ISSUE #366: [bug #43898] atmega644p remains stopped after JTAG transaction
leres <None>
Sun 28 Dec 2014 07:42:04 PM UTC

Back when I upgraded from 5.11 to 6.0.1 I noticed that my atmega644p did not resume after programming or reading. I'm using an AVR Dragon.

I've tracked this down to revisions 1070 and 1071 of jtagmkII.c.  From the commit messages there appears to be an issue with the xmega.  The old code issued the CMND_GO when device_descriptor_length was set. The current code only isses a CMND_GO when AVRPART_HAS_PDI is set.

Attached is a patch vs 1353 that changes jtagmkII_close() to also issue CMND_GO when PGM_FL_IS_JTAG is set which gives me 5.X behavior however is likely wrong for the xmega case. I suspect the correct fix is to just avoid issuing a GO when the target is an xmega.

[file #32725: patch-jtagmkII.c](https://savannah.nongnu.org/bugs/download.php?file_id=32725)

This issue was migrated from https://savannah.nongnu.org/bugs/?43898

MCUdude:
This simple fix resolves the issue where a connected AVR JTAG target doesn't start its program after programming has occurred. Fix has been tested on real hardware.

Closes #366 and #377

### DIFF
--- a/src/jtagmkII.c
+++ b/src/jtagmkII.c
@@ -1840,8 +1840,8 @@ void jtagmkII_close(PROGRAMMER * pgm)
 
   avrdude_message(MSG_NOTICE2, "%s: jtagmkII_close()\n", progname);
 
-  if (pgm->flag & PGM_FL_IS_PDI) {
-    /* When in PDI mode, restart target. */
+  if (pgm->flag & (PGM_FL_IS_PDI | PGM_FL_IS_JTAG)) {
+    /* When in PDI or JTAG mode, restart target. */
     buf[0] = CMND_GO;
     avrdude_message(MSG_NOTICE2, "%s: jtagmkII_close(): Sending GO command: ",
 	      progname);


dl8dtl:
What confuses me a bit about that PR is the entire addition of an SPI programmer to the code.

If I see it right, the actual fix would be a one-liner, doesn't it?

MCUdude:
Whoa, sorry! It _is_ supposed to be a oneliner. I've accidentally added some linuxspi as well. I'll fix it right away!

dl8dtl:
I can confirm it fixes the Dragon issue.
On the JTAGICEmkII, the firmware eventually starts even without the patch, but with the patch applied, it's a lot faster.

PR #984, opened 2022-05-31
Revert "Ensure page_size is always at least 1"

umbynos:
This reverts commit da0e437eaa6fab3a3cc27a1425b06ed7c6f3eb97.

Reason for this revert https://github.com/ElTangas/jtag2updi/issues/16#issuecomment-781645764
We use the jtag2udpi fw on the SAMD11 to be able to program the ATMega4809. That commit was breaking the upload with AVRdude

Related conversation with @MCUdude https://github.com/umbynos/avrdude/commit/9ae854d2ae2b7046f31266b41211bf5e0a5e9bc9#commitcomment-73386063

### DIFF
--- a/src/avrpart.c
+++ b/src/avrpart.c
@@ -255,7 +255,6 @@ AVRMEM * avr_new_memtype(void)
   }
 
   memset(m, 0, sizeof(*m));
-  m->page_size = 1; // ensure not 0
 
   return m;
 }
--- a/src/config_gram.y
+++ b/src/config_gram.y
@@ -1308,13 +1308,7 @@ mem_spec :
 
   K_PAGE_SIZE       TKN_EQUAL TKN_NUMBER
     {
-      int ps = $3->value.number;
-      if (ps <= 0)
-        avrdude_message(MSG_INFO,
-                        "%s, line %d: invalid page size %d, ignored\n",
-                        infile, lineno, ps);
-      else
-        current_mem->page_size = ps;
+      current_mem->page_size = $3->value.number;
       free_token($3);
     } |
 


MCUdude:
By looking at this PR, it's not obvious to me why the config_gram.y has to be modified.
With this PR it just blindly accepts whatever `$3->value.number` holds, instead of checking for <= 0 and printing a warning.

As for the removed line in jtagmkii.c I don't know if this will have any consequences  for other hardware that uses this protocol. @dl8dtl do you have any thoughts on this one? AFAIK, `page_size` should be defined for all memory types for all targets anyways.

dl8dtl:
I'm against reverting that change – after all, it has been made for a reason (fixing a bug or crash).
There must be a better solution, but I didn't look into the details so far.

mcuee:
In this case, we can close this one.

mcuee:
@umbynos Can you ask the affected user to try out latest avrdude 7.0 release or even git master to see if the issue has been fixed? I think there are a few improbements done on this area.

MCUdude:
I can try this with my Nano Every later today!

mcuee:
Re-open this topic, just want to check if the issue has been fixed in avrdude already or not yet.

mcuee:
I do not have issues using avrdude-7.0 and git with Arduino Nano Every (with SAM D11 as the JTAG2UPDI controller). Another user also reported the susccess story in another issue.
Ref:
* https://github.com/avrdudes/avrdude/issues/980
* https://github.com/avrdudes/avrdude/issues/991
*

mcuee:
@umbynos 

In this case, I think this issue can really be closed. Any objections?

stefanrueger:
Looks like AVRDUDE 7.0 does not expose the issue that this PR tries to solve, so not reverting.

PR #985, opened 2022-06-04
adding support for some fast baud rates

prchal:
If optiboot can work at higher bauds, why not avrdude.

### DIFF
--- a/src/ser_posix.c
+++ b/src/ser_posix.c
@@ -77,6 +77,15 @@ static struct baud_mapping baud_lookup_table [] = {
 #ifdef B230400
   { 230400, B230400 },
 #endif
+#ifdef B250000
+  { 250000, B250000 },
+#endif
+#ifdef B500000
+  { 500000, B500000 },
+#endif
+#ifdef B1000000
+  { 1000000, B1000000 },
+#endif
   { 0,      0 }                 /* Terminator. */
 };
 


dl8dtl:
I see you have already closed it.
Anyway, just so someone else who might find it by some Internet search can see it: you can specify arbitrary baudrates with the `-b` option. The table is only used for operating systems that don't have a 1:1 mapping between the baud rate number and the `Bxxxx` definition.

prchal:
On 04. 06. 22 18:37, Jörg Wunsch wrote:
> I see you have already closed it.
Closed it? Accidentally. I'm new to github, sorry.

> Anyway, just so someone else who might find it by some Internet search 
> can see it: you can specify arbitrary baudrates with the |-b| option. 
> The table is only used for operating systems that don't have a 1:1 
> mapping between the baud rate number and the |Bxxxx| definition.
Well, in ubuntu it didn't work until I add them to the table. It cost me 
1 day trying all rates until I discovered, that it's avrdude fault why 
it's not communicating: I set -b 500000 with some error, after while I 
keep -b 500000 but optiboot flashed 115200 and it worked! So I had a 
look to source code, add to table and wow, it's OK now!
It's sad that avrdude took baud rates and not set them to hw without any 
error.
Tested on CH340G.

prchal:
So is there any way to reopen it?

MCUdude:
@prchal I see you've added the additional baud rates @dl8dtl requested, but for some reason that commit didn't end up in this PR. Please see if you fix this so this PR can be merged. Thanks!

prchal:
Sorry, completely lost in github, please guide me or feel free to copy - paste fix.

MCUdude:
I'm not _that_ good with git either, but @per1234 is. It looks like your last commit only referenced this PR instead of actually committing to it. I'm not sure what to do. Let's wait a little and see if other users know what to do without closing this one and creating a new PR.

dl8dtl:
I don't fully understand it either. @prchal did you force-push your commit? I can only see a single commit in the history.
Force-pushing is a bad idea for that, because from Github's point of view, there's now nothing new that happened …

dl8dtl:
Suggestion:

There's a comment
`/* There are a lot more baud rates we could handle, but what's the point? */`

This one is going to be outdated by your change. Edit it to something more meaningful, like
```/* These are all the known Bxxxx values for Linux.
  * Most other OSes can pass the baud rate without Bxxx values. */```
and push that change. Then, Github sees a new change, and will update the pull request.

MCUdude:
Closed in favor of #993

per1234:
>Force-pushing is a bad idea for that, because from Github's point of view, there's now nothing new that happened …

GitHub has no problem with force pushes. It even clearly indicates what happened. Although force pushes should never be done to a production branch, it is a common practice to force push to a PR branch in order to squash "fix up" commits to ensure an atomic commit history.

Here is an example of how GitHub handles force pushes: https://github.com/avrdudes/avrdude/pull/908

image

PR #993, opened 2022-06-16
adding support for all Linux baud rates v.2

prchal:
If optiboot can work at higher bauds, why not avrdude.
Versoin 2 of #985.
Linux uses the old-style bitmapped version of the Bxxxx macros. Most other OSes can pass the baud rate without Bxxx values.


MCUdude:
Looks good to me!

PR #1018, opened 2022-07-04
Fix Issue #995 ft245r paged read for ATmega2560 et al

stefanrueger:
Issue #995 fixed. 

Untested. @mcuee Could I trouble you for testing? (I first need to wire/solder up my FT245r chip on the USB BUB ii module and then create a programmer entry for this in my `~/.avrduderc`).

The bulk of the PR is refactoring the logic of paged read and write (the latter worked, but used hard-coded SPI programming bytes as opposed to the SPI ops that are defined in `avrdude.conf`). Unfortunately, this means reviewing will need a deep dive into the `ft245r.c` programmer. I am convinced, though, that the paged r/w functions are now crisper and clearer.

(@mcuee I predict that testing might fail for byte-wise flash writes in the terminal. This is outside the scope of this PR and independent from the underlying issue. I noticed the problem whilst reading `avr.c` and `term.c`. In fact, I predict flash byte writes in the terminal will fail for quite a few programmers and quite a few parts. I'll raise an issue in the next few days.)

### DIFF
--- a/src/ft245r.c
+++ b/src/ft245r.c
@@ -107,10 +107,10 @@ void ft245r_initpgm(PROGRAMMER * pgm) {
 
 #else
 
-#define FT245R_CYCLES	2
-#define FT245R_FRAGMENT_SIZE  512
-#define REQ_OUTSTANDINGS	10
-//#define USE_INLINE_WRITE_PAGE
+#define FT245R_CYCLES          2
+#define FT245R_CMD_SIZE       (4 * 8*FT245R_CYCLES)
+#define FT245R_FRAGMENT_SIZE  (8 * FT245R_CMD_SIZE)
+#define REQ_OUTSTANDINGS      10
 
 #define FT245R_DEBUG	0
 /*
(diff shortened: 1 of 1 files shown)


mcuee:
> (@mcuee I predict that testing might fail for byte-wise flash writes in the terminal. This is outside the scope of this PR and independent from the underlying issue. I noticed the problem whilst reading avr.c and term.c. In fact, I predict flash byte writes in the terminal will fail for quite a few programmers and quite a few parts. I'll raise an issue in the next few days.)

I can confirm that as well with either the ft232r or the usbasp. Wiring bootloader seems to work fine.

But I will wait for your issues and fixes (if any) to carry out more testing.

stefanrueger:
Thanks for testing, @mcuee. Great to know the PR works!

Will wait a couple of weeks for a review, but merge after that time if no review comes forward.

MCUdude:
@ mcuee can you recommend a decent ft232 based programmer i can purchase for testing purposes? I do have an FTDI basic board, but this is only suited for regular UART.

mcuee:
@MCUdude 

Right now I use something like this. You should be able to find similar things from Amazon or Aliexpress.
https://i.stack.imgur.com/fIXQV.jpg

mcuee:
@MCUdude 
I do not have FT245R based board yet but I plan to buy one.

FT232H or FT2232H based programmer will be faster than FT232R based. You can get those breakout board as well. I had the original JTAGkey2 courtesy from Amontec many years ago due to my involvement in OpenOCD. There are many similar things out there.

stefanrueger:
... and finally I can test my PR. And conform it works with my own rig.

stefanrueger:
Merging on the basis that the code has been tried and tested at several sites, and solves a problem.

PR #1023, opened 2022-07-10
Fix PICKit2 ATmega2560 flash paged flash read

stefanrueger:
The paged read in pickit2.c has two errors:
  - It drops load extended address commands unless a paged read happens at a
    64k byte boundary; this is invalid when reading files with holes
  - It wrongly assumed that flash memory is byte addressed

The fix is to carry out a load extended address command, if needed, at the
beginning of each paged flash read with the correct word address. Although the
pickit2_paged_load() has independent parameters page_size, addr and n_bytes,
AVRDUDE only ever calls paged read/write functions with page_size and n_bytes
both set to mem->page_size and addr aligned with a page boundary. Therefore, it
is sufficient to set the load extended address at the beginning of each page
read.

### DIFF
--- a/src/pickit2.c
+++ b/src/pickit2.c
@@ -491,18 +491,15 @@ static int  pickit2_paged_load(PROGRAMMER * pgm, AVRPART * p, AVRMEM * mem,
 
     pgm->pgm_led(pgm, ON);
 
+    if (lext) {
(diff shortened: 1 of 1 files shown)


mcuee (approved):
The change is good. I have tested with PICKit 2 and ATmega2560 and the change works.

PR #1033, opened 2022-07-18
Ignore `-s` flag as safemode is no longer supported

### LINKED ISSUE #1032: Unknown option '-s' on command line
The so-called "safe mode" feature was removed in #859 and with it the option parsing for the `-s` flag.  That causes problems with scripts calling avrdude after upgrading to version 7.0.  A well-behaved command line program should accept, but ignore, deprecated options instead of choking on them.

MCUdude:
Resolves #1032

### DIFF
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -44,9 +44,7 @@
 .Op Fl O
 .Op Fl P Ar port
 .Op Fl q
-.Op Fl s
 .Op Fl t
-.Op Fl u
 .Op Fl U Ar memtype:op:filename:filefmt
 .Op Fl v
 .Op Fl x Ar extended_param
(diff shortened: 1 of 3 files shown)


acolomb:
Thanks for the quick fix!

Shouldn't these still be mentioned in `src/doc/avrdude.texi` and `src/avrdude.1` as deprecated options?  Just dropping every note about them has greater potential for confusion I would think.

dl8dtl (approved):

dl8dtl:
Mentioning them in avrdude.1 and avrdude.texi might be fine, though I really think their only merit is to cater old scripts. Scripts usually don't read documentation anyway. ;-)

acolomb:
Good argument ;-) Just thought I'd mention it in case nobody had thought about the docs yet.

stefanrueger:
I like @acolomb's considered suggestion on documentation (never hurts).

MCUdude:
I've updated the documentation! `-s` and `-u` are now mentioned.

stefanrueger (approved):
Perfect. Good to go.

PR #1034, opened 2022-07-18
Cache config_file components in AVRPART and PROGRAMMER structures

stefanrueger:
Some 90% of the space of AVRPART and some 50% of PROGRAMMER is occupied by a 4 kB array config_file[] that contains the configuration file name. In preparation of developer options that output a raw dump of the part descriptions, this commit changes the config_file components from a large array, which is duplicated in each part and programmer description, to a cached string for each config file allowing for smaller raw dumps.

This commit also changes the config file name to its realpath(), eg, shortens unwarranted `/bin/../etc/` file name components. For example, on Linux a syntax error in the config file would no longer look like
```
avrdude: error at /usr/local/bin/../etc/avrdude.conf:9583: syntax error
avrdude: error reading system wide configuration file "/usr/local/bin/../etc/avrdude.conf"
```
Applying this PR makes the error message print
```
avrdude: error at /usr/local/etc/avrdude.conf:9583: syntax error
avrdude: error reading system wide configuration file "/usr/local/etc/avrdude.conf"
```

It also changes the global variable names `infile` and `lineno` to cfg_infile and cfg_lineno for an ever so slight improvement of code clarity.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -852,11 +852,11 @@ int main(int argc, char * argv [])
         port = DEFAULT_USB;
         break;
 
-#ifdef HAVE_LINUXSPI
       case CONNTYPE_SPI:
-        port = *default_spi ? default_spi : "unknown";
-        break;
+#ifdef HAVE_LINUXSPI
+        port = *default_spi? default_spi: "unknown";
 #endif
+        break;
     }
   }
 


stefanrueger:
Looks like realpath() is not available on some systems:

  * msvc (x86, Win32) cancelled (1 annotation)
  * msvc (x64, x64) cancelled (2 annotations)
  * msvc (arm64, ARM64) failed (1 annotation)

Ideas?

mcuee:
@stefanrueger 
From here:
https://stackoverflow.com/questions/45124869/cross-platform-alternative-to-this-realpath-definition

So looks like you can use the alternative from MSVC here.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fullpath-wfullpath?view=msvc-170

mcuee:
It is also not available under MinGW. ` _fullpath` is the suggested fix for simple use case.
Ref: https://sourceforge.net/p/mingw/patches/256/

mcuee:
With the change I can build under MSVC and MSYS2 mingw64.

```
C:\work\avr\avrdude_test\avrdude_sr [config_file ≡ +1 ~1 -0 !]> git diff
diff --git a/src/config.c b/src/config.c
index 4308407..952f0db 100644
--- a/src/config.c
+++ b/src/config.c
@@ -32,6 +32,13 @@

 #include "config_gram.h"

+#if defined(_MSC_VER) || defined(__MINGW32__)
+    #ifndef PATH_MAX
+    #define PATH_MAX _MAX_PATH
+    #endif
+    #define realpath(N,R) _fullpath((R),(N),_MAX_PATH)
+#endif
+
 char default_programmer[MAX_STR_CONST];
 char default_parallel[PATH_MAX];
 char default_serial[PATH_MAX];

C:\work\avr\avrdude_test\avrdude_sr [config_file ≡ +1 ~0 -0 !]> .\out\build\x64-Release\src\avrdude.exe -v

avrdude.exe: Version 7.0-20220718 (f95a1d3)
             Copyright (c) Joerg Wunsch

             System wide configuration file is "C:/work/avr/avrdude_test/avrdude_sr/out/build/x64-Release/src/avrdude.conf"


avrdude.exe: no programmer has been specified on the command line or the config file
             Specify a programmer using the -c option and try again

/c/work/avr/avrdude_test/avrdude_sr
$ ./build_mingw64_nt-10.0-19043/src/avrdude.exe -v

avrdude.exe: Version 7.0-20220718 (f95a1d3)
             Copyright (c) Joerg Wunsch

             System wide configuration file is "C:/work/avr/avrdude_test/avrdude_sr/build_mingw64_nt-10.0-19043/src/avrdude.conf"


avrdude.exe: no programmer has been specified on the command line or the config file
             Specify a programmer using the -c option and try again
```

stefanrueger:
> [`_fullpath()`] With the change I can build under MSVC and MSYS2 mingw64.

Brilliant tip. I moved the realpath() replacement macro in `config.h`, though it's probably better placed into `msvc/msvc_compat.h` just below the definition of PATH_MAX. @mariusgreuel ?

mcuee:
> > [`_fullpath()`] With the change I can build under MSVC and MSYS2 mingw64.
> 
> Brilliant tip. I moved the realpath() replacement macro to `config.h`, though it's probably better placed into `msvc/msvc_compat.h` just below the definition of PATH_MAX.  

Not a good idea as MinGW also does not have `realpath()`, so it is not only MSVC.

stefanrueger:
> Not a good idea to put under `msvc/msvc_compat.h` as MinGW also does not have `realpath()`, so it is not only MSVC.

Would it be reasonable, feasible to introduce similar compatibility files `mingw/mingw_compat.[ch]` to which we can move compiler/OS specific items? I notice we already have  `src/serbb_posix.c`,  `src/serbb_win32.c`,  `src/ser_posix.c` and `src/ser_win32.c`. Long term some sort of consolidation might be desirable (though low prio for me).

mcuee:
The situation for `src/serbb_posix.c, src/serbb_win32.c, src/ser_posix.c and src/ser_win32.c` are probably a bit different as the serial API is quite different between POSIC and WIN32. 

But I agree that it is a good idea to add `mingw/mingw_compat.[ch]` in the long run.

dl8dtl (commented):
Well, occasionally I'd love to see a few more spaces, but otherwise, it's fine with me.

Issue #1041, opened 2022-07-22
Writing fusebits to the MCU from a file does not working


masya-chel:
I am trying to write fusebits to the MCU from a file, this function does not working - an error is displayed. See the screenshots below.

![изображение](https://user-images.githubusercontent.com/87937592/180478100-463dedb7-73dc-472a-992b-6bb55fd827b0.png)

![изображение](https://user-images.githubusercontent.com/87937592/180476115-595c22dc-e9cd-41a0-9f4a-9713e0e1cfeb.png)

`avrdude.exe: Device signature = 0x1e9307 (probably m8)`
`avrdude.exe: reading input file "C:\CLI\AVRdude_CLI\fw_write\high.txt"`
`avrdude.exe: fileio: invalid operation=0`
`avrdude.exe: read from file 'C:\CLI\AVRdude_CLI\fw_write\high.txt' failed`

If I try to read fusebits from the MCU to a file, then all works correctly - files with the names "high" and "low" are created in the folder, and the value of fusebits is contained inside the files. See the screenshots below.

![изображение](https://user-images.githubusercontent.com/87937592/180476187-b6a8cf23-cf67-42ba-b9ee-0cdf532fa4f8.png)

![изображение](https://user-images.githubusercontent.com/87937592/180478520-385d4342-e428-4647-87fc-6324f75ec31e.png)

If I try to write fusebits to the MCU directly from the console, then fusebits are written correctly. See the screenshot below.

![изображение](https://user-images.githubusercontent.com/87937592/180476221-ffe950be-aa13-4890-b81a-121631a93331.png)

Tested on avrdude versions: 6.3; 6.4; 7.0.
Tested on microcontrollers: ATtiny13, ATmega8.
OS: Windows 7 x64.

dl8dtl:
> If I try to write fusebits to the MCU directly from the console, then fusebits are written correctly.

That suggests that something in your other (GUI, whatever it might be) execution environment must be strange.

However, what really puzzles me:

> `avrdude.exe: fileio: invalid operation=0`

`FIO_READ` *is* 0, but why does the switch/case instruction falls into the `default` clause?

I'm afraid noone outside your GUI environment might be able to reproduce that.

dl8dtl:
My only other idea: is it really sure that both invocations (GUI and commandline) are calling the exact same binary?

[...]

PR #1042, opened 2022-07-23
Handle invalid -U file format specifiers for input

dl8dtl:
The file format specifiers 'h', 'd', 'o', and 'b' are only valid for
outputting data. Reject them with a proper error message when
attempting to use them for input.

Supposed to fix issue #1041


stefanrueger:
Does that not need to be considered for device verify as well (also file read)?

Can I suggest changing fileio.c instead of update.c. I think the misleading error message comes from 
https://github.com/avrdudes/avrdude/blob/affe4cb50a42c34a873be9bde8b5e650032cdcb5/src/fileio.c#L1278-L1285
Changing the error message to the effect of file read with formats `:d`, `:h`, `:o`, `:b` not supported would do the trick. Also, if someone were to implement the file read for these formats they'd do it here and needn't worry about update.c

dl8dtl:
> Does that not need to be considered for device verify as well (also file read)?
> 

Prpbably, yes.

> Can I suggest changing fileio.c instead of update.c.

The corresponding message for write attempts to 'm' was already there, so I did about the same.

But we could also move everything to fileio.c.

dl8dtl:
OK, I moved all checks to `fileio.c`.

stefanrueger (approved):
Perfect - good to go.

stefanrueger (changes requested):
I just realised that the check for `FMT_IMM` actually better be kept in `update.c`:
  - Write immediate is *impossible* (same reason why there is no `sti` assembler command)
  -  Check in `update.c` is earliest possible and prevents the memories from being read (and then not being able to write which was known before)

It is definitely good to *also* check for `FMT_IMM` in `fileio.c` considering the desired future library nature of the code (which might be called independent of update code).

It is good that the :h, :d, ... format checks happen in `fileio.c` and not in`update.c` as it is the *current* implementation in fileio.c that does not allow reading these formats (there is nothing in principle that prevents this from happening, and update should be oblivious of this).

The single suggested change to the current PR is to keep `src/update.c` as it was.

stefanrueger (approved):

Issue #969, opened 2022-05-17
A few niggling issues in `stk500.c`


stefanrueger:
The last error message in `stk500_loadaddr()` says

 https://github.com/avrdudes/avrdude/blob/4601bee4afa76fb4aa8392ffdd4a4dd56ede1bae/src/stk500.c#L785-L791 

It should say **stk500_loadaddr()** expected **Resp_STK_OK**, not that loadaddr() expected Resp_STK_INSYNC.

Similarly wrong error messages (again copy/paste relicts) exist in `stk500_paged_write()` and in `stk500_getparm()`.

---------

In `stk500_getsync()` the return value of `stk500_recv(pgm, resp, 1)` is not checked; this means that `resp[0]` might not have been set, so `stk500_getsync()` might act as if in sync when it's not:

 https://github.com/avrdudes/avrdude/blob/4601bee4afa76fb4aa8392ffdd4a4dd56ede1bae/src/stk500.c#L123-L126

---------

Issue #942 mentions other occasions in `stk500.c` when variables are not (fully) initialised before use.

---------

Not doing a PR yet as there is a bit of a queue already (including PR #943 that would solve Issue #942) and I might see more small issues wanting to be mended.

mcuee:
I think both findings (error message and `stk500_getsync()`) are valid.

MCUdude:
@stefanrueger now that PR #942 has been merged, this issue might be ready to be worked on.

Apart from the issues mentioned above, the most serious one appears to be the one mentioned in #942

> (@stefanrueger) Rationale: There have been quite a few errors in .conf by users making mistakes in the SPI command descriptions. This intervention makes x bits behave like 0 and removes a potential pitfall. In case the part cares about the value of x bits (and, hence, the command description in .conf was wrong) the suggested change makes avrdude behave predictably and therefore eases debugging. Some of the 48 calls of avr_set_bits() set bits on uninitialised memory leaving x bits in a truly unknown state (mostly in stk500.c and stk500v2.c), thereby giving utility to clear the x bits of SPI commands in avr_set_bits().

mcuee:
I believe #1046 will be merged soon to fix this issue.

stefanrueger:
Closed by PR #1046

Issue #421, opened 2021-12-10
[bug #48131] Problem with EEprom address of avrdude for different processors!


avrs-admin:
kh_kuebbeler <None>
Sat 04 Jun 2016 06:06:37 PM UTC

I have found out, that the address of EEprom load with the stk500v1 protocol is send as word address for the target ATmega328 and as a byte address for the target ATmega8.
I have build a bootloader from the optiboot source in assembly language with additional EEprom support, which match in a 512 byte page.
If I handle the transmitted address different for the ATmega328 and
the ATmega8, all works well. For the ATmega328 I must multiply the
transmitted address by 2.
The problem is already reported at [https://www.mikrocontroller.net/topic/398978](https://savannah.nongnu.orghttps://www.mikrocontroller.net/topic/398978) in german language. Joerg Wunsch has this information already.
I can handle the different address type in my bootloader version,
if I know, which processors use the byte address (mega8, mega16, mega32, ...). But what is with other transfer programs as avrdude?
The AVR061 from Atmel does not specify the address type of the
Cmnd_STK_LOAD_ADDRESS command.

This issue was migrated from https://savannah.nongnu.org/bugs/?48131

avrdude-admin:
Joerg Wunsch <joerg_wunsch>
Sat 04 Jun 2016 06:29:52 PM UTC

The problem is apparently that piece of code in stk500.c:

```
if ((m->op[AVR_OP_LOADPAGE_LO]) || (m->op[AVR_OP_READ_LO]))
a_div = 2;
else
a_div = 1;

```
Later on, a_div is used to determine whether the address in
the load command needs to be divided by 2:

```
stk500_loadaddr(pgm, m, addr/a_div);

```
(Similar code exists in both, stk500_paged_write() as well as
stk500_paged_load()).

I think the intention was to divide the load address by 2 when
accessing flash.  Now, the ATmega328 entry (as well as others)
has an AVR_OP_LOADPAGE_LO instruction even in the EEPROM memory
section.  Thus, load addresses are halved for it when accessing
EEPROM.

If the only decision needed is whether flash or EEPROM is going
to be accessed, my suggestion is to combine that with the 'F' and
'E' decision a few lines before:

```
if (strcmp(m->desc, "flash") == 0) {
memtype = 'F';
a_div = 2;
}
else if (strcmp(m->desc, "eeprom") == 0) {
memtype = 'E';
a_div = 1;
}
else {
return -2;
}

```

[...]

PR #1046, opened 2022-07-24
Deprecate original STK500 v1 protocol in favour of optiboot and Arduino as ISP

stefanrueger:
Closes Issue #969 A few niggling issues in stk500.c
Closes Issue #967 Do STK500/arduino/xbee programmers fail EEPROM r/w for some parts?

 - Initialise memory before avr_set_bits() calls in stk500.c and stk500v2.c
 - Check stk500_recv() actually worked before accepting SYNC byte
 - Consolidate error messages for stk500.c
 - Resolve signed/unsigned comparisons in stk500.c and stk500v2.c
 - Deprecate original STK500 v1 protocol in favour of optiboot and Arduino as ISP
    
For paged read/write early AVRDUDE implementations of the STK500 v1 protocol communicated a word address (below a_div=2) or byte address (a_div=1) based on the following code irrespective of which memories were used:
```
  if(m->op[AVR_OP_LOADPAGE_LO] || m->op[AVR_OP_READ_LO])
    a_div = 2;
  else
    a_div = 1;
```
This turned out to be a bug: it really should have been a_div=2 for flash and a_div=1 for eeprom. At the time presumably no one noted because Atmel was at the cusp of replacing their FW 1.x with FW 2 (and the STK500 v2 protocol).

It seems that the world (optiboot, Arduino as ISP, ...) has compensated for the bug by assuming AVRDUDE sends *all* eeprom addresses as word addresses. Actually these programmers overcompensated for the bug because for six out of the 146 known SPI programmable parts with eeprom and page size > 1, AVRDUDE would still send the eeprom addresses as byte addresses (ATmega8 ATmega8A ATmega64 ATmega64A ATmega128 ATmega128A) owing to above code.

It makes no sense to correct the bug now seeing that virtually no one uses the old 2005 STK 500 v1 firmware. This commit now follows optiboot, Arduino as ISP and other projects, and simply sends all addresses for paged read or write as word addresses. There are no longer (little known) exceptions for ATmega8 et al that surprised some optiboot etc users.

### DIFF
--- a/src/stk500v2.c
+++ b/src/stk500v2.c
@@ -991,6 +991,7 @@ static int stk500v2_chip_erase(PROGRAMMER * pgm, AVRPART * p)
   buf[0] = CMD_CHIP_ERASE_ISP;
   buf[1] = p->chip_erase_delay / 1000;
   buf[2] = 0;	// use delay (?)
+  memset(buf+3, 0, 4);
   avr_set_bits(p->op[AVR_OP_CHIP_ERASE], buf+3);
   result = stk500v2_command(pgm, buf, 7, sizeof(buf));
   usleep(p->chip_erase_delay);
@@ -1121,8 +1122,8 @@ retry:
   buf[5] = p->bytedelay;
   buf[6] = p->pollvalue;
   buf[7] = p->pollindex;
+  memset(buf+8, 0, 4);
   avr_set_bits(p->op[AVR_OP_PGM_ENABLE], buf+8);
-  buf[10] = buf[11] = 0;
 
   rv = stk500v2_command(pgm, buf, 12, sizeof(buf));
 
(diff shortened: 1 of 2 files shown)


stefanrueger:
Also closes Issue #421

src/stk500v2.c
```diff
@@ -2549,6 +2553,7 @@ static int stk500v2_paged_load(PROGRAMMER * pgm, AVRPART * p, AVRMEM * m,
             progname, p->desc);
     return -1;
   }
+  memset(cmds, 0, sizeof cmds);
```
MCUdude, line 2556: Perhaps use `sizeof(cmd)` instead of `sizeof cmd` to match the style in the rest of the file?

MCUdude (approved):
Looks good to me! Works well with ATmega8 running Optiboot. I can read and write to EEPROM just fine, something I couldn't do without this PR

stefanrueger:
@MCUdude Thanks for sterling review. Looks like it can be merged

Issue #1037, opened 2022-07-20
design issue: jtag2updi is hardcoded to fail for parts with "has_pdi = yes"


ElTangas:
The jtag2updi programmer fails to recognize parts with PDI interface i.e. "has_pdi = yes".

In particular this impacts avrdude.conf files where, for compatibility with old avrdude versions, UPDI parts are defined with "has_pdi = yes" instead of "has_updi = yes".

Simple experiment: create a new programmer in avrdude.conf, named jtag2xxx, which is an exact copy of jtag2updi definition except for the name.
Now, UPDI parts with a "has_pdi = yes" definition work just fine with this programmer.

This means jtag2updi was special cased, hardcoded, inside avrdude, to fail for "has_pdi = yes". I fail to see what useful purpose this serves. In truth, I consider poor programming practice to special case user defined programmers from the avrdude.conf file, such as jtag2updi, inside avrdude itself.

In understand this was introduced here: https://github.com/avrdudes/avrdude/pull/789
I submit that the behaviour of the arduino port of avrdude, accepting both "has_pdi = yes" and "has_updi = yes" for all programmers using jtagmkii_pdi as base is preferable.

Note: using either "has_pdi = yes" or "has_updi = yes" actually generates different behaviour.
"has_updi = yes" -> performs complete chip erase before programming, uses page write for programming.
"has_pdi = yes" -> does not perform complete chip erase before programming, uses page erase/write for programming.
However both work correctly and both may be useful in specific situations, I don't advocate in any way fixing this "issue"! Therefore I think jtag2updi should not be precluded from using "has_pdi = yes".

MCUdude:
Well, I'm the one to blame here. I added Arduino's UPDI patch to `jtagmkii.c`, and a primitive check to make sure jtag2updi would print a reasonable error message when used with an ATxmega part. I didn't think of the backward compatibility error, since I've previously only used Arduino's patched version of Avrdude. Sorry for that.

This is the part causing the issue:
https://github.com/avrdudes/avrdude/blob/affe4cb50a42c34a873be9bde8b5e650032cdcb5/src/jtagmkII.c#L1308-L1314

Is there a different way we can detect ATxmegas and print a reasonable error message?

ElTangas:
Why is it necessary to detect xmegas when using jatg2updi?
They will generate an error down the line anyway, even the wiring for PDI is different from UPDI. I don't think anyone would actually connect jtag2updi to a xmega.

[...]

PR #1048, opened 2022-07-26
Add `jtagmkii_updi` programmer option

MCUdude:
Resolves issue #1037.

This PR adds a new programmer type that all future jtag2updi implementations should use. However, it also allows "old" avrdude.conf files to be used with the jtag2updi programmer, which has been around way before Avrdude supported UPDI at all.

From @ElTangas:
> The jtag2updi programmer fails to recognize parts with PDI interface i.e. "has_pdi = yes".
  
> In particular this impacts avrdude.conf files where, for compatibility with old avrdude versions, UPDI parts are defined with "has_pdi = yes" instead of "has_updi = yes".

> Simple experiment: create a new programmer in avrdude.conf, named jtag2xxx, which is an exact copy of jtag2updi definition except for the name.
Now, UPDI parts with a "has_pdi = yes" definition work just fine with this programmer.

> This means jtag2updi was special cased, hardcoded, inside avrdude, to fail for "has_pdi = yes". I fail to see what useful purpose this serves. In truth, I consider poor programming practice to special case user defined programmers from the avrdude.conf file, such as jtag2updi, inside avrdude itself.


mcuee:
Comments here by ElTangas
* https://github.com/avrdudes/avrdude/issues/1037#issuecomment-1196505359

MCUdude:
> I've tested your patch, and it now works with both old and new versions of avrdude.conf, which is a plus IMO.
This ordeal has also exposed a subtle bug in jtag2updi that I'm correcting now.

> However, I'd like to propose a change to your error detection logic as follows

```  /* Abort and print error if programmer does not support the target microcontroller */
 (original) if ((strncmp(pgm->type, "JTAGMKII_UPDI", strlen("JTAGMKII_UPDI")) == 0 && p->flags & AVRPART_HAS_PDI) ||
 (proposal) if ((strncmp(pgm->type, "JTAGMKII_UPDI", strlen("JTAGMKII_UPDI")) == 0 && !(p->flags & AVRPART_HAS_UPDI)) ||
```
> Because whether a part has PDI or not is not the real issue. The issue is the absence of UPDI and that is what should be checked for.
For example, a future part might have both PDI and UPDI interfaces, and then we would be hunting bugs again - your logic would see PDI and immediately fail, even if UPDI is also supported.

Thanks for your input @Eltangas, I've applied your proposed fix!

PR #1078, opened 2022-08-20
Use const for programmer functions where useful

stefanrueger:
This PR updates the AVRDUDE programmer interface:
 - Use `const` in the 40 or so PROGRAMMER functions where appropriate; this to keep developers who create a new programmer on the straight and narrow to
     + Enforce `cookie` is used for allocated data (and not some "unused" entry in the AVRPART or PROGRAMMER structure)
     + Document which of the programmer function do not (should not) change AVRPART, AVRMEM or PROGRAMMER structures (note that buf and tags contents can still be modified as they are pointers)
 - Change `void (*enable)(PROGRAMMER * pgm)` to `void (*enable)(PROGRAMMER *pgm, const AVRPART *p)`; this allows changes in the PROGRAMMER structure after the part is known (think UPDI, PDI, ... that were not around when the programmer interface was first conceived)

In order to get meaningful `const` properties for the PROGRAMMER, AVRPART and AVRMEM arguments, some code needed to be moved around, otherwise a network of "tainted" assignments risked rendering not much `const`:
 - Change `void (*enable)(PROGRAMMER * pgm)` to `void (*enable)(PROGRAMMER *pgm, const AVRPART *p)`; this allows changes in the PROGRAMMER structure after the part is known. For example, use TPI, UPDI, PDI functions in that programmer appropriate to the part. This used to be done later in the process, eg, in the `initialize()` function, which "taints" all other programmer functions wrt `const` and sometimes requires other finessing with flags etc. Much clearer with the modified `enable()` interface.
 - Move TPI initpgm-type code from `initialize()` to `enable()` --- note that `initpgm()` does not have the info at the time when it is called whether or not TPI is required.
 - `buspirate.c`: move pgm->flag to PDATA(pgm)->flag (so legitimate modification of the flag does not change PROGRAMMER structure)
 - Move `AVRPART_INIT_SMC` and `AVRPART_WRITE` bits from the `flags` field in AVRPART to jtagmkII.c's private data `flags32` field as `FLAGS32_INIT_SMC` and `FLAGS32_WRITE` bits
 - Move the `xbeeResetPin` component to private data in `stk500.c` as this is needed by xbee when it saddles on the `stk500` code (previously, the `flags` component of the part was re-dedicated to this)
 - Change the way the "chained" private data are used in `jtag3.c` whilst keeping the PROGRAMMER structure read-only otherwise
 - In `stk500v2.c` move the STK600 initpgm update from `stk500v2_initialize()` to `stk500v2_enable()` so the former keeps the PROGRAMMER structure read-only (for `const` assertion). Whilst I know that this remains functionally the same, I did not understand initialisation conditions in the first place and still don't (always a sign of unnecessary complexity, but there you go)
 - In `usbasp` change the code from changing PROGRAMMER functions late to dispatching to TPI or regular SPI protocol functions at runtime; reason being the decision whether to use TPI protocol is done at run-time depending on the capability of the attached programmer

### DIFF
--- a/src/buspirate.c
+++ b/src/buspirate.c
@@ -76,14 +76,14 @@ struct pdata
 	unsigned char pin_dir;		/* Last written pin direction for bitbang mode */
 	unsigned char pin_val;		/* Last written pin values for bitbang mode */
 	int     unread_bytes;		/* How many bytes we expected, but ignored */
+        int     flag;
 };
 #define PDATA(pgm) ((struct pdata *)(pgm->cookie))
 
 /* ====== Feature checks ====== */
 static inline int
-buspirate_uses_ascii(struct programmer_t *pgm)
-{
-	return (pgm->flag & BP_FLAG_XPARM_FORCE_ASCII);
+buspirate_uses_ascii(const PROGRAMMER *pgm) {
+	return (PDATA(pgm)->flag & BP_FLAG_XPARM_FORCE_ASCII);
 }
 
 /* ====== Serial talker functions - binmode ====== */
@@ -105,8 +105,7 @@ static void dump_mem(const int msglvl, const unsigned char *buf, size_t len)
 		avrdude_message(msglvl, "\n");
 }
 
-static int buspirate_send_bin(struct programmer_t *pgm, const unsigned char *data, size_t len)
-{
+static int buspirate_send_bin(const PROGRAMMER *pgm, const unsigned char *data, size_t len) {
 	int rc;
 
 	avrdude_message(MSG_DEBUG, "%s: buspirate_send_bin():\n", progname);
@@ -117,8 +116,7 @@ static int buspirate_send_bin(struct programmer_t *pgm, const unsigned char *dat
 	return rc;
 }
 
-static int buspirate_recv_bin(struct programmer_t *pgm, unsigned char *buf, size_t len)
-{
+static int buspirate_recv_bin(const PROGRAMMER *pgm, unsigned char *buf, size_t len) {
 	int rc;
 
 	rc = serial_recv(&pgm->fd, buf, len);
@@ -131,12 +129,12 @@ static int buspirate_recv_bin(struct programmer_t *pgm, unsigned char *buf, size
 	return len;
 }
 
-static int buspirate_expect_bin(struct programmer_t *pgm,
+static int buspirate_expect_bin(const PROGRAMMER *pgm,
 				unsigned char *send_data, size_t send_len,
 				unsigned char *expect_data, size_t expect_len)
 {
 	unsigned char *recv_buf = alloca(expect_len);
-	if ((pgm->flag & BP_FLAG_IN_BINMODE) == 0) {
+	if ((PDATA(pgm)->flag & BP_FLAG_IN_BINMODE) == 0) {
 		avrdude_message(MSG_INFO, "BusPirate: Internal error: buspirate_send_bin() called from ascii mode\n");
 		return -1;
 	}
@@ -148,7 +146,7 @@ static int buspirate_expect_bin(struct programmer_t *pgm,
 	return 1;
 }
 
-static int buspirate_expect_bin_byte(struct programmer_t *pgm,
+static int buspirate_expect_bin_byte(const PROGRAMMER *pgm,
 					unsigned char send_byte, unsigned char expect_byte)
 {
 	return buspirate_expect_bin(pgm, &send_byte, 1, &expect_byte, 1);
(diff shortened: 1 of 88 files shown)


mcuee:
This is a big change so that I am not so sure how to carry out tests -- I think this requires more on the code review side and not so much on the testing using physical programmers.

Just one minor thing, the comment change to `avrdude.conf.in` does not seem to belong to here.

mcuee:
I carried out some simple tests using stk500v2 and wiring bootloader and there seems to be no regressions.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1078.exe -c stk500v2 -P COM4 -B1 -qqp m2560
 -U .\hex\blink-mega2560_lext-test.hex && echo OK
OK

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1078.exe -c stk500v2 -P COM4 -B1 -qqp m2560
 -U .\hex\stk500boot_v2_mega2560_eeprom_fix.hex && echo OK
OK

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1078.exe -c wiring -P COM3 -qqp m2560
 -D -U .\hex\blink-mega2560_lext-test.hex && echo OK
OK

```

src/buspirate.c
```diff
@@ -76,14 +76,14 @@ struct pdata
 	unsigned char pin_dir;		/* Last written pin direction for bitbang mode */
 	unsigned char pin_val;		/* Last written pin values for bitbang mode */
 	int     unread_bytes;		/* How many bytes we expected, but ignored */
+        int     flag;
```
MCUdude, line 79: @stefanrueger  I'm looking through the code to see if things look alright. Would be great if you would fix the indentation on this line (buspirate.c, line 79)!

MCUdude:
Thanks for your work @stefanrueger! I looked though the whole thing, and I don't see anything that's worrying me.

I've tested this PR using a USBasp, USBtinyISP, PICkit4 in ISP and UPDI mode, SerialUPDI and jtag2updi, and it looks like there's no breaking changes for these programmers.

```
$ ./avrdude -cusbasp -patmega1284p -Uflash:w:blink_1284p.hex:i -qq && echo OK
OK

$ ./avrdude -cavrispmkii -patmega1284p -Uflash:w:blink_1284p.hex:i -qq && echo OK
OK

$ ./avrdude -cusbtiny -patmega1284p -Uflash:w:blink_1284p.hex:i -qq && echo OK
OK

$ ./avrdude -cpickit4_isp -patmega1284p -Uflash:w:blink_1284p.hex:i -qq && echo OK
OK

$ ./avrdude -cserialupdi -P/dev/cu.usbserial-1410 -patmega4808 -Uflash:w:blink_4808.hex -qq && echo OK
avrdude: UPDI link initialization OK
avrdude: NVM type 0: 16-bit, page oriented write
avrdude: Entering NVM programming mode
avrdude: Leaving NVM programming mode
OK

$ ./avrdude -cserialupdi -P/dev/cu.usbserial-1410 -patmega4808 -Uflash:w:blink_4808.hex -qq && echo OK
avrdude: UPDI link initialization OK
avrdude: NVM type 0: 16-bit, page oriented write
avrdude: Entering NVM programming mode
avrdude: Leaving NVM programming mode
OK

$ ./avrdude -cpickit4_updi -patmega4808 -Uflash:w:blink_4808.hex -qq && echo OK
         Vtarget                      : 5.06 V
         JTAG clock megaAVR/debug     : 100 kHz
         PDI/UPDI clock Xmega/megaAVR : 100 kHz

OK

```

stefanrueger:
Thanks for comments and tests, @MCUdude and @mcuee. Well appreciated. 

I only have Linux, so would be great to compile under Windows/MacOS after enabling appropriate options that are normally off in https://github.com/avrdudes/avrdude/blob/38aa1313f94fb2ece97ea9a87d5384e8b7cb2b89/CMakeLists.txt#L30-L36

I already tested compilation under Linux with what I thought could be added under Linux:
```
option(HAVE_LINUXGPIO "Enable Linux sysfs GPIO support" ON) 
option(HAVE_LINUXSPI "Enable Linux SPI support" ON) 
option(HAVE_PARPORT "Enable parallel port support" ON)
```

src/buspirate.c
stefanrueger: Ahh, I used 8 spaces to align, not a tab, so hopefully will look OK in source? Git diff probably thinks a tab is 4 spaces.

MCUdude: Well, it will depend on the text editor settings. I personally prefer indents to be two characters,  so in the case above it won't look right. It's probably best to either replace all tabs in the struct with spaces, or use tab instead to match the rest of the code.

stefanrueger: OK, good point, fixed.

Normally I try to adapt spacing of small changes to existing code (which I failed to do here).

[For new code segments I will normally use my own "handwriting" --- even at the risk of a source file becoming inconsistent in style --- but I rarely change spacing/style of existing code lest reviews and later commit diffs see too many lines without actual content change.]

stefanrueger:
Unless there are objections, I'll merge this PR soon-ish

PR #1095, opened 2022-09-04
CMake: Fix dependency chain between avrdude.conf.in and avrdude.conf

### LINKED ISSUE #1035: CMake: avrdude.conf dependency to avrdude.conf.in is broken
My workflow during development is like this: The first time I check out a branch I do
```
$ rm -rf build_linux
$ cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -B build_linux
$ cmake --build build_linux
```
After changes when I want to test, it's repeatedly the following, which nicely compiles and installs the to-be-tested avrdude:
```
cmake --build build_linux; sudo cmake --build build_linux --target install
```

However, I noticed that the systemwide `avrdude.conf` does not get installed/changed with the latter command even after explicitly changing `src/avrdude.conf.in`. It gets installed with the first-time creation of the build_linux dir.

What I'd like is a dependency of the systemwide conf file on the source conf.in. How do I effect that?

mariusgreuel:
Fixes #1035

This PR corrects the dependency chain between avrdude.conf.in and avrdude.conf,
i.e. avrdude.conf is rebuilt when avrdude.conf.in changes.

Also, a new target 'conf' has been added for building 'avrdude.conf'.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -282,27 +282,6 @@ endif()
 add_subdirectory(src)
 
 # =====================================
-# Setup default port names
-# =====================================
-
-if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
-    set(DEFAULT_PAR_PORT "/dev/parport0")
-    set(DEFAULT_SER_PORT "/dev/ttyS0")
-elseif (CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
-    set(DEFAULT_PAR_PORT "/dev/ppi0")
-    set(DEFAULT_SER_PORT "/dev/cuad0")
-elseif (CMAKE_SYSTEM_NAME STREQUAL "Solaris")
-    set(DEFAULT_PAR_PORT "/dev/printers/0")
-    set(DEFAULT_SER_PORT "/dev/term/a")
-elseif (CMAKE_SYSTEM_NAME STREQUAL "Windows")
-    set(DEFAULT_PAR_PORT "lpt1")
-    set(DEFAULT_SER_PORT "com1")
-else()
-    set(DEFAULT_PAR_PORT "unknown")
-    set(DEFAULT_SER_PORT "unknown")
-endif()
-
-# =====================================
 # Configuration
 # =====================================
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -82,29 +82,51 @@ else()
 endif()
 
 # =====================================
+# Setup default port names
+# =====================================
+
+if (CMAKE_SYSTEM_NAME STREQUAL "Linux")
+    set(DEFAULT_PAR_PORT "/dev/parport0")
+    set(DEFAULT_SER_PORT "/dev/ttyS0")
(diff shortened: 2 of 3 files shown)


stefanrueger:
I can confirm this works now as expected (and is vvv useful for semi-automated rewrites of avrdude.conf.in).

While we are at correcting dependencies: I have had a problem recently with the top-level `CMakeLists.txt` (not the one in `src`) in that when I changed `option(HAVE_LINUXGPIO "Enable Linux sysfs GPIO support" OFF)` from `OFF` to `ON` and back some of the code was compiled with `OFF` and some with `ON`. I believe I rebuilt with `cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -B build_linux` in between.

Given https://github.com/avrdudes/avrdude/blob/dab99c082382159522770fa1e9f91a7c4aa1c38b/src/libavrdude.h#L412-L418 the PROGRAMMER structure with a pin array of `N_PINS` elements had different sizes in different compilation units. Catastrophic hard to debug errors ensued, not surprisingly, if one function thinks there are 32 pins and the other function thinks there are 401 pins. Anyway, I tried reproducing this but failed (maybe there was more involved than switching back and forth between `OFF` and `ON`, eg, aborting make build with ctrl-c, I can only speculate here). I don't know my way around `cmake`, so cannot even help by looking at the source for the build process whether this was my problem or a real problem in a missing dependency on `CMakeLists.txt`

mariusgreuel:
> I can confirm this works now as expected

Thank you!

> when I changed `option(HAVE_LINUXGPIO "Enable Linux sysfs GPIO support" OFF)`

Instead of modifing the options in the source, you should reconfigure CMake by passing the options on the command-line, i.e. run `cmake -D HAVE_LINUXGPIO=1 -B build_linux`

Not sure what is supposed to happen when you change an option in `CMakeLists.txt`. My guess is that it still uses its options from its cache, so it does not rebuild at all.

However, that does not explain the problem you had (only some objects being recompiled). The good news is that you may be able to troubleshoot the problem yourself, as CMake typically generates plain makefiles by default (otherwise use `-G "Unix Makefiles"`). You can run `make VERBOSE=1`, `make -n`, and all the good stuff. You can also pass the make options via cmake, for instance run `cmake --build build_linux -- VERBOSE=1`.

Assuming you have CMake generate makefiles, I suspect that make is to blame. When I configure the project using makefiles, I can tell that the dependencies on the generated `ac_cfg.h` file (which holds the HAVE_LINUXGPIO #define) are correct (look at build_linux/src/CMakeFiles/libavrdude.dir/depend.make). Assuming this define changes, you should get one or the other, but not a mixture. So much for the theory.

That said, make uses timestamps, not hashes, to figure out what needs to rebuild. If you save a source file in the middle of the build, I think you can break it.

If you ever observe such behavior again, perhaps you can tar the entire avrdude directory, and then I will take a look.

stefanrueger:
@mariusgreuel Thanks for detailed theoretical analysis - OK will use a different way of specifying build options.

mcuee:
I think this pull request can be merged now.

PR #1103, opened 2022-09-25
saving errno of ioctl() call in a temporary variable to prevent overwrite

s-wakaba:
Thank you for accepting PR #1100 !
The purpose of this PR is bugfix to avoid unintended overwirte for `errno` variable.


stefanrueger:
@s-wakaba I don't see how this PR makes a difference. Does calling strerror(var) really  change var? Do you have an example of different behaviour with and without this PR?

stefanrueger:
OK, I read up upon the printf() family used in avrdude_message(), and it can set errno, indeed, in which case this PR protects the ioctrl errno setting from being overwritten. I would be hard pressed to construct a situation for this particular avrdude_message()  where printf() sets errno and is perfectly usable afterwards, though. 

In summary: I consider the PR valid and good-practice coding.

mcuee:
Great. Then this PR can be merged.

PR #1109, opened 2022-10-05
Add missing AVR variants the original JTAG ICE supports

MCUdude:
Variants with identical signatures as their main part will work with JTAG ICE.

I've removed ATmega165 from this PR. AVR Studio 4 states that the ATmega165 isn't supported, but various JTAG ICE clones on the internet state that they support ATmega165. This should be properly tested and confirmed first.

I have the parts I need to build a JTAG ICE, and IIRC I do have an ATmega165A and ATmega165P I can test with. Give me a few days, and I should have it figured out.

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -3828,7 +3828,6 @@ part
 part parent "m64"
     desc                   = "ATmega64A";
     id                     = "m64a";
-    prog_modes             = PM_SPM | PM_ISP | PM_HVPP | PM_JTAG;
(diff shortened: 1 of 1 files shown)


mcuee:
Looks good to me. I do not have the ATmega165 and ATmega165 variant to test though.

MCUdude:
Thanks for the hint @stefanrueger, that was a really neat process! I rebased the PR, and now there's just no duplicate data that would otherwise be inherited!

Issue #1027, opened 2022-07-13
Switch a PICkit4 or MPLAB SNAP programmer to AVR mode


MCUdude:
Brand new PICkit4 and MPLAB SNAP programmers are running in "PIC mode" by default, and can't program AVRs unless you download MPLAB X, create a project with an AVR, connect your programmer and _then_ attempt to connect to the target MCU. It would be very convenient if Avrdude could detect a PICkit4 or a SNAP in PIC mode, and set it to AVR mode instead of using MPLAB X.

From the discussion in #1015:

> (@MCUdude) Out of curiosity. To convert a PK4 or a SNAP from PIC to AVR mode is this just a simple command, or is the tool reflashed by MPLABX/MS7? If it's just a simple command, this is something that would be neat to either implement here or create a simple FOSS CLI tool that does just this.

> (@janegilruud) It's "just a simple" command. Yes, I agree there should be an option outside the IDEs.

> (@MCUdude) Great, that means it's possible for tools like Avrdude to "quickly" switch to AVR mode. My dream would be something like this:
```
 $ avrdude -c pickit4_updi -p atmega4809

avrdude: Detected PICkit4 programmer in PIC mode. Do you want to switch to AVR mode? (Y/n)
Y
avrdude: PICkit4 programmer switched to AVR mode
avrdude: Please reconnect the programmer for the changes to take effect

avrdude done.  Thank you.
```

mcuee:
A USB traffic capture (when running MPLAB X) should help to reverse engineering the mode switching logic from PIC mode to AVR mode.

mcuee:
Also for old PICKit 4/SNAP, it is said that you may still need MPLAB X to upgrade the FW.
Ref: https://bloom.oscillate.io/docs/mplab-snap-avr-mode

The USB PID changes from PIC mode to AVR mode, so I believe FW change is involved.

janegilruud:
A challenge with a brand new PICkit4 straight from the factory is that the FW it contains is so old that it doesn't have the AVR mode. That is why a FW upgrade is necessary the first time you want to set a new PICkit4 to AVR mode.

Since Microchip Studio 7.0.2594 both Studio and MPLAB X can change mode and update FW on PICkit4 and Snap.

mcuee:
I will try to get the SNAP. It is not as good as PICKit 4 but it is probably good enough for me.

mcuee:
[...]

PR #1137, opened 2022-10-19
Detect PICkit4 and SNAP in PIC mode

### LINKED ISSUE #1135: To supress the warning for Microchip SNAP
Right now there is always a warning messge `avrdude_git.exe: usbhid_open(): No device found` when you use Microchip SNAP.

I believe the reason is that the current code loops through the three USB PIDs (`0x217F, 0x2180, 0x2181`). Because it can not find the first one (`0x217F`), it prints out the warning message.

@MCUdude 
Just wondering where you get the info that SNAP has three USB PIDs. Thanks. Do you see the same warning message for Microchip PICKit 4?

From your pull-request that both SNAP and PICKit 4 will have three USB PIDs.
* https://github.com/avrdudes/avrdude/pull/791/files

I am thinking maybe there is only one USB PID for the AVR mode. The other is for PIC mode and yet the other one is probably for FW update.

MCUdude:
This PR makes it possible for Avrdude to detect a PICkit4 or SNAP programmer in PIC programming mode, and prints a reasonable error that this (currently) has to be changed over to AVR mode using either MPLAB X or Microchip Studio.

Hopefully, @janegilruud will have time to look into #1027 someday, which is the ultimate solution to the entire "PIC programming mode problem". However, this PR makes it obvious to the USER why their brand new PICkit4/SNAP doesn't work as expected.

Avrdude output:
```

avrdude warning: MPLAB SNAP in PIC mode detected!
        Use MPLAB X or Microchip Studio to switch to AVR mode

avrdude error: unable to open programmer snap_updi on port usb

avrdude done.  Thank you.
```

#1027 related

Resolves #1135


MCUdude:
I totally screwed up some rebasing, and I can't fire out how to fix it. I'm opening a new PR

PR #1138, opened 2022-10-20
Detect PICkit4 and SNAP in PIC mode

### LINKED ISSUE #1135: To supress the warning for Microchip SNAP

MCUdude:
This PR makes it possible for Avrdude to detect a PICkit4 or SNAP programmer in PIC programming mode, and prints a reasonable error that this (currently) has to be changed over to AVR mode using either MPLAB X or Microchip Studio.

Hopefully, @janegilruud will have time to look into #1027 someday, which is the ultimate solution to the entire "PIC programming mode problem". However, this PR makes it obvious to the USER why their brand new PICkit4/SNAP doesn't work as expected.

Avrdude output:
```

avrdude warning: MPLAB SNAP in PIC mode detected!
        Use MPLAB X or Microchip Studio to switch to AVR mode

avrdude error: unable to open programmer snap_updi on port usb

avrdude done.  Thank you.
```

#1027 related

Resolves #1135


stefanrueger (commented):
Cool PR, thanks for this.

One comment: I am not keen on the open device function to comment failure with the notice2 level in some instances and error in others. IMO this should be either warnings/errors throughout or nothing/notice2 shifting the responsibility to the caller for communication.

https://github.com/avrdudes/avrdude/pull/1138/files#diff-c7816f81017dc84e71af87907727ebcf157b7b206bf75dec9bd0b176e6f5aaa3

Note: there is a pre-existing case with return -1 and *no* message. Whilst looking at this we should probably deal with this case in the same way, too.

MCUdude:
> One comment: I am not keen on the open device function to comment failure with the notice2 level in some instances and error in others. IMO this should be either warnings/errors throughout or nothing/notice2 shifting the responsibility to the caller for communication.

Sure! It's just that we don't really need to see that message when not in verbose mode. but `msg_warning` is probably OK.

> Note: there is a pre-existing case (line 97) with return -1 and no message. Whilst looking at this we should probably deal with this case in the same way, too.

I'll look into it. And I'll see if I can replace tabs with spaces in a separate commit. The formatting in usb_hidapi.c is a bit all over the place.

mcuee:
Tested with a PICkit 4 borrowed from local Microchip office.

```
C:\work\avr\avrdude_test\avrdude_bin> ./avrdude -c pickit4_updi -patmega4808
avrdude warning: USB device with VID: 0x03eb and PID: 0x2177 not found
avrdude warning: USB device with VID: 0x03eb and PID: 0x2178 not found
avrdude warning: USB device with VID: 0x03eb and PID: 0x2179 not found

avrdude error: PICkit4 in PIC mode detected!
        Use MPLAB X or Microchip Studio to switch to AVR mode

avrdude error: unable to open programmer pickit4_updi on port usb

avrdude done.  Thank you.

```

MCUdude:
Cool! The PR works as it should then 👍

mcuee:
> Cool! The PR works as it should then 👍

Yes it works perfectly.

PR #1154, opened 2022-10-23
Revert "Add readline for MSYS2 mingw32/mingw64 build"

stefanrueger:
Reverts avrdudes/avrdude#1148


mcuee:
Looking at the build log (https://github.com/avrdudes/avrdude/actions/runs/3308684165/jobs/5461248027), the issue is MSYS2 readline/termcap packaging issues mentioned here.
* https://github.com/msys2/MINGW-packages/issues/8105

On my local machine I can fix the issue by rebuilding termcap as per the
tips from MSYS2 github issue. 
* https://github.com/msys2/MINGW-packages/issues/8105#issuecomment-1229270428

I'll see what I can do to correct this issue using github action.

Issue #1155, opened 2022-10-24
Readline support for mingw


mcuee:
I created PR #1148 to add readline for mingw32/64 github action build. Unfortunately it caused problem and had to be reverted.
* #1148 

Looking at the failed build log (https://github.com/avrdudes/avrdude/actions/runs/3308684165/jobs/5461248027), the issue is still the MSYS2 readline/termcap packaging issues mentioned here.
* https://github.com/msys2/MINGW-packages/issues/8105

On my local machine I can fix the issue by rebuilding termcap as per the tips from MSYS2 github issue. 
* https://github.com/msys2/MINGW-packages/issues/8105#issuecomment-1229270428

I'll see what I can do to correct this issue using github action. I am not good at github action scripts. Hopefully someone more experienced in github action can help as well.

mcuee:
Some trial and error here, not so successful so far.
https://github.com/mcuee/avrdude/tree/readline_mingw_1

mcuee:
The best I can come out with is to use dynamic link for libreadline. 
https://github.com/mcuee/avrdude/blob/readline_mingw_1/.github/workflows/build.yml#L317

mcuee:
BTW, if your MSYS2 mingw installation has readline included, git main will fail the build.

It has two issues with the default build as the default build use static link.
1) missing libtermcap.a
2) even adding libtermcap.a it will still fail due to the above MSYS2 issue.

Solution is to modify termcap and add libtermcap.a in the build.

```
$ ./build.sh
-- Building for: Ninja
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: C:/msys64/usr/bin/git.exe (found version "2.37.2")
-- Found FLEX: C:/msys64/usr/bin/flex.exe (found version "2.6.4")
-- Found BISON: C:/msys64/usr/bin/bison.exe (found version "3.8.2")
-- Looking for libelf.h
-- Looking for libelf.h - found
-- Looking for libelf/libelf.h
-- Looking for libelf/libelf.h - found
-- Looking for usb.h
-- Looking for usb.h - not found
-- Looking for lusb0_usb.h
-- Looking for lusb0_usb.h - found
-- Looking for libusb.h
-- Looking for libusb.h - not found
[...]

PR #1158, opened 2022-10-24
Add readline and shared build for mingw -- WIP DNM

mcuee:
1) Add readline dependancy
2) Use dynamic link in order to avoid readline MSYS2 issue
https://github.com/msys2/MINGW-packages/issues/8105

This is probably not the right fix. Please DO NOT MERGE now.

Reference:
* https://github.com/avrdudes/avrdude/issues/1155


mcuee:
This is not the right fix and I will close this one.

PR #1174, opened 2022-11-10
Ensure default_programmer from config files is used

stefanrueger:
A quick PR. The `default_programmer` variable in `avrdude.conf` is not used. This PR sets `programmer` to the config file's default programmer in case the user had not specified one.


mcuee:
I agree this PR is a good change to merge.

Issue #1152, opened 2022-10-23
Terminal write's automated width for numbers can be surprising


MCUdude:
First reported by @dbuchwald  in #1111.

There seems to be a bug in terminal write mode when base 10 number greater than 127 is written. For some reason, a trailing 0x00 is also written to memory. See output below. Tested on ATmega4808 and AVR128DA32 using a SerialUPDI programmer.

```
$ ./avrdude -cserialupdi -patmega4808 -t

avrdude: AVR device initialized and ready to accept instructions

Reading | ################################################## | 100% 0.02s

avrdude: device signature = 0x1e9650 (probably m4808)
avrdude> read eeprom
>>> read eeprom 

Reading | ################################################## | 100% 0.07s

0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0010  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0020  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0030  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0040  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0050  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0060  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0070  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0080  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0090  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write eeprom 0x00 0
[...]

PR #1175, opened 2022-11-10
Make terminal write's automatic number width less surprising

stefanrueger:
Supposed to fix Issue #1152 

@MCUdude  and @dbuchwald please have a look whether this asymmetric range [int_min, **u**int_max] is less surprising. As a  side effect, this PR will therefore be backwards compatible with the write of yonder days when all write could do was writing single bytes.


MCUdude:
Thanks for the PR @stefanrueger! I gave it a try, and _I_ think the behavior is less surprising with this PR than without.

PR #1181, opened 2022-11-16
Add fallback for EM_AVR

dl8dtl:
OpenBSD still lacks this definition, so provide our own for them.


mcuee (approved):
Reviewed and tested to be good.

mcuee:
Yes this is good and can be merged.

PR #1182, opened 2022-11-18
Rewrite authors/copyrights

dl8dtl:
Rewrite the list of authors and copyright message.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,6 +2,13 @@ AVRDUDE was written by:
 
 	Brian S. Dean  <bsd@bdmicro.com>
 
+AVRDUDE is currently maintained by:
+
+	Hans Eirik Bull
+	Stefan Rueger
+	Xiaofan Chen
+	Joerg Wunsch
+
 Contributors:
 
 	Joerg Wunsch <j@uriah.heep.sax.de>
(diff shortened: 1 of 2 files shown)


MCUdude:
Thanks for the PR!

My only thought is that referring to a file called AUTHORS may not make sense where the user has just been handed a binary (Arduino installation, etc.). How about providing a link to the AUTHORS file on GitHub as well? Just a thought.

mcuee:
Thanks a lot for the PR and thanks for adding me in the team.

stefanrueger:
Thanks, I think this is a brilliant team.

dl8dtl:
> Thanks for the PR!
> 
> My only thought is that referring to a file called AUTHORS may not make sense where the user has just been handed a binary (Arduino installation, etc.). How about providing a link to the AUTHORS file on GitHub as well? Just a thought.

Replacing it by an URL is a brilliant idea. I already thought about how to include the full list e.g. with `-vv`, but using the URL is much more convenient.

Pushed.

Let's seen what Brian says to all this. As it modifies the copyright display, I'd like to obtain his consent as well.

bd34n:
Looks great to me!

PR #1184, opened 2022-11-19
Print HAVE_LIBREADLINE status

mcuee:
Readline is now required for better functionality when available. It is better to print out the status during CMake configure stage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -361,6 +361,12 @@ else()
     message(STATUS "DON'T HAVE libftdi1")
 endif()
 
+if(HAVE_LIBREADLINE)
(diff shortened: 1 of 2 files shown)


MCUdude:
Thanks for the PR!
Maybe you could add this to configure.ac as well?

mcuee:
> Thanks for the PR! Maybe you could add this to configure.ac as well?

Auto-tools are more difficult for me, but I will try.

mcuee:
@MCUdude 
It turns out the change to `configure.ac` is also simple. Please give it a try.

mcuee:
Take note thie PR will still treat macOS system bundled libedit as having Readline. It can be safely merged even after the merge of PR #1185.

PR #1185, opened 2022-11-20
Add a workaround for issue #1173

dl8dtl:
The readline compat layer for libedit which otherwise provides native readline functionality in NetBSD and MacOS appears to have a bug that causes subsequent prompts to be delayed until another line is being processed.

Re-installing the callback over and over again works around that, and does not seem to harm with GNU libreadline.

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -1290,6 +1290,15 @@ void term_gotline(char *cmdstr) {
         term_running = 0;
     }
     free(cmdstr);
+    /*
+     * This is a workaround for a bug apparently present in the
+     * readline compat layer of libedit which is natively present in
+     * NetBSD and MacOS.
+     *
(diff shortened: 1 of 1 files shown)


mcuee:
Tested to be good under macOS (Mac Mini M1).

MCUdude:
Thanks for the PR! Works perfectly with (legacy) macOS 10.14 libedit

MCUdude:
@mcuee do you think the build instructions for macOS should be updated? The user doesn't really need to install libreadline anymore

mcuee:
> @mcuee do you think the build instructions for macOS should be updated? The user doesn't really need to install libreadline anymore

Wiki updated.
https://github.com/avrdudes/avrdude/wiki/Building-AVRDUDE-for-macOS

PR #1189, opened 2022-11-22
Deprecate mosi/miso in favour of sdo/sdi

stefanrueger:
See https://www.oshwa.org/a-resolution-to-redefine-spi-signal-names/

We are actually a bit late to the party (https://www.washingtonpost.com/opinions/2020/06/12/tech-industry-has-an-ugly-master-slave-problem/)

The functionality of AVRDUDE is unchanged.


MCUdude:
One problem is that MISO/MOSI tells us where which device the data is coming from and where it goes to. SDI/SDO does not, and a user could easily think that SDI should be connected to SDO and vice versa when two devices are speaking SPI (like with UART). This will for sure cause confusion.

MISO/MOSI has not been replaced with SDI/SDO in any Microchip AVR datasheet I'm aware of.
Would it be an idea to keep MOSI/MISO, but refer to them as host and client instead? Every schematic I'm aware of with a classic AVR uses MISO and MOSI as names for their programming pins.

It would be interesting to hear @dl8dtl's opinion on this.

dl8dtl:
I would prefer PDI and PDO, because that's more descriptive for our purpose: program data in/out. There is historical precedence even, since the ATmega103 did not use the SPI pins for programming (but the UART ones), so they were labelled PDI/PDO. ATmega128 inherited that (as did ATmega1281 and ATmega128A, still sold today).

However, there's potential for confusion as we also have a programming interface that is called PDI. :-/

So SDI/SDO is the least common denominator for me.

stefanrueger:
There is also POCI and PICO (P=peripheral/C=controller), which is the Open Source hardware association suggests as drop-in replacement for MISO and MOSI. However SDO/SDI *are* standard signal names for SPI interfaces, they are similar in spirit and name to @dl8dtl's historically used PDO/PDI, and what I really like are centred around the device (so I do not have to imagine the role of the device).

mcuee:
> MISO/MOSI has not been replaced with SDI/SDO in any Microchip AVR datasheet I'm aware of.
> Would it be an idea to keep MOSI/MISO, but refer to them as host and client instead? Every schematic I'm aware of with a classic AVR uses MISO and MOSI as names for their programming pins.

As a HW engineer, I agree with @MCUdude. Changing MISO/MOSI to SDI/SDO will confuse many avrdude users.

stefanrueger:
> But it is good to mention in the beginning of the documentation that they are traditionally called MISO/MOSI

done now in man page and tex source

stefanrueger:
The errors of the github runners are unrelated to the PR:

remote: app=lariat msg="[FATAL]: Internal error" 
remote: aborting due to possible repository corruption on the remote side.
fatal: protocol error: bad pack header
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
==> Homebrew is run entirely by unpaid volunteers. Please consider donating:
  https://github.com/Homebrew/brew#donations

mcuee:
@stefanrueger 
Yes sometimes this temporary errors happen. I just triggered the rerun of the failed macOS build and now it is okay.

As for the this pull request, I am okay with the changes now.

Issue #1162, opened 2022-10-30
To add arduino-packing github action to create binary for Linux, macOS and Windows


mcuee:
I think @umbynos has done a good job in [avrdude-packing project](https://github.com/arduino/avrdude-packing) to be able to build static linking version for Linux, macOS and Windows (mingw32). It may be a good idea to adopt it as an addition to the current github actions.

mcuee:
For Linux and macOS, only one patch is needed. The only question is whether it is acceptable for avrdude project or not.
https://github.com/arduino/avrdude-packing/blob/main/patches/0006-Cmake-add-libtinfo-and-libncurses-as-dependencies.patch

I think as it is it can not be accepted because it will not work with MSVC build.

@mariusgreuel or other CMake experts may be able to figure out a way to modify it to be suitable to be adopted.

mcuee:
For MinGW, there is one more patch required, which is also not acceptable as it is. But again @mariusgreuel or other CMake experts may be able to figure out a way to modify it to be suitable to be adopted.
https://github.com/mcuee/avrdude-packing/blob/main/patches/0007-Cmake-fix-build-not-working-with-mingw-toolchain.patch

mcuee:
> The error is rather strange and point out potential issue with the way ncurses package is built in the container. Ref: https://stackoverflow.com/questions/8956455/undefined-reference-building-ncurses-on-cygwin

Indeed @um has figured out the issue and find the solution to bump up ncurses version.
* https://github.com/arduino/crossbuild/pull/8

mcuee:
Take note the Arduino Cross-build container is here and uses GNU Affero General Public License v3.0. So it should be acceptable to use. 
https://github.com/arduino/crossbuild

mcuee:
Snapshot release here, up to commit https://github.com/avrdudes/avrdude/commit/b5699663c4ebdb185c21f2f1ca02610cd078ed77.
https://github.com/mcuee/avrdude-packing/releases/tag/snapshot_30Oct2022

Two patches are applied on top of avrdude git main.
https://github.com/mcuee/avrdude-packing/tree/main/patches

mcuee:
Snapshot release here, up to latest git commit https://github.com/avrdudes/avrdude/commit/4c92030e3a486cfbaeb36e298f6f2929e8e031eb.
https://github.com/mcuee/avrdude-packing/releases/tag/snapshot_2Nov2022

Two patches are applied on top of avrdude git main.
[...]

PR #1190, opened 2022-11-22
Provide bootloader-hash tool

stefanrueger:
This bash script computes the hash-sum of stk500v1 bootloaders to ease the guesswork of `-c urclock`. The `-c urclock` programmer is compatible with `-c arduino`, but needs to know the size of the bootloader as it protects it *externally* from being overwritten. In contrast to urboot, optiboot et al  binaries do not advertise its size and properties. Hence, `urclock.c` maintains a table with hash-sums of popular bootloaders that are out there in the wild, so the user doesn't have to know, let alone specify, the size of the bootloader on their devices. This utility computes the table entry from a .hex files in the directory.

### DIFF
--- a/src/urclock.c
+++ b/src/urclock.c
@@ -1142,9 +1142,6 @@ static void guessblstart(const PROGRAMMER *pgm, const AVRPART *p) {
     { 2048, 0, 0x12ab8da0, 0xca46a3ca }, // ATmegaBOOT_168_diecimila.hex
     { 2048, 0, 0x3242ddd3, 0xf3e94dba }, // ATmegaBOOT_168_ng.hex
     { 2048, 0, 0x2eed30b3, 0x47d14ffa }, // ATmegaBOOT_168_pro_8MHz.hex
-    { 4096, 0, 0xc52edd05, 0xa3371f94 }, // Caterina-LilyPadUSB.hex
-    { 4096, 0, 0x663b8f7e, 0x7efdda2b }, // Caterina-Robot-Control.hex
-    { 4096, 0, 0x3c6387e7, 0x7e96eea2 }, // Caterina-Robot-Motor.hex
     { 2048, 0, 0x1cef0d75, 0x6cfbac49 }, // LilyPadBOOT_168.hex
     { 1024, 1, 0x6ca0f37b, 0x31bae545 }, // bigboot_328.hex
     {  512, 0, 0x035cbc07, 0x24ba435e }, // optiboot_atmega168.hex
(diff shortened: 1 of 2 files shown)


mcuee:
This pull request seems to be good.

But I found an issue when testing this pull request. We may need to increase the Windows drain timeout value from 80ms to 150ms for `-c urclock`.
https://github.com/avrdudes/avrdude/pull/1171#issuecomment-1324502480

mcuee:
@stefanrueger 

The only thing is that the user will have to rebuild avrdude if they want to include some bootloaders. I am thinking we can extend `avrdude.conf` and user can add more entries with the help of `bootloader-hash` tool so that `-c urclock` can be used for more optiboot bootloader hex files.

mcuee:
> The only thing is that the user will have to rebuild avrdude if they want to include more optiboot bootloaders. I am thinking we can extend avrdude.conf and user can add more entries with the help of bootloader-hash tool so that -c urclock can be used for more optiboot bootloader hex files.

@stefanrueger and @MCUdude 

I am thinking that `urboot` will take time to go into more main stream use. Once avrdude 7.1 is released, maybe @MCUdude can try to add `urboot` and `urclock` to his popular Arduino Core (eg: MiniCore, MightyCore and MegaCore). In that case, we will be able to reach many more users.

I understand that Arduino will take time to adopt new version of avrdude (they have not officially adopted 7.0 yet). But I think we can already build good static link avrdude version using the Arduino provided infrastructure. So it may not be a big issue any more.
* https://github.com/avrdudes/avrdude/issues/1162

stefanrueger:
> For example, I hope adding (at least the 115200bps version) of all the optiboot_flash hex files from @MCUdude should not slow down avrdude.

The intention with the hashes of a few popular bootloaders is to make it easy to change from `-c arduino` to `-c urboot`. Whilst it would not slow down avrdude (or make it noticeably bigger) to record 12-byte table entries for 10,000 bootloaders, I think it would be unseemly to have these externally in `avrdude.conf`. Maybe an option to add a couple of them in a private `avrdude.conf` is OK, though, but there would need to be a mechanism for the users to figure out what they need to put into ~/.avrduderc. The standalone tool is not an option for this; it is meant for developers (and so we don't forget how the hash table was created).

The way I think about this is that arduino's distributed bootloaders should be recognised amongst which is the v4.4 optiboot (that's probably on millions of chips all over the world). People who download and install custom bootloaders, eg, from @MCUdude's repository should be able to cope with an extra parameter `-xbootsize=...` and be able to hide this in their workflow.

mcuee:
@stefanrueger 

I see your point. Thanks.

I think this PR is good to go now.

PR #1195, opened 2022-11-23
Added get configurations fuses, lockbits and eeprom from ELF files.

McMCCRU:
Screenshot_20221117_011722
Screenshot_20221117_011617
Hi!
I added the ability to get configurations for fuses, lockbits and eeprom from elf files when programming flash. This eliminates the need for additional commands for mass programming, as is done in AVR Studio. For this to work, you need to insert the necessary macros in the source program and compile it.
Example for AVR C code.
For Atmega8:

```c
FUSES = { 0xC9, 0xEF };
LOCKBITS = 0xFC;
/* Fix add .eeprom section */
#define _EEMEM  __attribute__((__used__, __section__ (".eeprom")))
#define _EEPROM uint8_t __eeprom[512] _EEMEM

_EEPROM = { 0x11, 0x22, };
```

For Attiny10(Not support FUSES macros!!!):

```c
#define FUSES uint8_t __fuses FUSEMEM
FUSES = 0xFE;
LOCKBITS = 0xFC;
```


stefanrueger:
Thanks for the PR @McMCCRU. This looks like an interesting and worthwhile new feature. I have not carefully reviewed the code changes, but on a first glance the code looks competently written. I may not be the right person on the team to comment, so I trust someone else will chip in. Until then a couple of questions: 
 - It *looks like* memories are detected by use of certain *global variables* in the source file; is this a standard convention? I genuinely do not know. Would users know about this? Naively I would have thought that different linker *sections* in the elf would/could describe different memories.
 - The screenshots show that `-U flash:w:application.elf:a` triggers writing of lockbits, fuses etc;  whilst in one way this is elegant and useful for automation, it will no doubt surprise users. For me, it would be more logical if 
    + `-U `*memory*`:w:file.elf:a` would *only* pick and write *memory* from the .elf file
    + `-U file.elf` could do what this PR suggests: write all the memories that are encoded in the `.elf`
 - As this is such a useful (and welcome) addition to avrdude, it needs documentation in avrdude.1 and doc/avrdude.texi

McMCCRU:
Hi!
1. https://microchipsupport.force.com/s/article/How-to-generate-ELF-Prodution-file-for-AVR-8-bit-MCU
Screenshot_20221125_130951
You can find more information about Production file in Atmel Studio User guide.
P.S. Some macros are better to create yourself, rather than using ready-made ones from avr-libs, not all of them work correctly.
For example, the original version <avr/eeprom.h>:
#define EEMEM __attribute__((section(".eeprom")))
but when linking, this section will disappear, so that it doesn’t disappear, it’s right to do this:
#define EEMEM __attribute__ ((__used__, __section__ (".eeprom")))
2. -U flash:w:file.elf:a can be simplified to -U file.elf, but it needs to check that it's an elf file and rewrite a lot of wrappers used avrdude, is it worth it?
3. In the avrdude documentation, you can refer to the official.(see 1)

MCUdude:
Thanks for the PR @McMCCRU!

> The screenshots show that -U flash:w:application.elf:a triggers writing of lockbits, fuses etc; whilst in one way this is elegant and useful for automation, it will no doubt surprise users. For me, it would be more logical if
-U memory:w:file.elf:a would only pick and write memory from the .elf file
-U file.elf could do what this PR suggests: write all the memories that are encoded in the .elf

I agree with @stefanrueger here. when the memory is explicitly specified, Avrdude should IMO only read/write to this memory. However `-U file.elf` is short and neat, and isn't surprising, since no memory is explicitly specified. the suffix after the file name (`:a`,`:h`,`:i`,`:r` etc.) are optional, and `:a`, auto is selected if no suffix is present (I believe)

Then this would be possible:

```
# Write only write flash and fuse bits specified in the ELF file
$ avrdude -c usbasp -p atmega328p -U flash:w:file.elf -U lfuse:w:file.elf -U hfuse:w:file.elf -U efuse:w:file.elf

# Write to all memories specified in the ELF file
$ avrdude -c usbasp -p atmega328p -U file.elf
```

McMCCRU:
-U file.elf works the same as -U flash:w:file.elf, I tested it recently. I don’t understand the point of writing everything separately, this is what I wanted to get rid of, as in AVR Studio...

MCUdude:
>  I don’t understand the point of writing everything separately, this is what I wanted to get rid of, as in AVR Studio...

The idea is that it lets Avrdude extract parts of an ELF file, which may be very convenient. If you _just_ want the fuse bits or EEPROM content in an ELF file someone sent you, you can do that.

McMCCRU:
This violates the principle of automation, the idea is just to automate this process with one command. To do this, you include the necessary, selectively, macros in the code itself for compilation.

MCUdude:
> This violates the principle of automation, the idea is just to automate this process with one command. To do this, you include the necessary, selective, macros in the code itself for compilation.

But one could still write _everything_ the ELF file contains by not specifying a memory? Then no memory is specified, only `-U`:
`$ avrdude -c usbasp -p atmega328p -U file.elf`

I'm not trying to be difficult, I'm just trying to express what I would find more intuitive 🙂

McMCCRU:
> Но можно ли записать _все_ , что содержится в файле ELF, не указывая память? Тогда память не указывается, только `-U`:
> `$ avrdude -c usbasp -p atmega328p -U file.elf`

This is exactly how it works now.

Screenshot_20221125_152451

stefanrueger:
| -U file.elf works the same as -U flash:w:file.elf

And this is *exactly* the issue with this PR. The syntax for `-U memory:w:file:type` is that the device `memory` is written to from contents of `file`. I predict that a user deploying `-U flash:w:file.elf:a` will be surprised, perhaps furious even, if the fuses of the board and/or potentially carefully curated eeprom will be overwritten. The code should *pick* the specified memory from the `.elf` file (if present) and *only* that.

As to `-U file.elf`: the suggested functionality sounds like a good and worthwhile extension that is easy to explain in the documentation.

McMCCRU:
OK!
Screenshot_20221125_171054
Screenshot_20221125_171140

McMCCRU:
In original source code, everything worked like that, but there is a bug, there is no section check, so the "flash" command wrote all sections, but it should have written only .text, which could lead to flash overflow. I fixed this...

stefanrueger:
Thank you @McMCCRU for your original contribution and for accommodating the avrdude syntax re `-U`.

Someone in the team will carry out a careful review of the code. In the meantime, feel free  to review and test the code yourself (as you have already done) and push further commits. Once the code is approved, we still like the PR to hang around for a few days before it is merged so that people can try out the feature. (And as an advance warning: we are all volunteers, so all this may take time, and if worse came to worst it could be that the PR will only be merged after the next release 7.1)

But, yes, I like the idea of capturing all memories in the .elf and enabling the user to write them!

McMCCRU:
I thought that it was not convenient to use ELF files, as it was supposed in avrdude, so I redid it, as AVR Studio suggests, especially since many applications working with avrdude do not allow using format -U file.elf.
Screenshot_20221125_193621
It’s not very clear to me why use ELF for each operation separately, the idea was to automate everything with one flash command, and if fuses, lockbits and eeprom are not needed, then you don’t need to use them in the firmware code.

McMCCRU:
avr-libs not supported section .fuse for ATtiny4/5/9/10 in link scripts. For place .fuse use .config. A new version example for AVR C is now like this:
`#define _FUSEMEM __attribute__((__used__, __section__(".config")))            
#define FUSES uint8_t __fuses _FUSEMEM
FUSES = 0xFE;`
I also had to add new changes to the patch.

MCUdude:
Thank your for your excellent work @McMCCRU!

I just tried this PR with an ATmega328p, and this is really cool! As expected, I can write to all memories specified in the ELF file, but I can also manually specify which memory to write to. Very, very cool!

Quick question. I looked into why `FUSES = {0xFF}` won't work for ATtiny10, and it looks like a bug/typo in `iotn10.h`.
In this file `#define FUSE_MEMORY_SIZE 0` is defined.

This prevents the following code from working:


``` c
// avr/fuse.h
#if !(defined(__ASSEMBLER__) || defined(__DOXYGEN__))

#ifndef FUSEMEM
#define FUSEMEM  __attribute__((__used__, __section__ (".fuse")))
#endif

#if FUSE_MEMORY_SIZE > 3

typedef struct
{
    unsigned char byte[FUSE_MEMORY_SIZE];
} __fuse_t;


#elif FUSE_MEMORY_SIZE == 3

typedef struct
{
    unsigned char low;
    unsigned char high;
    unsigned char extended;
} __fuse_t;

#elif FUSE_MEMORY_SIZE == 2

typedef struct
{
    unsigned char low;
    unsigned char high;
} __fuse_t;

#elif FUSE_MEMORY_SIZE == 1

typedef struct
{
    unsigned char byte;
} __fuse_t;

#endif
```

But instead of using the `.config` section, why not use `.fuse` instead? This seems to be present.

MCUdude:
> avr-libs not supported section .fuse for ATtiny4/5/9/10 in link scripts. For place .fuse use .config.

OK, sorry for the dumb question. Where can I see where all the sections that's defined for the part? (.fuse, .config etc)

MCUdude:
It turns out that the `FUSE_MEMORY_SIZE` bug has been fixed in the [latest Microchip ATtiny pack](https://packs.download.microchip.com).

In the latest pack, `FUSE_MEMORY_SIZE` is defined as 1.

But does this change anything? Do we still have to use `.config`, or can `.fuse` be used instead?
Sorry for all the questions, I'm just trying to understand

McMCCRU:
> OK, sorry for the dumb question. Where can I see where all the sections that's defined for the part? (.fuse, .config etc)
All these sections are described in the link files in the official toolchains. Example:
Screenshot_20221125_235236
Screenshot_20221125_235319
See MEMORY { }

McMCCRU:
> It turns out that the `FUSE_MEMORY_SIZE` bug has been fixed in the [latest Microchip ATtiny pack](https://packs.download.microchip.com).
> In the latest pack, `FUSE_MEMORY_SIZE` is defined as 1.
> But does this change anything? Do we still have to use `.config`, or can `.fuse` be used instead? Sorry for all the questions, I'm just trying to understand

In the linking scripts ATtiny4/5/9/10 themselves, there is no description for the "fuse" section, it is suggested to use "config" instead. But there is no macro for it, so you need to do it yourself, as I wrote in the example above for ATtiny4/5/9/10....

stefanrueger:
@McMCCRU I was about to start a review when I noticed that the code uses a global variable all_write_elf and static is_flash. This made me wonder whether the code has considered chains of `-U`:
```
avrdude ... -U efuse:w:file1.elf:a -U eep:r:backup.eep:I -U file2.elf -U fl:r:backup.hex:I -U ee:w:file3.elf:a
```
You have a mixture of .elf file usage here, some where a memory is picked and one where the *new* syntax writes all memories that are present the .elf file. I was wondering whether one global variable can actually handle this. And whether a global variable is a good tool here. Maybe a better idea to explore how that information flow should be handled. Can the updateflags be of any use (have a flag that says, actually all memories of the elf file should be written, not only the one mentioned in the command)? Or invent a memory type "*" for all memories? I don't have answers, but I couldn't imagine that a global variable can handle a `-U` chain.

McMCCRU:
> @McMCCRU I was about to start a review when I noticed that the code uses a global variable all_write_elf and static is_flash.

You're right, I used these variables for debugging, and I was going to change this code. I updated it...

McMCCRU:
Hi!
I have completed all with -U the checks and made the last improvements, until I see the point in changing anything else.

mcuee:
I will learn more about avr-gcc and carry out some  tests next week. I am not a programmer but I think I can learn just enough to carry out the testing.

I will see if I can modify the Blink sketch for the Arduino Uno with fuse and EEPROMs to carry out the testing. 

Or is it better to use plain AVR-gcc?

McMCCRU:
> Or is it better to use plain AVR-gcc?

Arduino uses the same avr-gcc...

MCUdude:
@mcuee It's very easy to test this PR out.

Here's some simple test code for an ATtiny328P that you can test by compiling for an Arduino UNO. The EEPROM can read by Avrdude to veryfy that it has been written. You can get the ELF file from Arduino IDE by turning on verbose compilation output and copying the path to the ELF file that appears at the bottom of the output log.

```c
#define _EEMEM  __attribute__((__used__, __section__ (".eeprom")))
#define _EEPROM uint8_t __eeprom[32] _EEMEM
_EEPROM = { "Hello World!" };

FUSES = { 0xF7, 0xD6, 0xFD }; // lfuse, hfuse, efuse

void setup() {
  pinMode(13, OUTPUT);
}

void loop() {
  digitalWrite(13, HIGH);
  delay(1000);
  digitalWrite(13, LOW);
  delay(1000);
}
```

McMCCRU:
If you use eeprom as in this example, then you do not need to create your own attributes, the .eeprom section in the ELF file will not be cut off when linking.

```
#include <avr/eeprom.h>

uint8_t EEMEM NonVolatileChar;
uint16_t EEMEM NonVolatileInt;
uint8_t EEMEM NonVolatileString[10];

void main(void) {
    uint8_t SRAMchar;
    uint16_t SRAMint;
    uint8_t SRAMstring;
    SRAMchar = eeprom_read_byte(&NonVolatileChar);
    SRAMint = eeprom_read_word(&NonVolatileInt);
    eeprom_read_block((void*)&SRAMstring, (const void*)&NonVolatileString, 10);
}

```

stefanrueger:
Looks very good now, @McMCCRU. Thank you. A few minor comments (again *without* doing a proper code review)
 - Better to replace strcmp(..., "flash"/"eeprom") with one of the following (these know application, boot etc)
https://github.com/avrdudes/avrdude/blob/baa52e3a02b67dbebc01211b284032d11d73feac/src/libavrdude.h#L902-L908
 - MAX_FUSE_SIZE and MAX_LOCK_SIZE may be incorrect. There are already parts with 4 lock bytes, eg,
``` sh
$ avrdude -p*/At | grep lock.size | grep -v "[^0-9]1$"
.ptmm	AVR-Dx family common values	lock	size	4
.ptmm	AVR32DA28	lock	size	4
.ptmm	AVR32DA32	lock	size	4
...
```
and who knows about future parts? Maybe use a generous array of, say, 32 bytes for fuses and locks each and have a run-time check whether this was sufficient, or use `cfg_malloc(__func__, size)` to allocate the needed space at runtime and free() after use? 
 - The type ELF_CMD is only used in fileio.c, so perhaps better to define this type locally there, not in `libavrdude.h`?
 - The documentation needs to mention the new syntax. Currently, `src/avrdude.1` says "As an abbreviation, the form -U filename is equivalent to specifying -U flash:w:filename:a"; this should change to something to the effect of "As an abbreviation, the form -U filename is equivalent to specifying -U flash:w:filename:a unless filename is an ELF file, in which case -U filename means Write to the device all memories that are defined in the ELF file."; the same for `src/doc/avrdude.texi`

I agree with @MCUdude that this is a vvv neat innovation!

McMCCRU:
@stefanrueger, I took into account your recommendations and changed the code.

McMCCRU:
In ```avr/lock.h``` it can be seen that the lock bits can be one byte or four bytes. 
```
#if defined(LOCKBITS_SIZE) && (LOCKBITS_SIZE == 4)
#define LOCKBITS unsigned long __lock LOCKMEM
#else
#define LOCKBITS unsigned char __lock LOCKMEM
#endif /* !(defined(LOCKBITS_SIZE) && (LOCKBITS_SIZE == 4)) */
```
I don't know what order these bytes should be written in when reading from ELF.

stefanrueger:
Thanks for the updates!

> don't know what order these bytes should be written in when reading from ELF

Nor do I. @dl8dtl?

The order of `-U` processing does not seem right. This PR makes `main.c`'s  `updates` list global. Its use seems to be:
```
$ grep updates src/fileio.c 
extern LISTID updates;
                ladd(updates, upd);
              ladd(updates, upd);
              ladd(updates, upd);
        ladd(updates, upd);
        ladd(updates, upd);
        for (ln=lfirst(updates); ln; ln=lnext(ln)) {
```
Does this mean that `-U file.elf` encounters are *appended* to the updates list? This would mean that the operations are done in a *different* order than the user will expect them to be done. A user who asks
```
$ avrdude .... -U file.elf -U eeprom:w:my.eep:i
```
would expect that whatever eeprom contents there was in `file.elf` be overwitten by `my.eep`. If my hunch is correct (I have not read the code carefully) then the memories encountered in `file.elf` will be treated *last*. Any eeprom contents in `file.elf` *might overwrite* what the user definitely wanted on the chip (`my.eep`).

So the strategy of this PR does not yet seem completely right... (and the updates to the documentation are missing).

mcuee:
First test using the example given by @MCUdude and it seems to work.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195 -c usbasp -p m168p -U .\Blink_eeprom.ino.ino.elf

avrdude_pr1195: AVR device initialized and ready to accept instructions
avrdude_pr1195: device signature = 0x1e940b (probably m168p)
avrdude_pr1195: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1195: erasing chip
avrdude_pr1195: found in ELF eeprom, size = 32
avrdude_pr1195: found in ELF lfuse = 0xF7
avrdude_pr1195: found in ELF hfuse = 0xD6
avrdude_pr1195: found in ELF efuse = 0xFD
avrdude_pr1195: reading input file .\Blink_eeprom.ino.ino.elf for flash
                with 1030 bytes in 1 section within [0, 0x405]
                using 9 pages and 122 pad bytes
avrdude_pr1195: writing 1030 bytes flash ...

Writing | ################################################## | 100% 0.34 s

avrdude_pr1195: 1030 bytes of flash written
avrdude_pr1195: verifying flash memory against .\Blink_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.21 s

avrdude_pr1195: 1030 bytes of flash verified
avrdude_pr1195: reading input file .\Blink_eeprom.ino.ino.elf for eeprom
                with 32 bytes in 1 section within [0, 0x1f]
                using 8 pages and 0 pad bytes
avrdude_pr1195: writing 32 bytes eeprom ...

Writing | ################################################## | 100% 0.39 s

avrdude_pr1195: 32 bytes of eeprom written
avrdude_pr1195: verifying eeprom memory against .\Blink_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.03 s

avrdude_pr1195: 32 bytes of eeprom verified
avrdude_pr1195: reading input file 0xF7 for lfuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte lfuse ...
avrdude_pr1195: 1 byte of lfuse written
avrdude_pr1195: verifying lfuse memory against 0xF7
avrdude_pr1195: 1 byte of lfuse verified
avrdude_pr1195: reading input file 0xD6 for hfuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte hfuse ...
avrdude_pr1195: 1 byte of hfuse written
avrdude_pr1195: verifying hfuse memory against 0xD6
avrdude_pr1195: 1 byte of hfuse verified
avrdude_pr1195: reading input file 0xFD for efuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte efuse ...
avrdude_pr1195: 1 byte of efuse written
avrdude_pr1195: verifying efuse memory against 0xFD
avrdude_pr1195: 1 byte of efuse verified

avrdude_pr1195 done.  Thank you.

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude -c usbasp -p m168p -qqt
avrdude> dump lfuse
>>> dump lfuse
0000  f7                                                |.               |

avrdude> dump hfuse
>>> dump hfuse
0000  d6                                                |.               |

avrdude> dump efuse
>>> dump efuse
0000  fd                                                |.               |

avrdude> dump eeprom 0 0x40
>>> dump eeprom 0 0x40
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 00 00 00  |Hello World!....|
0010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|

avrdude> quit
>>> quit
avrdude>

```

mcuee:
> The order of -U processing does not seem right. 

@stefanrueger and @McMCCRU 

Indeed this is an issue.

```
PS C:\work\avr\avrdude_test\avrdude_bin> cat .\entest_256B.eep
:020000040000FA
:2000000054686520717569636B2062726F776E20666F78206A756D7073206F76657220740E
:200020006865206C617A7920646F670A54686520717569636B2062726F776E20666F78207C
:200040006A756D7073206F76657220746865206C617A7920646F670A5468652071756963FD
:200060006B2062726F776E20666F78206A756D7073206F76657220746865206C617A7920D4
:20008000646F670A54686520717569636B2062726F776E20666F78206A756D7073206F76B5
:2000A000657220746865206C617A7920646F670A54686520717569636B2062726F776E20FE
:2000C000666F78206A756D7073206F76657220746865206C617A7920646F670A54686520C2
:2000E000717569636B2062726F776E20666F78206A756D7073206F76657220746865206C16
:00000001FF
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195 -c usbasp -p m168p -U .\Blink_eeprom.ino.ino.elf -U eeprom:w:entest_256B.eep:i

avrdude_pr1195: AVR device initialized and ready to accept instructions
avrdude_pr1195: device signature = 0x1e940b (probably m168p)
avrdude_pr1195: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1195: erasing chip
avrdude_pr1195: found in ELF eeprom, size = 32
avrdude_pr1195: found in ELF lfuse = 0xF7
avrdude_pr1195: found in ELF hfuse = 0xD6
avrdude_pr1195: found in ELF efuse = 0xFD
avrdude_pr1195: reading input file .\Blink_eeprom.ino.ino.elf for flash
                with 1030 bytes in 1 section within [0, 0x405]
                using 9 pages and 122 pad bytes
avrdude_pr1195: writing 1030 bytes flash ...

Writing | ################################################## | 100% 0.34 s

avrdude_pr1195: 1030 bytes of flash written
avrdude_pr1195: verifying flash memory against .\Blink_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.22 s

avrdude_pr1195: 1030 bytes of flash verified
avrdude_pr1195: reading input file entest_256B.eep for eeprom
                with 256 bytes in 1 section within [0, 0xff]
                using 64 pages and 0 pad bytes
avrdude_pr1195: writing 256 bytes eeprom ...

Writing | ################################################## | 100% 3.14 s

avrdude_pr1195: 256 bytes of eeprom written
avrdude_pr1195: verifying eeprom memory against entest_256B.eep

Reading | ################################################## | 100% 0.26 s

avrdude_pr1195: 256 bytes of eeprom verified
avrdude_pr1195: reading input file .\Blink_eeprom.ino.ino.elf for eeprom
                with 32 bytes in 1 section within [0, 0x1f]
                using 8 pages and 0 pad bytes
avrdude_pr1195: writing 32 bytes eeprom ...

Writing | ################################################## | 100% 0.39 s

avrdude_pr1195: 32 bytes of eeprom written
avrdude_pr1195: verifying eeprom memory against .\Blink_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.03 s

avrdude_pr1195: 32 bytes of eeprom verified
avrdude_pr1195: reading input file 0xF7 for lfuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte lfuse ...
avrdude_pr1195: 1 byte of lfuse written
avrdude_pr1195: verifying lfuse memory against 0xF7
avrdude_pr1195: 1 byte of lfuse verified
avrdude_pr1195: reading input file 0xD6 for hfuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte hfuse ...
avrdude_pr1195: 1 byte of hfuse written
avrdude_pr1195: verifying hfuse memory against 0xD6
avrdude_pr1195: 1 byte of hfuse verified
avrdude_pr1195: reading input file 0xFD for efuse
                with 1 byte in 1 section within [0, 0]
avrdude_pr1195: writing 1 byte efuse ...
avrdude_pr1195: 1 byte of efuse written
avrdude_pr1195: verifying efuse memory against 0xFD
avrdude_pr1195: 1 byte of efuse verified

avrdude_pr1195 done.  Thank you.

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude -c usbasp -p m168p -qqt
avrdude> dump eeprom
>>> dump eeprom
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 00 00 00  |Hello World!....|
0010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
0020  68 65 20 6c 61 7a 79 20  64 6f 67 0a 54 68 65 20  |he lazy dog The |
0030  71 75 69 63 6b 20 62 72  6f 77 6e 20 66 6f 78 20  |quick brown fox |
0040  6a 75 6d 70 73 20 6f 76  65 72 20 74 68 65 20 6c  |jumps over the l|
0050  61 7a 79 20 64 6f 67 0a  54 68 65 20 71 75 69 63  |azy dog The quic|
0060  6b 20 62 72 6f 77 6e 20  66 6f 78 20 6a 75 6d 70  |k brown fox jump|
0070  73 20 6f 76 65 72 20 74  68 65 20 6c 61 7a 79 20  |s over the lazy |
0080  64 6f 67 0a 54 68 65 20  71 75 69 63 6b 20 62 72  |dog The quick br|
0090  6f 77 6e 20 66 6f 78 20  6a 75 6d 70 73 20 6f 76  |own fox jumps ov|
00a0  65 72 20 74 68 65 20 6c  61 7a 79 20 64 6f 67 0a  |er the lazy dog |
00b0  54 68 65 20 71 75 69 63  6b 20 62 72 6f 77 6e 20  |The quick brown |
00c0  66 6f 78 20 6a 75 6d 70  73 20 6f 76 65 72 20 74  |fox jumps over t|
00d0  68 65 20 6c 61 7a 79 20  64 6f 67 0a 54 68 65 20  |he lazy dog The |
00e0  71 75 69 63 6b 20 62 72  6f 77 6e 20 66 6f 78 20  |quick brown fox |
00f0  6a 75 6d 70 73 20 6f 76  65 72 20 74 68 65 20 6c  |jumps over the l|
```

stefanrueger:
Thanks - I'm hoping that someone in the team can do a thorough review, @MCUdude?

McMCCRU:
I just noticed that in the fileio.c file for Win32(fileio_fmt_autodetect()) does not correctly use the open() function, I copied this code into my function and noticed it upon closer inspection.

```
#if defined(WIN32)
  f = fopen(filename, "r");
#else
  f = fopen(filename, "rb");
#endif
```
In my opinion it should be the other way around....

stefanrueger:
Correct: feel free to fix this problem en passant. No need to open an issue and a separate PR

src/main.c
```diff
@@ -1426,9 +1426,43 @@ int main(int argc, char * argv [])
...
+  rc = 0;
+  for (ln=lfirst(updates); ln; ln=lnext(ln)) {
+    upd = ldata(ln);
+    if (avr_memtype_is_flash_type(upd->memtype) && (upd->op == DEVICE_WRITE) &&
+        upd->elf_all_write) {
+      rc++;
+    }
+  }
+  if (rc > 1) {
+    pmsg_error("option -U file.elf can be used only once!!!\n");
```
MCUdude, line 1439: We should try to make this error a little less angry. How about
`pmsg_error("option -U file.elf can only be used once.\n");` ?

src/fileio.c
```diff
@@ -801,16 +829,167 @@ static int elf_mem_limits(const AVRMEM *mem, const AVRPART *p,
...
+
+FUSES = { 0xC9, 0xEF };
+LOCKBITS = 0xFC;
+// Fix add .eeprom section
+#define _EEMEM  __attribute__((__used__, __section__ (".eeprom")))
+#define _EEPROM uint8_t __eeprom[512] _EEMEM
+
+_EEPROM = { 0x11, 0x22, };
+
+For Attiny4/5/9/10(Not support FUSES macros and .fuse section!!!):
```
MCUdude, line 845: A little too angry perhaps. How about:
`For ATtiny4/5/9/10 (FUSES macro and .fuses section not supported by avr-libc):` ?

src/main.c
McMCCRU: I did.

src/fileio.c
McMCCRU: I did

McMCCRU:
> I'll see if I can come up with something. And perhaps get some input from @stefanrueger with regards to spelling and formatting.

That would be excellent.

stefanrueger:
> `pmsg_error("option -U file.elf can only be used once.\n");`

Out of curiosity, why is that limitation in place? A use case could reasonably *want*
```
$ avrdude ... -U unlock.elf -U beef.elf -U lock.elf
```
with three different .elf files, one `unlock.elf` putting the fuses/locks into a state that allows the file `beef.elf` with eeprom, application, apptable, calibration etc memories (the "beef" or core of the application) and one `lock.elf` to secure the part against unauthorised access. A classic pipeline for factory type installation, not?

My impression is that this PR does not explore the full potential of the excellent(!) idea to allow the *new* syntax `-U file.elf` to write all the memories that are in file.elf (as opposed to the current syntax that translates it to `-U flash:w:file.elf:a`). Somehow I think it might be easier to adapt the *parsing* to the idea, less so the file processing side.

A neat documentation with examples will be cool. For now, at the very least, the doc must say that syntax of `-U file` is more general when file is an ELF file.

mcuee:
>>> The order of -U processing does not seem right.
> > Indeed this is an issue.
> 
> Fixed...

Yes this is now fixed.

mcuee:
> It gets a ✅ from me, tested on an ATmega16 and an ATmega328P on MacOS.

@MCUdude 

Great that you have looked at the codes and tested it. 

I am thinking that we need to carry out futher tests with different AVR chips with somewhat different memory organization. One thing you can help is probbably to test with XMega AVRs. 

I will try to test the new UPDI AVRs like the ATmega4808.

McMCCRU:
> Из любопытства, почему это ограничение на месте?

This is very dangerous, firstly, there will be a constant attempt to overwrite flash memory, in your example it will be 3 times, and secondly, collecting elf files with different, non-crossing sections will not be an easy task. Therefore, this is a bad idea and can only confuse users.

mcuee:
> Out of curiosity, why is that limitation in place? A use case could reasonably _want_
> 
> ```
> $ avrdude ... -U unlock.elf -U beef.elf -U lock.elf
> ```
> 
> with three different .elf files, one `unlock.elf` putting the fuses/locks into a state that allows the file `beef.elf` with eeprom, application, apptable, calibration etc memories (the "beef" or core of the application) and one `lock.elf` to secure the part against unauthorised access. A classic pipeline for factory type installation, not?

@stefanrueger 

I think you have an interesting use case here. My thoughts:
1) I tend to think they will not use ELF file -- I have never seen ELF file used myself, usually hex/bin file with some programming instructions to set the fuses or configuration bits.
2) Usually the above three steps will not run at the same time, for example, calibration may be a different step.

McMCCRU:
> My impression is that this PR does not explore the full potential of the excellent(!) idea to allow the _new_ syntax `-U file.elf` to write all the memories that are in file.elf (as opposed to the current syntax that translates it to `-U flash:w:file.elf:a`). Somehow I think it might be easier to adapt the _parsing_ to the idea, less so the file processing side.

It will not work to put everything in an elf, the number of sections is limited in linking scripts. Here is the maximum list that can be placed in an elf file.
Screenshot_20221129_030913
You will need to completely redo this list, make a new memory card, so as not to break or spoil anything.

mcuee:
> This is very dangerous, firstly, there will be a constant attempt to overwrite flash memory, in your example it will be 3 times, 

But that is the user asks avrdude to do so. 

If you look at @stefanrueger's example, `unlock.elf` and `lock.elf` deal with fuses and `beef.elf` deal with Flash and EEPROM. So in reality no overwriting of Flash.

> and secondly, collecting elf files with different, non-crossing sections will not be an easy task.

I see, that is a valid reason. In this case, I think we can leave  this PR as it is. Future PRs may improve on this aspect.

@stefanrueger, any objections to leave this PR as it is for now?

McMCCRU:
> If you look at @stefanrueger's example, `unlock.elf` and `lock.elf` deal with fuses and `beef.elf` deal with Flash and EEPROM. So in reality no overwriting of Flash.

The .text section won't disappear, unless you delete it yourself with binutils, flash memory will be overwritten, -U file.elf is the equivalent of the -U flash:w:file.elf command.

mcuee:
> The .text section won't disappear, unless you delete it yourself with binutils, flash memory will be overwritten.

I see. Thanks for the explanation. Then ELF is not good to be used for the particular use case.

> -U file.elf is the equivalent of the -U flash:w:file.elf command.

Hmm, that is not really correct. The forst command deal with all the memory section specified in the ELF file, including fuse and EEPROM. The second command should only deal with Flash memory, not others.

McMCCRU:
> Hmm, that is not really correct. The forst command deal with all the memory section specified in the ELF file, including fuse and EEPROM. The second command should only deal with Flash memory, not others.

From code:
Screenshot_20221129_033914
Yes, I made it so that when -U file.elf is called, all sections from the elf file are read, before only flash memory writing worked.

McMCCRU:
I would also like to add that if you want to create your own section, which is not specified in the linking script, then this section will be created inside the .date section. For example:

```
#define _CALMEM  __attribute__((__used__, __section__ (".calibration")))
#define _CALIBRATION uint8_t __calibration _CALMEM
_CALIBRATION = 0x69;
```
But keep in mind that this section is limited by the size of the RAM, and will also require code improvements.

mcuee:
> Yes, I made it so that when -U file.elf is called, all sections from the elf file are read, before only flash memory writing worked.

@McMCCRU 
Thanks. This is a good addition to avrdude. 

@MCUdude @stefanrueger @McMCCRU 
Now we are have a good convenient way to write all the memory sections using a single elf file. What about memory reading?

I am thinking of a new memory name called `all` and I think it is nice to have the following command. Any comments?
```
avrdude -c xxx -p xxx -U:all:r:readhackall.elf
```

MCUdude:
> I am thinking that we need to carry out futher tests with different AVR chips with somewhat different memory organization. One thing you can help is probbably to test with XMega AVRs. I do not have good means to test them as of now .

Sure, I can give it a spin on an ATxmega128A1U and an ATxmega256A3BU. Can you try on various AVRs with UPDI?
On these, the fuses section would look like this is C and optionally for C++:

``` c
//FUSE default constants defined in avr/iom4809.h

// Compatible with C. I'm pretty sure reserved fuses also needs to be present
FUSES = {
  FUSE0_DEFAULT,
  FUSE1_DEFAULT,
  FUSE2_DEFAULT,
  0x00,
  0x00,
  FUSE5_DEFAULT,
  FUSE6_DEFAULT,
  FUSE7_DEFAULT,
  FUSE8_DEFAULT
};

// Compatible with C++
FUSES = {
  .WDTCFG  = FUSE_WDTCFG_DEFAULT,  // 0x00, FUSE0_DEFAULT
  .BODCFG  = FUSE_BODCFG_DEFAULT,  // 0x00, FUSE1_DEFAULT
  .OSCCFG  = FUSE_OSCCFG_DEFAULT,  // 0x02, FUSE2_DEFAULT
  .SYSCFG0 = FUSE_SYSCFG0_DEFAULT, // 0xC4, FUSE5_DEFAULT
  .SYSCFG1 = FUSE_SYSCFG0_DEFAULT, // 0x07, FUSE6_DEFAULT
  .APPEND  = FUSE_APPEND_DEFAULT,  // 0x00, FUSE7_DEFAULT
  .BOOTEND = FUSE_BOOTEND_DEFAULT  // 0x00, FUSE8_DEFAULT
};
```

MCUdude:
> Out of curiosity, why is that limitation in place? A use case could reasonably want
```
$ avrdude ... -U unlock.elf -U beef.elf -U lock.elf
```

@stefanrueger couldn't this be solved quite easily by specifying the memories? Unlock, write all memories specified in beef.elf, and then lock again.

```
$ avrdude ... -U lock:w:unlock.elf -U beef.elf -U lock:w:lock.elf
```

McMCCRU:
> I am thinking of a new memory name called `all` and I think it is nice to have the following command. Any comments? This can be in a new enhancement.
> 
> ```
> avrdude -c xxx -p xxx -U:all:r:readhackall.elf
> ```

An ELF file can only be written to, but not read into. This can be confusing for users, best left as is for now.

MCUdude:
> An ELF file can only be written to, but not read into. This can be confusing for users, best left as is for now.

It's probably a lot of work, but I think he meant that Avrdude can read the necessary data from an AVR and create an ELF file based on the data. This way, it's very easy to duplicate chips. But this means that Avrdude has to be able to _create_ valid ELF files.

mcuee:
> but I think he meant that Avrdude can read the necessary data from an AVR and create an ELF file based on the data. This way, it's very easy to duplicate chips. But this means that Avrdude has to be able to _create_ valid ELF files.

Yes, that is what I mean. Thanks.

stefanrueger:
> @stefanrueger couldn't this be solved quite easily by specifying the memories? 

It can. This is a *hypothetical* example how people *might* want to use avrdude. There are better examples. Read on.

I still do not agree with a limit on the number of times `-U file.elf` can be specified. There is no such limitation on the number of times one can write to flash otherwise. And it would be *wrong* to limit that: for example, in,
```
$ avrdude ... -U flash:w:app.hex:i -U flash:w:pgmdata1.hex:i -U flash:w:pgmdata_2.hex:i -U flash:w:boot.hex:i
```
a user might legitimately write to four (or 40 for that matter) different flash areas from four different files. And these four different flash areas might live in four different .elf files, or?

The longer I think about the use of the global *updates* list the more I feel at unease about this: main is an *application* of libavrdude and not part of libavrdude itself. fileio.c is part of libavrdude. So, sharing a global variable between these two files is *not a good idea*. There is precedence, but over time, these incidents need to be reduced not increased.

I believe the right structure for implementing @McMCCRU's excellent idea for the *new* meaning of `-U file` is: 
 - While parsing the 'U' argument in main: if parse_op() expresses it is of the form `-U file` (Note 1) call a *new* function, say, fileio_memories() in fileio.c to receive a list of memories encoded in file (Note 2). Add to the `updates` list as many update nodes as there are memories (basically as if the user had written `-U memory1:w:file:a`,  `-U memory2:w:file:a`, ... `-U memoryN:w:file:a` instead)
 - Write code in fileio.c for ELF files so that `-U memory:w:a` extracts *only* the specified  memory (Note 3)

This approach does not need global variables, it keeps what happens transparent, it is vvv easy to document, and it does not limit what user may want to do (and what do we know how they might want to use avrdude?). This also might make it  unnecessary to define the max number of bytes for a memory section: if an .elf file has 217 bytes for fuses, the `do_op()` and the `avr_write()` function in there will notice anyway, not? If however, tomorrow, Microchip releases a part with 217 fuse bytes, we would not have to change avrdude. This approach comes at a cost that the ELF file will need to be opened and read *N* times rather than once, but hey that probably only takes a millisecond each time and the user won't notice. 

(Note 1): `parse_op` will need a modification to create a way of saying that the argument to `-U` was a single file; there are numerous possibilites of doing so, one would be to enter the memory "*" - but other techniques exist.

(Note 2): This could be done by opening file, checking whether it is an ELF and if so returning all memory strings found in the ELF file as a list of strings, eg, ("flash", "fuse2", "eeprom", "lock"). If it's not an ELF file then this function returns the list ("flash"), as the documentation says (or will say) `-U file` means `-U flash:w:file:a` except if file is an ELF file then it means write to all memories that there are in the ELF file.

(Note 3): This is probably harder than one might think, for example, it would need to consider the recently introduced memory aliases. So, if the .elf file defines fuse0 and fuse1, the user still might *only* want to write `-U bodcfg:w:file.elf:a` without knowing, or having to know, that `bodcfg` is an alias to `fuse1`. It is not difficult to implement, though: rather than comparing the ELF memory string with the user specified memory string, one would `avr_locate_mem()` both strings and compare the resulting `mem->desc` strings.

> I think we can leave this PR as it is. Future PRs may improve on this aspect if necessary.

I disagree. Given the PR in its current form, future improvement will need to change the structure substantially, so that it is better to get the structure right in the first place. I would also warn against rushing code into a new release that substantially changes what main.c does (`-U` processing, substantially modified in *another* file) without proper code review. Having only seen glimpses of the code, I agree that it is competently written, but it still may not do the right thing. I simply do not have the time right now to go through nearly 300 lines of code and all its ramifications.

Summarising, I believe this PR needs a complete rewrite considering all the points raised. I am convinced that a freshly started PR will weigh in at considerably fewer than 300 lines (which would mean that it is easier to review) with much crisper and clearer code and *without* the use of global variables.

McMCCRU:
> a user might legitimately write to four (or 40 for that matter) different flash areas from four different files. And these four different flash areas might live in four different .elf files, or?

It is not possible in the current version to make the implementation of writing to different parts of memory. This format is not meant to store configuration as you would like. Even AVR Studio can't do that...

mcuee:
>> I think we can leave this PR as it is. Future PRs may improve on this aspect if necessary.
>>
> I disagree. Given the PR in its current form, future improvement will need to change the structure substantially, so that it is better to get the structure right in the first place. I would also warn against rushing code into a new release that substantially changes what main.c does (-U processing, substantially modified in another file) without proper code review. Having only seen glimpses of the code, I agree that it is competently written, but it still may not do the right thing. I simply do not have the time right now to go through nearly 300 lines of code and all its ramifications.
> 
> Summarising, I believe this PR needs a complete rewrite considering all the points raised. I am convinced that a freshly started PR will weigh in at considerably fewer than 300 lines (which would mean that it is easier to review) with much crisper and clearer code and without the use of global variables.

I see. I will trust you here. 

And I also agree with you that changes to coommon files like `main.c` need to be carefully reviewed (or other files like `update.c` and `fileio.c`

mcuee:
> > a user might legitimately write to four (or 40 for that matter) different flash areas from four different files. And these four different flash areas might live in four different .elf files, or?
> 
> It is not possible in the current version to make the implementation of writing to different parts of memory. This format is not meant to store configuration as you would like. Even AVR Studio can't do that... You will also have to make your own version of toolchains, who will take on such work?

@McMCCRU 
Please help to explain why it needs customized toolchains. Thanks.

stefanrueger:
> who will take on such work?

A competent user?

McMCCRU:
> Please help to explain why it needs customized toolchains. Thanks.

And how can I write to different parts of flash memory? We have a .text section where the image for recording is located, it starts from address 0, and if I need to write from address 1024, for example, the subsequent image from the elf file? How to do it?

McMCCRU:
> > who will take on such work?
> 
> A competent user?

We need to orient on the existing capabilities of the AVR toolchain, its build scripts, there are not as many options that we can do as we would like.

stefanrueger:
> if I need to write from address 1024, for example, the subsequent image from the elf file? How to do it?

Have a look at how .hex files are read in fileio.c; look at the use of `mem->tags[]`. Hex files can occupy a series of non-overlapping memory segments with holes in between. (And as an aside, .hex files are generated from .elf, are they not?)

McMCCRU:
I focused on what can be obtained in a simple existing way. I highly doubt that anyone in production would want to use -U file.elf more than once or other than this command. This was what I was guided by to create this PR. If you think that this is not enough to include my code in the program, then this PR can be closed.

McMCCRU:
> Have a look at how .hex files are read in fileio.c; look at the use of `mem->tags[]`. Hex files can occupy a series of non-overlapping memory segments with holes in between. (And as an aside, .hex files are generated from .elf, are they not?)

To put an end to this issue, I suggest you try do it this, I can’t, the flash is constantly overwritten entirely by on new one.

stefanrueger:
I have no doubt that your code and, particularly your idea of utilising more .elf sections, has useful applications. What I would like to encourage is to think about this from two more perspectives
 - How AVRDUDE is documented
 - How users *might* want to use this based on the documentation

For your application, a user might say:  "Brilliant, I create an .elf file with all the relevant info, including how I wish to lock the device *after* writing all memories. Hang on a second, the documentation does not give me any guarantees that the lock bits are written last. I better use a tool to split my .elf file into contents.elf and locks.elf (with no or empty .text) and use `-U contents.elf -U lock.elf` in that order."

> the flash is constantly overwritten entirely by on new one

How then does the current `elf2b()` implementation manage to place `boot`, `apptable` and `application` into different 
areas of the flash memory?

mcuee:
This PR also works with ATmega4808.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195v1 -c pickit4_updi -p m4808 -U .\Blink_m408_fuse_eeprom.ino.ino.elf

                  Vtarget                      : 4.69 V
                  JTAG clock megaAVR/debug     : 100 kHz
                  PDI/UPDI clock Xmega/megaAVR : 100 kHz
avrdude_pr1195v1: AVR device initialized and ready to accept instructions
avrdude_pr1195v1: device signature = 0x1e9650 (probably m4808)
avrdude_pr1195v1: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1195v1: erasing chip
avrdude_pr1195v1: found in ELF eeprom, size = 32
avrdude_pr1195v1: found in ELF fuse0 = 0x00
avrdude_pr1195v1: found in ELF fuse1 = 0x54
avrdude_pr1195v1: found in ELF fuse2 = 0x02
avrdude_pr1195v1: found in ELF fuse4 = 0x00
avrdude_pr1195v1: found in ELF fuse5 = 0xC4
avrdude_pr1195v1: found in ELF fuse6 = 0x07
avrdude_pr1195v1: found in ELF fuse7 = 0x00
avrdude_pr1195v1: found in ELF fuse8 = 0x00
avrdude_pr1195v1: reading input file .\Blink_m408_fuse_eeprom.ino.ino.elf for flash
                  with 796 bytes in 1 section within [0, 0x31b]
                  using 7 pages and 100 pad bytes
avrdude_pr1195v1: writing 796 bytes flash ...

Writing | ################################################## | 100% 0.43 s

avrdude_pr1195v1: 796 bytes of flash written
avrdude_pr1195v1: verifying flash memory against .\Blink_m408_fuse_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.22 s

avrdude_pr1195v1: 796 bytes of flash verified
avrdude_pr1195v1: reading input file .\Blink_m408_fuse_eeprom.ino.ino.elf for eeprom
                  with 32 bytes in 1 section within [0, 0x1f]
                  using 1 page and 32 pad bytes
avrdude_pr1195v1: writing 32 bytes eeprom ...

Writing | ################################################## | 100% 0.06 s

avrdude_pr1195v1: 32 bytes of eeprom written
avrdude_pr1195v1: verifying eeprom memory against .\Blink_m408_fuse_eeprom.ino.ino.elf

Reading | ################################################## | 100% 0.02 s

avrdude_pr1195v1: 32 bytes of eeprom verified
avrdude_pr1195v1: reading input file 0x00 for fuse0/wdtcfg
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse0/wdtcfg ...
avrdude_pr1195v1: 1 byte of fuse0/wdtcfg written
avrdude_pr1195v1: verifying fuse0/wdtcfg memory against 0x00
avrdude_pr1195v1: 1 byte of fuse0/wdtcfg verified
avrdude_pr1195v1: reading input file 0x54 for fuse1/bodcfg
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse1/bodcfg ...
avrdude_pr1195v1: 1 byte of fuse1/bodcfg written
avrdude_pr1195v1: verifying fuse1/bodcfg memory against 0x54
avrdude_pr1195v1: 1 byte of fuse1/bodcfg verified
avrdude_pr1195v1: reading input file 0x02 for fuse2/osccfg
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse2/osccfg ...
avrdude_pr1195v1: 1 byte of fuse2/osccfg written
avrdude_pr1195v1: verifying fuse2/osccfg memory against 0x02
avrdude_pr1195v1: 1 byte of fuse2/osccfg verified
avrdude_pr1195v1: reading input file 0x00 for fuse4/tcd0cfg
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse4/tcd0cfg ...
avrdude_pr1195v1: 1 byte of fuse4/tcd0cfg written
avrdude_pr1195v1: verifying fuse4/tcd0cfg memory against 0x00
avrdude_pr1195v1: 1 byte of fuse4/tcd0cfg verified
avrdude_pr1195v1: reading input file 0xC4 for fuse5/syscfg0
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse5/syscfg0 ...
avrdude_pr1195v1: 1 byte of fuse5/syscfg0 written
avrdude_pr1195v1: verifying fuse5/syscfg0 memory against 0xC4
avrdude_pr1195v1: 1 byte of fuse5/syscfg0 verified
avrdude_pr1195v1: reading input file 0x07 for fuse6/syscfg1
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse6/syscfg1 ...
avrdude_pr1195v1: 1 byte of fuse6/syscfg1 written
avrdude_pr1195v1: verifying fuse6/syscfg1 memory against 0x07
avrdude_pr1195v1: 1 byte of fuse6/syscfg1 verified
avrdude_pr1195v1: reading input file 0x00 for fuse7/append
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse7/append ...
avrdude_pr1195v1: 1 byte of fuse7/append written
avrdude_pr1195v1: verifying fuse7/append memory against 0x00
avrdude_pr1195v1: 1 byte of fuse7/append verified
avrdude_pr1195v1: reading input file 0x00 for fuse8/bootend
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse8/bootend ...
avrdude_pr1195v1: 1 byte of fuse8/bootend written
avrdude_pr1195v1: verifying fuse8/bootend memory against 0x00
avrdude_pr1195v1: 1 byte of fuse8/bootend verified

avrdude_pr1195v1 done.  Thank you.

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195v1 -c pickit4_updi -p m4808 -qqt
avrdude> dump fuses
>>> dump fuses
0000  00 54 02 ff 00 c4 07 00  00                       |.T.......       |

avrdude> dump eeprom 0 0x20
>>> dump eeprom 0 0x20
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 00 00 00  |Hello World!....|
0010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|

avrdude> quit
>>> quit
avrdude>

```

McMCCRU:
> How then does the current `elf2b()` implementation manage to place `boot`, `apptable` and `application` into different
> areas of the flash memory?

Look at the elf_mem_limits() function, everything starts from address 0...
Screenshot_20221129_152917
This function lists the general memory map for AVR, as it is used by the -U xxx:x:file.elf commands, you cannot use other commands that are not indicated in it.

stefanrueger:
> This PR also works with ATmega4808

What happens if you want to pick just `bodcfg` from the file? `avrdude ... -U bodcfg:w:file.elf:a`

stefanrueger:
> everything starts from address 0

Do .elf sections have a size? Can the size be 0? Can a section be left off?
```
$ avr-readelf my.elf -S | grep -P "\[(Nr| 2)\]"
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 2] .text             PROGBITS        00000000 000074 0008e2 00  AX  0   0  2
```

stefanrueger:
In which order are the memories mentioned in an .elf file written? At which position/place are the lock bits written?

McMCCRU:
> What happens if you want to pick just `bodcfg` from the file? `avrdude ... -U bodcfg:w:file.elf:a`

This command is not in the list of the elf_mem_limits() function, avrdude will give an error.

mcuee:
> > This PR also works with ATmega4808
> 
> What happens if you want to pick just `bodcfg` from the file? `avrdude ... -U bodcfg:w:file.elf:a`

I changed the fuse setting first and then run the above command. And it works as expected.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195v1 -c pickit4_updi -p m4808 -qqt
avrdude> dump fuses
>>> dump fuses
0000  00 54 02 ff 00 c4 07 00  00                       |.T.......       |

avrdude> dump bodcfg
>>> dump bodcfg
0000  54                                                |T               |

avrdude> write bodcfg 0 0x00
>>> write bodcfg 0 0x00
avrdude> flush
>>> flush
avrdude> dump bodcfg
>>> dump bodcfg
0000  00                                                |.               |

avrdude> dump fuses
>>> dump fuses
0000  00 00 02 ff 00 c4 07 00  00                       |.........       |

avrdude> quit
>>> quit
avrdude>
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195v1 -c pickit4_updi -p m4808 -U bodcfg:w:.\Blink_m408_fuse_eeprom.ino.ino.elf:a

                  Vtarget                      : 4.67 V
                  JTAG clock megaAVR/debug     : 100 kHz
                  PDI/UPDI clock Xmega/megaAVR : 100 kHz
avrdude_pr1195v1: AVR device initialized and ready to accept instructions
avrdude_pr1195v1: device signature = 0x1e9650 (probably m4808)
avrdude_pr1195v1: reading input file .\Blink_m408_fuse_eeprom.ino.ino.elf for fuse1/bodcfg
                  with 1 byte in 1 section within [0, 0]
avrdude_pr1195v1: writing 1 byte fuse1/bodcfg ...
avrdude_pr1195v1: 1 byte of fuse1/bodcfg written
avrdude_pr1195v1: verifying fuse1/bodcfg memory against .\Blink_m408_fuse_eeprom.ino.ino.elf
avrdude_pr1195v1: 1 byte of fuse1/bodcfg verified

avrdude_pr1195v1 done.  Thank you.

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1195v1 -c pickit4_updi -p m4808 -qqt
avrdude> dump bodcfg
>>> dump bodcfg
0000  54                                                |T               |

avrdude> quit
>>> quit
avrdude>

```

mcuee:
> > What happens if you want to pick just `bodcfg` from the file? `avrdude ... -U bodcfg:w:file.elf:a`
> 
> This command is not in the list of the elf_mem_limits() function, avrdude will give an error.

It works because that `bodcfg` is an alias for `fuse1`.

```
    memory "bodcfg"
        alias "fuse1";
    ;
```

McMCCRU:
I understand that no one has been supporting elf format in project for a long time, even in the existing implementation there are errors, for example, if the elf contains not only the .text section, then data from other sections will be written to the end of the data for flash. This can be a critical situation and will result in a programming error if this data goes beyond the size of the flash memory. I also understand that this format is practically not used by ordinary users, however, the factory that produces our device asked for an elf file with all the contents for fuses and lockbits, because do not want to deal with individual settings when programming. I don't know what they use as a programmer and software, but the conditions said about AVR Studio. I do not work with this development environment, and it was not possible to check the operation of the elf file, but I found a working version with avrdude, they checked at the factory that everything is programmed as it should and gave a positive answer...

stefanrueger:
> [order] flash, eeprom, fuses and lockbits

This should be put into the documentation of `-U file.elf`. How would a user know otherwise? [Falling short of reading 60k lines of code]

> in the existing implementation there are errors

I appreciate this. And this may be one reason why this is not a popular input format for flash. Let's put this right!

Does the PR consider .data, the section that contains the initialised data, which reside in flash? 

The avr-toolchain generates a hex file from an .elf file by taking all PROGBITS sections except `.comment` and `.debug_*` and packs these into the flash-type memory (flash, application, apptable, boot) contents from, as you correctly say address 0:
```
$ avr-objcopy -O ihex -R my.elf my.hex
$ avr-readelf -S my.elf | grep PROGBITS
  [ 1] .data             PROGBITS        00800100 007e54 000010 00  WA  0   0  1
  [ 2] .text             PROGBITS        00000000 000094 001032 00  AX  0   0  2
  [ 3] .progdata         PROGBITS        00001032 0010c6 006d8e 00   A  0   0  1
  [ 5] .comment          PROGBITS        00000000 007e64 000011 01  MS  0   0  1
  [ 7] .debug_info       PROGBITS        00000000 007eb8 0005f4 00      0   0  1
  [ 8] .debug_abbrev     PROGBITS        00000000 0084ac 0005a2 00      0   0  1
  [ 9] .debug_line       PROGBITS        00000000 008a4e 00001a 00      0   0  1
  [10] .debug_str        PROGBITS        00000000 008a68 000208 00      0   0  1
```

Note that I often create a .progdata section in flash within my avr C sources eg, 
```
const double KeplerElements[N] __attribute__((section (".progdata"))) = { 
/* data in flash go here */
};
```

The name `.progdata` is irrelevant and user-defined, but it is packed into flash by the toolchain. The generated .hex file contains above three sections, `.data`, `.text` and `.progdata`. So, when I use  `-U flash:w:my.elf` can I expect all PROGBITS sections to be there with the exception of `.comment` and all `.debug_*`?

In this case I would expect flash to contain 32208 = 0x7e64-0x94 bytes from address 0 onwards, first, `.text`, then `.progdata`, then `.data`. It may well be that the current .elf implementation also forgot `.data` and user-defined flash sections. One more reason to put things right.

@mcuee: could you create a mildly complex input .elf file with a non-empty data section and see whether it is written to flash? The .hex file should have the .data section, and at least when you write the .hex that should write the .data section too.

I still think it is unnecessary to make `updates` global :)

McMCCRU:
The content for the flash is only taken from .text, not my idea. The .data section is usually empty, it is used to count the used RAM memory, and nothing else. I made a command interpreter based on reading additional sections, nothing more. This format is difficult to work with, it is not flexible, however, in production it is a popular format and they do not like hex / bin there, because you can’t plant a monkey that will stupidly flash the MCU without going into details

McMCCRU:
> avr-readelf -S my.elf | grep PROGBITS

Try another command:
```
avr-objdump -s -j .text -j .data -j .progdata my.elf
```

McMCCRU:
> I still think it is unnecessary to make `updates` global :)

In fact, it is already global, although it was used only within the boundaries of main.c, I just don’t see an easy way to queue commands for execution.

stefanrueger:
> The .data section is usually empty, it is used to count the used RAM memory, and nothing else.

*Decidedly no.* The .data section contains the *initialised* SRAM data of the program. It is *vital* that is is copied to flash! If you use a global variable `int hello = 12;` in your sketch, then the int `12` ends up in `.data`. The compiler puts code into the application that, before *main()* is started, the to-be-initialised global variables are initialised from `.data`. Leave `.data` out and `0xff`s are copied into the global variable (or whatever was there in flash before if it wasn't erased), so `hello` has the initial value of -1 in the program.

> I just don’t see an easy way to queue commands for execution

  - While parsing the 'U' argument in main: if parse_op() expresses it is of the form `-U file` (Note 1) call a *new* function, say, fileio_memories() in fileio.c to receive a list of memories encoded in file (Note 2). Add to the `updates` list as many update nodes as there are memories (basically as if the user had written `-U memory1:w:file:a`,  `-U memory2:w:file:a`, ... `-U memoryN:w:file:a` instead)
 - Write code in fileio.c for ELF files so that `-U memory:w:file:a` extracts *only* the specified  memory (Note 3)

stefanrueger:
> > In which order are the memories mentioned in an .elf file written? At which position/place are the lock bits written?

> I did so, flash, eeprom, fuses and lockbits. Everything is written based on information from avrdude.conf.

Re-reading your reply, it occurs to me that you might mean "the order of writing `flash`, `eeprom`, `fuses` and `lockbits` is based on the order in which they appear in `avrdude.conf`"? That would not be good. Imagine a user defines their part with the `lock` memory first. If the lock bits are written first, and they secure the device against reading and writing, would the rest of the command work out? IMO it makes sense to guarantee the order `flash`, `eeprom`, `fuses and `lockbits`... not?

stefanrueger:
> The content for the flash is only taken from .text, not my idea. 

It doesn't matter whose idea that was. I think it is still wrong, even if it is existing code.

McMCCRU:
> Re-reading your reply, it occurs to me that you might mean "the order of writing `flash`, `eeprom`, `fuses` and `lockbits` is based on the order in which they appear in `avrdude.conf`"? That would not be good. Imagine a user defines their part with the `lock` memory first. If the lock bits are written first, and they secure the device against reading and writing, would the rest of the command work out? IMO it makes sense to guarantee the order `flash`, `eeprom`, `fuses` and `lockbits`... not?

Why should I decide in which order the user uses -U? I only added write commands on based sections elf file, it's logical to first write to "flash", then "eeprom", then "fuses" and at the very end "lockbits". I have been trying to prove for a long time that it is not desirable to use the recording of the full elf file with other commands together, otherwise there is no point in using it and this PR.

McMCCRU:
> * Write code in fileio.c for ELF files so that `-U memory:w:file:a` extracts _only_ the specified  memory (Note 3)

This already works in the existing code, but there is no check for sections, as a result, data from other sectors, not .text and not .data, will be written immediately after them to flash.

stefanrueger:
Look, I am really trying to help integrating this PR with AVRDUDE. Here my advice:
 - The flash extract of the .ELF file should be the same as that of the corresponding hex file, typically created with the standard toolchain command `avr-objcopy -O ihex -R .eeprom my.elf my.hex`
    + In particular `.text`, `.data` *and all user-defined sections for flash* should be copied over (this excludes all `.comment` and `.debug_*` sections, see also Issue #1204)
 - There is no need to restrict the number of `-U file.elf` commands on the command line
 - `updates` should not be referenced outside `main.c`; if, eg, `updates` was used in fileio.c then the external reference ends up in libavrdude; this leads to an unresolved symbol for projects that use libavrdude to create their own avrdude application (think a GUI version of command-line AVRDUDE); main.c is not part of libavrdude
 - The number of fuse bytes/lock bytes in the .ELF file does not need be limited by some arbitrary number, AVRDUDE's `avr_write()` can take care of too big sizes
 - Have a defined order in which the memories in the .ELF file are written to the device (this PR already does, good!)
     + It makes sense to write `lock` last (you say you already do, good!)
 - Document the new syntax of `-U file` and the order in which the .ELF memories are written on encountering `-U file.elf`

Once you think this PR is ready, someone will need to do a *careful* and *thorough* code review (@dl8dtl?), functionality testing (@mcuee?) and do-not-surprise-the-user testing (@MCUdude?)

I will no longer be available to help (am busy otherwise), but I am sure the team will look after this PR.

mariusgreuel:
I have been late to the party, and other issues aside, I am a little ambivalent about this PR: I very much like the simple form `-U file.elf`, however, that is does the fuses, too, is scary.

As `-U file.elf` is much more intuitive and less typing than the explicit form, it is probably going to be adopted by users soon.

However, many users are likely not aware that they can break their devices by writing the wrong fuses and lock bits. In fact, many people never need to do so at all. So when you compile other peoples C code (which is what Arduino folks do) that has the wrong section pragma somewhere, it is going to end up in disaster.

For fuses/lock bits, I tend to say that explicit is better than implicit. Perhaps some `--include-fuses-and-lock-bits-too` option is in order?

mcuee:
> For fuses/lock bits, I tend to say that explicit is better than implicit. Perhaps some --include-fuses-and-lock-bits-too option is in order?

I tend to agree with fuses and lock bits. But then people may argue for EEPROM as well. Maybe we can skip these things by default but then allow the users to force the operation. Maybe three extra options can be added to allow the writing of fuses, lock bits and EEPROM.

stefanrueger:
The main trouble with `-U filename` is that it has the documented meaning of `-U flash:w:filename:a`. Making it now populate all memories present in `-U file.elf` would break documented use and possibly upset users who regularly have been using `-U file.elf` in the existing, documented use.

A better syntax would be to invent a new memory `all` as in `-U all:w:file.elf:a` with the meaning of addressing all present memories. Selecting memories as in `-U eeprom:w:file.elf:a` or `-U efuse:w:file.elf:a` should work, too, as it probably already does. That way we don't need to invent a lot of long options.

mariusgreuel:
> The main trouble with -U filename is that it has the documented meaning of -U flash:w:filename:a

Ah, understood, I did not know that this shortcut exists today. Then it is pretty obvious that we must not change the semantics of `-U`!

mcuee:
> A better syntax would be to invent a new memory all as in -U all:w:file.elf:a with the meaning of addressing all present memories. Selecting memories as in -U eeprom:w:file.elf:a or -U efuse:w:file.elf:a should work, too, as it probably already does. That way we don't need to invent a lot of long options.

I tend to agree with this new memory option `all`. But I tend to think we need to have enough level of warnings about this option as it includes fuse and lock bits.

mcuee:
@McMCCRU 

It has been a while. Just wondering if you intend to further work on this PR or not, based on the comments from @stefanrueger.

mcuee:
@stefanrueger 

I think we all agree that this is a nice enhancement to avrdude. It is a pity that @McMCCRU did not come back with the reply. Just wondering if you want to continue from here or even create another PR to replace this one.

stefanrueger:
I'd rather not try to modify this PR, as it's not the right direction. I have no doubt that it works for what its author intended, but it doesn't fit well into the AVRDUDE structure.  Amongst the 60 or so open issues there far more significant  improvements to be had.  So low prio for me.

This idea does not bring about new functionality. Basically, `avrdude -U all:w:file.elf:a` is an abbreviation for  sth like
```
avrdude -U eeprom:w:blink.elf:a -U flash:w:blink.elf:a -U lfuse:w:blink.elf:a -U hfuse:w:blink.elf:a \
  -U efuse:w:blink.elf:a -U lock:w:blink.elf:a -U signature:w:blink.elf:a -U calibration:w:blink.elf:a
```
assuming a `ATmega328p` part. If the user doesn't know which memories a part has, a little bash-fu can help:
```
   avrdude `avrdude -pATmega328p/At | grep ^.ptmm.A | cut -f3 | uniq | sed "s/.*/-U &:w:blink.elf:a/"`
```
which is what I did to generate above line.

I suggest closing this PR, and coming back to the idea once the issue queue is somewhat shorter.

mcuee:
Some further discussions here.
* #1354

PR #1206, opened 2022-12-01
Fix reading of .elf files

stefanrueger:
This PR fixes Issue #1204

`elf2b()` only considers the first suitable section in any one program segment. However, the ELF format allows for a series of sections in segments; this PR fixes `elf2b()`'s deficiency.

The PR consists of 2 commits, the first of which rewrites `elf2b()` in an equivalent way by pulling in the function `elf_get_scn()` that used to find the *first* suitable section of a program segment.

The second commit changes the behaviour of `elf2b()` to do the right thing.

### DIFF
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -118,10 +118,9 @@ char * fileio_fmtstr(FILEFMT format)
 }
 
 
-static int b2ihex(const unsigned char *inbuf, int bufsize,
-           int recsize, int startaddr,
-           const char *outfile, FILE *outf, FILEFMT ffmt)
-{
+static int b2ihex(const unsigned char *inbuf, int bufsize, int recsize,
+  int startaddr, const char *outfile_unused, FILE *outf, FILEFMT ffmt) {
+
   const unsigned char *buf;
   unsigned int nextaddr;
   int n, nbytes, n_64k;
@@ -397,10 +396,9 @@ static int ihex2b(const char *infile, FILE *inf,
   }
 }
 
-static int b2srec(const unsigned char *inbuf, int bufsize,
-           int recsize, int startaddr,
-           const char *outfile, FILE *outf)
-{
+static int b2srec(const unsigned char *inbuf, int bufsize, int recsize,
+  int startaddr, const char *outfile_unused, FILE *outf) {
+
   const unsigned char *buf;
   unsigned int nextaddr;
   int n, nbytes, addr_width;
@@ -767,10 +765,9 @@ static int elf_mem_limits(const AVRMEM *mem, const AVRPART *p,
 }
 
 
-static int elf2b(const char *infile, FILE *inf,
-                 const AVRMEM *mem, const AVRPART *p,
-                 int bufsize, unsigned int fileoffset)
-{
+static int elf2b(const char *infile, FILE *inf, const AVRMEM *mem,
+  const AVRPART *p, int bufsize_unused, unsigned int fileoffset_unused) {
+
   Elf *e;
   int rv = 0, size = 0;
   unsigned int low, high, foff;
@@ -1045,9 +1042,9 @@ static int fileio_rbin(struct fioparms *fio,
 }
 
 
-static int fileio_imm(struct fioparms *fio,
-             const char *fname, FILE *f, const AVRMEM *mem, int size)
-{
+static int fileio_imm(struct fioparms *fio, const char *fname, FILE *f_unused,
+ const AVRMEM *mem, int size) {
+
   int rc = 0;
   char *e, *p, *filename;
   unsigned long b;
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -772,7 +772,7 @@ static int elf2b(const char *infile, FILE *inf,
                  int bufsize, unsigned int fileoffset)
 {
   Elf *e;
-  int rv = -1;
+  int rv = 0, size = 0;
   unsigned int low, high, foff;
 
   if (elf_mem_limits(mem, p, &low, &high, &foff) != 0) {
@@ -897,6 +897,7 @@ static int elf2b(const char *infile, FILE *inf,
 
       if (sh == NULL) {
         pmsg_error("unable to read section #%u header: %s\n", (unsigned int) ndx, elf_errmsg(-1));
+        rv = -1;
         continue;
       }
       // Only interested in PROGBITS, ALLOC sections
@@ -927,8 +928,9 @@ static int elf2b(const char *infile, FILE *inf,
        * from it, using the "foff" offset obtained above.
        */
       if (mem->size != 1 && sh->sh_size > (unsigned) mem->size) {
-        pmsg_error("section %s does not fit into %s memory:\n"
-          "    0x%x + %u > %u\n", sname, mem->desc, lma, sh->sh_size, mem->size);
+        pmsg_error("section %s of size %u does not fit into %s of size %d\n",
+          sname, sh->sh_size, mem->desc, mem->size);
+        rv = -1;
         continue;
       }
 
@@ -939,31 +941,38 @@ static int elf2b(const char *infile, FILE *inf,
         if (mem->size == 1) {
           if (d->d_off != 0) {
             pmsg_error("unexpected data block at offset != 0\n");
+            rv = -1;
           } else if (foff >= d->d_size) {
             pmsg_error("ELF file section does not contain byte at offset %d\n", foff);
+            rv = -1;
           } else {
             msg_notice2("    Extracting one byte from file offset %d\n", foff);
             mem->buf[0] = ((unsigned char *)d->d_buf)[foff];
             mem->tags[0] = TAG_ALLOCATED;
-            rv = 1;
+            size = 1;
           }
         } else {
-          unsigned int idx;
+          int idx = lma-low + d->d_off;
+          int end = idx + d->d_size;
 
-          idx = lma - low + d->d_off;
-          if ((int)(idx + d->d_size) > rv)
-            rv = idx + d->d_size;
-          msg_debug("    Writing %ld bytes to mem offset 0x%x\n",
-            (long) d->d_size, idx);
-          memcpy(mem->buf + idx, d->d_buf, d->d_size);
-          memset(mem->tags + idx, TAG_ALLOCATED, d->d_size);
+          if(idx >= 0 && idx < mem->size && end >= 0 && end <= mem->size && end-idx >= 0) {
+            if (end > size)
+              size = end;
+            imsg_debug("writing %d bytes to mem offset 0x%x\n", end-idx, idx);
+            memcpy(mem->buf + idx, d->d_buf, end-idx);
+            memset(mem->tags + idx, TAG_ALLOCATED, end-idx);
+          } else {
+            pmsg_error("section %s [0x%04x, 0x%04x] does not fit into %s [0, 0x%04x]\n",
+              sname, idx, (int) (idx + d->d_size-1), mem->desc, mem->size-1);
+            rv = -1;
+          }
         }
       }
     }
   }
 done:
   (void)elf_end(e);
-  return rv;
+  return rv<0? rv: size;
 }
 #endif  /* HAVE_LIBELF */
 
--- a/src/fileio.c
+++ b/src/fileio.c
@@ -708,41 +708,6 @@ int is_section_in_segment(Elf32_Shdr *sh, Elf32_Phdr *ph)
     return 1;
 }
 
-/*
- * Return the ELF section descriptor that corresponds to program
- * header `ph'.  The program header is expected to be of p_type
- * PT_LOAD, and to have a nonzero p_filesz.  (PT_LOAD sections with a
- * zero p_filesz are typically RAM sections that are not initialized
- * by file data, e.g. ".bss".)
- */
-static Elf_Scn *elf_get_scn(Elf *e, Elf32_Phdr *ph, Elf32_Shdr **shptr)
-{
-  Elf_Scn *s = NULL;
-
-  while ((s = elf_nextscn(e, s)) != NULL) {
-    Elf32_Shdr *sh;
-    size_t ndx = elf_ndxscn(s);
-    if ((sh = elf32_getshdr(s)) == NULL) {
-      pmsg_error("unable to read section #%u header: %s\n", (unsigned int)ndx, elf_errmsg(-1));
-      continue;
-    }
-    if ((sh->sh_flags & SHF_ALLOC) == 0 ||
-        sh->sh_type != SHT_PROGBITS)
-      /* we are only interested in PROGBITS, ALLOC sections */
-      continue;
-    if (sh->sh_size == 0)
-      /* we are not interested in empty sections */
-      continue;
-    if (is_section_in_segment(sh, ph)) {
-      /* yeah, we found it */
-      *shptr = sh;
-      return s;
-    }
-  }
-
-  pmsg_error("cannot find a matching section for program header entry @p_vaddr 0x%x\n", ph->p_vaddr);
-  return NULL;
-}
 
 static int elf_mem_limits(const AVRMEM *mem, const AVRPART *p,
                           unsigned int *lowbound,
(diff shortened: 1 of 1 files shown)


mcuee:
Tested here.
* https://github.com/avrdudes/avrdude/issues/1204#issuecomment-1335300564

stefanrueger:
Turns out `elf2b()` only checked the size of segments against memory size, but did not consider that the segment offset, eg of .data, can push it beyond memory limits. For testing make the PROGMEM array (nearly) fill 32 kB, then observe:
```
$ avrdude -qqU morse.elf
avrdude error: section .data [0x7ffe, 0x8017] does not fit into flash [0, 0x7fff]
avrdude error: read from file morse.elf failed
```

Also, en passant, improved error messaging.

stefanrueger:
I am pretty sure 0x7ffff as upper end of flash as suggested here https://github.com/avrdudes/avrdude/blob/ea2d7eec95b24d2b9d8f969c0e0f60c88b1b491d/src/fileio.c#L763-L769 is actually a bug: 0x7Fffff makes more sense as the `jmp`/`call` commands can use 22-bit for the *word* address (and the comment suggests that too); I realise in practice there is no 8-bit AVR part that has even 512 kB flash, so this bug has had no practical consequence so far.

mcuee:
> Turns out `elf2b()` only checked the size of segments against memory size, but did not consider that the segment offset, eg of .data, can push it beyond memory limits. For testing make the PROGMEM array (nearly) fill 32 kB, then observe:
> 
> ```
> $ avrdude -qqU morse.elf
> avrdude error: section .data [0x7ffe, 0x8017] does not fit into flash [0, 0x7fff]
> avrdude error: read from file morse.elf failed
> ```
> 
> Also, en passant, improved error messaging.

@stefanrueger 

Yes this seems to work fine. I change linker option to add `-Wl,--section-start=.progdata=0x07ff8` so that `.progdata` exceeds the flash range. This PR works fine for this test case.

@dl8dtl 
Do you think this is a limitation of avr binutils or a "feature"? I thought the linker would emit an error message.

```
$ make
avr-gcc -Os -DF_APU=16000000UL -mmcu=atmega328p -c -o led.o led.c
avr-gcc -mmcu=atmega328p -Wl,--section-start=.progdata=0x07ff8 -o led.elf led.o
avr-objcopy -O ihex -R .eeprom -R .fuse led.elf led.hex

$ avr-readelf -S led.elf
There are 15 section headers, starting at offset 0x23ec:

Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .data             PROGBITS        00800100 0001a4 000012 00  WA  0   0  1
  [ 2] .progdata         PROGBITS        00007ff8 0001b6 00000e 00   A  0   0  1
  [ 3] .text             PROGBITS        00000000 0000d4 0000d0 00  AX  0   0  2
  [ 4] .eeprom           PROGBITS        00810000 0001c4 000020 00  WA  0   0  1
  [ 5] .fuse             PROGBITS        00820000 0001e4 000003 00  WA  0   0  1
  [ 6] .comment          PROGBITS        00000000 0001e7 000011 01  MS  0   0  1
  [ 7] .note.gnu.av[...] NOTE            00000000 0001f8 000040 00      0   0  4
  [ 8] .debug_info       PROGBITS        00000000 000238 0005f4 00      0   0  1
  [ 9] .debug_abbrev     PROGBITS        00000000 00082c 0005a2 00      0   0  1
  [10] .debug_line       PROGBITS        00000000 000dce 00001d 00      0   0  1
  [11] .debug_str        PROGBITS        00000000 000deb 000208 00      0   0  1
  [12] .symtab           SYMTAB          00000000 000ff4 000c40 10     13 135  4
  [13] .strtab           STRTAB          00000000 001c34 000725 00      0   0  1
  [14] .shstrtab         STRTAB          00000000 002359 000092 00      0   0  1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), p (processor specific)

$ ./avrdude_pr1206v1 -c usbasp -p m328p -U led.elf -q

avrdude_pr1206v1: AVR device initialized and ready to accept instructions
avrdude_pr1206v1: device signature = 0x1e950f (probably m328p)
avrdude_pr1206v1: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1206v1: erasing chip
avrdude_pr1206v1 error: section .progdata [0x7ff8, 0x8005] does not fit into flash [0, 0x7fff]
avrdude_pr1206v1 error: read from file led.elf failed

avrdude_pr1206v1 done.  Thank you.

$ ./avrdude_pr1206v1 -c usbasp -p m328p -U led.hex -q

avrdude_pr1206v1: AVR device initialized and ready to accept instructions
avrdude_pr1206v1: device signature = 0x1e950f (probably m328p)
avrdude_pr1206v1: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1206v1: erasing chip
avrdude_pr1206v1 error: address 0x8006 out of range at line 16 of led.hex
avrdude_pr1206v1 error: read from file led.hex failed

avrdude_pr1206v1 done.  Thank you.
```

stefanrueger:
BTW @dl8dtl
 - Does avr-gcc use the same address space for the modern parts (those with UPDI i/face) as below?
 - If not, do the modern parts have a simple table or does that depend on the part?

Implemented in `elf_mem_limits()`:

|Low|High|Memories|
| --: | --: | :-- |
| 0x00000000 | 0x0007FFFF | Flash |
| 0x00800000 | 0x0080FFFF |RAM |
| 0x00810000 | 0x0081FFFF | EEPROM |
| 0x00820000 | 0x0082FFFF | Fuses |
| 0x00830000 | 0x0083FFFF | Lock byte(s) |

mcuee:
> I am pretty sure 0x7ffff as upper end of flash as suggested here
> 
> https://github.com/avrdudes/avrdude/blob/ea2d7eec95b24d2b9d8f969c0e0f60c88b1b491d/src/fileio.c#L763-L769
> 
> is actually a bug: 0x7Fffff makes more sense as the `jmp`/`call` commands can use 22-bit for the _word_ address (and the comment suggests that too); I realise in practice there is no 8-bit AVR part that has even 512 kB flash, so this bug has had no practical consequence so far.

Or is it actually the code is correct but at the time the author forgot to change the comment? Maybe @dl8dtl can shed some light.

stefanrueger:
>  is it actually the code is correct (no AVR parts have more than 512KB flash)

The [jmp](https://onlinedocs.microchip.com/pr/GUID-0B644D8F-67E7-49E6-82C9-1B2B9ABE6A0D-en-US-19/index.html?GUID-66E47A38-76B8-4703-A78B-C30BD8F7CF66) command can use 22-bit for the word address, so need 23 bit (0x7Fffff) for the flash address space

It's more interesting to figure out whether `elf_mem_limits()` need adapting for newer parts (conceivably, avr-gcc might map to their linear address space using negative addresses, ie, high bit set)

> thought the linker would emit an error message

IIRC, one doesn't necessarily need to specify the part for which to compile, it is sufficient to say how "evolved" the part is (there are different avr classes, eg, as whether the part knows `adiw`, `spm` etc). So, although the Gnu compiler collection *could* know the size of memories when the user specifies the part fully, it doesn't use that info to the fullest possible extent.

mcuee:
@stefanrueger 

I believe the newer UPDI parts are still following the  same address space.

```
MINGW64 /c/Users/xiaof/My Documents/Arduino/Nano4808/Blink_eeprom.ino
$ avr-readelf -S Blink_eeprom.ino.ino.elf
There are 19 section headers, starting at offset 0x8d10:

Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .data             PROGBITS        00802800 0003d1 000000 00  WA  0   0  1
  [ 2] .text             PROGBITS        00000000 0000b4 0002f4 00  AX  0   0  2
  [ 3] .bss              NOBITS          00802800 0003a8 000004 00  WA  0   0  1
  [ 4] .eeprom           PROGBITS        00810000 0003a8 000020 00  WA  0   0  1
  [ 5] .fuse             PROGBITS        00820000 0003c8 000009 00  WA  0   0  1
  [ 6] .comment          PROGBITS        00000000 0003d1 000011 01  MS  0   0  1
  [ 7] .note.gnu.av[...] NOTE            00000000 0003e4 000040 00      0   0  4
  [ 8] .debug_aranges    PROGBITS        00000000 000428 0000a8 00      0   0  8
  [ 9] .debug_info       PROGBITS        00000000 0004d0 003751 00      0   0  1
  [10] .debug_abbrev     PROGBITS        00000000 003c21 002448 00      0   0  1
  [11] .debug_line       PROGBITS        00000000 006069 000889 00      0   0  1
  [12] .debug_frame      PROGBITS        00000000 0068f4 000098 00      0   0  4
  [13] .debug_str        PROGBITS        00000000 00698c 00151a 00      0   0  1
  [14] .debug_loc        PROGBITS        00000000 007ea6 000220 00      0   0  1
  [15] .debug_ranges     PROGBITS        00000000 0080c6 000030 00      0   0  1
  [16] .shstrtab         STRTAB          00000000 008c4e 0000c2 00      0   0  1
  [17] .symtab           SYMTAB          00000000 0080f8 0006b0 10     18  34  4
  [18] .strtab           STRTAB          00000000 0087a8 0004a6 00      0   0  1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), p (processor specific)
```

mcuee:
Same for ATTiny817 using the code here.
* https://github.com/avrdudes/avrdude/pull/1195#issuecomment-1330388013

```
$ avr-readelf -S Blink_Pro817_eeprom_fuse.ino.elf
There are 19 section headers, starting at offset 0x7418:

Section Headers:
  [Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            00000000 000000 000000 00      0   0  0
  [ 1] .data             PROGBITS        00803e00 0002b1 000000 00  WA  0   0  1
  [ 2] .text             PROGBITS        00000000 0000b4 0001d4 00  AX  0   0  2
  [ 3] .bss              NOBITS          00803e00 000288 00000a 00  WA  0   0  1
  [ 4] .eeprom           PROGBITS        00810000 000288 000020 00  WA  0   0  1
  [ 5] .fuse             PROGBITS        00820000 0002a8 000009 00  WA  0   0  1
  [ 6] .comment          PROGBITS        00000000 0002b1 000011 01  MS  0   0  1
  [ 7] .note.gnu.av[...] NOTE            00000000 0002c4 00003c 00      0   0  4
  [ 8] .debug_aranges    PROGBITS        00000000 000300 000070 00      0   0  8
  [ 9] .debug_info       PROGBITS        00000000 000370 002eff 00      0   0  1
  [10] .debug_abbrev     PROGBITS        00000000 00326f 001c02 00      0   0  1
  [11] .debug_line       PROGBITS        00000000 004e71 000640 00      0   0  1
  [12] .debug_frame      PROGBITS        00000000 0054b4 000064 00      0   0  4
  [13] .debug_str        PROGBITS        00000000 005518 00127f 00      0   0  1
  [14] .debug_loc        PROGBITS        00000000 006797 00017f 00      0   0  1
  [15] .debug_ranges     PROGBITS        00000000 006916 000038 00      0   0  1
  [16] .shstrtab         STRTAB          00000000 007356 0000c2 00      0   0  1
  [17] .symtab           SYMTAB          00000000 006950 0005f0 10     18  34  4
  [18] .strtab           STRTAB          00000000 006f40 000416 00      0   0  1
Key to Flags:
  W (write), A (alloc), X (execute), M (merge), S (strings), I (info),
  L (link order), O (extra OS processing required), G (group), T (TLS),
  C (compressed), x (unknown), o (OS specific), E (exclude),
  D (mbind), p (processor specific)

```

mcuee:
And from the DFPs, I can see that AVR DA/DB/DD parts will use the same linker option as either `avrxmega2` or `avrxmega3` or `avrxmega4`. So they will follow the same linker option as well.
C:\Program Files (x86)\Atmel\Studio\7.0\packs\atmel\AVR-Dx_DFP\2.1.146\gcc\dev

stefanrueger:
So, it would make sense to extend `elf_mem_limits()` to

|Low|High|Memories|
| --: | --: | :-- |
| 0x00000000 | 0x007FFFFF | Flash |
| 0x00800000 | 0x0080FFFF |RAM |
| 0x00810000 | 0x0081FFFF | EEPROM (*) |
| 0x00820000 | 0x0082FFFF | Fuses |
| 0x00830000 | 0x0083FFFF | Lock bytes |
| 0x00840000 | 0x0084FFFF | Signature |
| 0x00850000 | 0x0085FFFF | User signatures (*) |

(*) Not available in all parts

stefanrueger:
Would be interesting to know which of the memories https://github.com/avrdudes/avrdude/blob/ea2d7eec95b24d2b9d8f969c0e0f60c88b1b491d/src/avr.c#L1327-L1339 map to the table above, and where to, and which memories do not get mapped. For example, do userrow and usersig map to the same addr space 0x850000?

mcuee:
First cut -- **TO BE UPDATED**

1) EEPROM -- "eeprom"     
2) FLASH -- "flash",        "application",  "apptable",   "boot"        
3)  FUSE -- 
    "lfuse",        "hfuse",        "efuse",    "fuse",         "fuse0",        "wdtcfg",       "fuse1", 
   "bodcfg",       "fuse2",        "osccfg",       "fuse3",    "fuse4",        "tcd0cfg",      "fuse5",        "syscfg0", 
   "fuse6",        "syscfg1",      "fuse7",        "append",    "codesize",     "fuse8",        "fuse9",        "bootend", 
   "bootsize",     "fuses"
     
4) LOCK --  "lock", "lockbits"

5) SIGNATURE -- "signature"

6) USER SIGNATURE -- "usersig",      "userrow", 

7) Un-mapped --   "tempsense",     "calibration",  "osccal16",     "osccal20",     "osc16err",   "osc20err"

8) TBD -- "data"

mcuee:
@stefanrueger @MCUdude @dl8dtl 
I beleive the above is about right. Please review and comment.

stefanrueger:
Thanks for your research, @mcuee. Some memories are not considered, eg, `prodsig`.

The likes of `wdtcfg`, `bodcfg` etc work in `elf_mem_limits()` b/c they are aliases to `fuseN` so `elf_mem_limits()` only ever sees `fuseN`. The (larger) memory `fuses`  **needs to be added to the code,** though, as does `usersig`, `userrow` and `data`. I realise SRAM is volatile, and the utility of loading SRAM is questionable as presumably a reset does not guarantee SRAM data to survive, and I know a power cycle defo won't; still I think treating the AVR memory `data` here in the same way does not hurt (though should add a comment that the section `.data` is different and belongs to flash).

Seems like `tempsense`, `calibration`, `osccal16`, `osccal20`, `osc16err`, `osc20err` are not modelled by avr-gcc, so we are out of luck with these, but there you go.

BTW, a cleaner way of handling the different avr-gcc sections would be to introduce a parameter `section` in `avrdude.conf` memories and assign symbolic values such as `MEM_FLASH`, `MEM_EEPROM` etc. There are numerous occasions in AVRDUDE where the code needs to know is this a flash-type memory or an eeprom-type memory (mainly in the paged r/w routines of the `-c` programmer code but also elsewhere. Would be much cleaner if that could be queries like `mem->section == MEM_EEPROM` rather than the current `!strcmp(mem->desc, "eeprom")`.  Could have a section `MEM_ETC` or `MEM_UNDEFINED` for memories not treated/assigned by avr-gcc.

mcuee:
This PR may need some reviews from @dl8dtl before it can be merged.

stefanrueger:
Just utilised @mcuee's findings on avr-gcc memory sections for the elf files. Extending to "data" may not be practical unless there is a programmer that can actually write to SRAM of the ATxmega parts and this makes sense in practice. It may be that there is currently no such programmer, but what do we know? It is conceivable that a bootloader has a write-to-SRAM function and would benefit from that kind of thing so I included that memory. `fuses`, `usersig` and `userrow` should work, but I don't have the hardware to test (and for the benefit of @mcuee, just to be clear that I *don't* give guarantees that the part won't be bricked when writing to the `fuses` memory that is a concatenation of  all fuses on some parts)

mcuee:
> Just utilised @mcuee's findings on avr-gcc memory sections for the elf files. Extending to "data" may not be practical unless there is a programmer that can actually write to SRAM of the ATxmega parts and this makes sense in practice. It may be that there is currently no such programmer, but what do we know? It is conceivable that a bootloader has a write-to-SRAM function and would benefit from that kind of thing so I included that memory. 

This is interesting. I do not know if any bootloader or programmer which is using this feature. 

> `fuses`, `usersig` and `userrow` should work, but I don't have the hardware to test (and for the benefit of @mcuee, just to be clear, I _don't_ give guarantees that the part won't be bricked when writing to the `fuses` memory that is a concatenation of all fuses on some parts :)

Haha, I will need to be careful about fuses. I think I have bricked a few boards so far due to fuses, at least two Uno clones because of DW fuse. I believe I forgot to unlcok the parts before enable DW fuse bit. It is supposedly recoverable using high-voltage programmer but it seems to be rather troublesome. Since they are cheap enough, I just removed the ATmega328P SMD chips and plan to use them as USB to Serial Converter instead (CH340 adapter on-board).

Then the AVR64DD32 Curiosity Nano board was indirectly killed because I was playing with high voltage UPDI fuse settings with the PICKit 4 (borrowed from local Microchip office). That was a success. But apparently I damaged the board later when playing with high voltage UPDI, probably due to wrong connections between the board and PICKit 4. I will keep the board and replace the chip once I got the SMD AVR64DD32 chip in the future.

Issue #1218, opened 2022-12-14
Avrdude fails to detect PICkit4 and SNAP in PIC mode after a firmware recovery


MCUdude:
It turns out that the SNAP and PICkit4 has a different USB PID when the programmer has just received new firmware after being in recovery mode.

The PICkit4 has PID 0x9017 (instead of the usual 0x9012 in PIC mode), and the SNAP has PID 0x9019 (instead of the usual 0x9018).

I'll submit a PR where the 0x9017 and 0x9019 PIDs are taken into account

```
$ ./avrdude -csnap_isp -patmega16 -v

avrdude: Version 7.0-20220508
         Copyright the AVRDUDE authors;
         see https://github.com/avrdudes/avrdude/blob/main/AUTHORS

         System wide configuration file is /Users/hans/Downloads/avrdude/src/avrdude.conf
         User configuration file is /Users/hans/.avrduderc

         Using Port                    : usb
         Using Programmer              : snap_isp
avrdude usbhid_open() warning: USB device with VID: 0x03eb and PID: 0x2180 not found
avrdude usbhid_open() warning: USB device with VID: 0x03eb and PID: 0x217f not found
avrdude usbhid_open() warning: USB device with VID: 0x03eb and PID: 0x2181 not found
avrdude jtag3_open_common() error: did not find any device matching VID 0x04d8 and PID list: 0x2180, 0x217f, 0x2181
avrdude main() error: unable to open programmer snap_isp on port usb

avrdude done.  Thank you.
```

<img width="451" alt="image" src="https://user-images.githubusercontent.com/3238730/207687356-9946edc4-869f-4a6a-bd15-4cd66c04cae2.png">

MCUdude:
Fixed in #1219

PR #1219, opened 2022-12-14
Add alternative USB PID for PICkit4/SNAP in PIC mode

MCUdude:
See issue #1218.

A SNAP or PICkit has a different USB PID than what Avrdude would expect after they've been though a firmware recovery.
With this PR, Avrdude can detect this and display an appropriate error message.


mcuee:
This looks good to me.

stefanrueger (commented):
Looks perfect

PR #1239, opened 2022-12-26
Temporarily disable brew update in macOS github action

mcuee:
This is to workaround the current problem that macOS github action is broken. 
`brew update` as of now is not needed as the macOS github image has been updated. We can revert this later once the issue is fixed or we encounter new issues.


mcuee:
I will take the liberty to merge this so that other PRs can proceed.

PR #1198, opened 2022-11-26
Tie up loose ends for urclock programmer

stefanrueger:
This PR ties up loose ends for the urclock programmer
 - Add autobaud_sync to avrdude.conf part description
 - Establish avr_*timestamp() library routines for timing
 - Adapt urclock_getsync() to consider legacy bootloaders
 - Update documentation re the urclock programmer
 - Get boot section info from avrdude.conf and not from avrintel.c

In particular, I changed the synchronisation routine to better deal with non-urboot bootloaders. They now work under Linux out of the box, only picoboot needs the new flag `-xstrict`. Urclock and urboot have a fast synchronisation, but other bootloaders *might* need an additional delay of `-xdelay=<n>` ms, depending on the board and the OS (Windows vs Linux). This number <n> can be negative, in which case it reduces the standard delay of 110 ms after board reset.

Under Linux with urclock boards and legacy bootloaders no `-xdelay=...` is needed. The following bash script tests 24 delays from -15 ms to 100 ms.

``` sh
for d in {-15..100..5}; do
  avrdude -curclock -pm328p -xdelay=$d -xshowall >& /dev/null
  echo -n $? | tr 01 .X
done; echo
```
A `.` means the `-xdelay=` value works; an `X` means it fails. Delay 0 (absence of `-xdelay=...`) is the fourth result letter:

| Bootloader | Result | Result using `-xstrict` | Comments |
| --: | --: | --: | :-- |
urboot fixed baud rate|`........................`|`........................`| All good |
urboot autobaud |`........................`|`........................`| All good |
optiboot v8.3 |`X......XXXXXXX.....XXXXX`|`...XXXX........XXXX.....`| Only some `-xdelay=` work |
optiboot v4.4 |`........................`|`XXXXXXXXXXXXXXXXXXXXXXXX`| Must not use `-xstrict` |
picoboot |`XXXXXXXXXXXXXXXXXXXXXXXX`|`........................`| Must use `-xstrict` |

urclock-urboot-test

I also tested synchronisation times with `/bin/time -f%e avrdude -qqpm328p -xshowversion [-xstrict]`:
| Bootloader | Version | Synchronisation time | ... using `-xstrict` |
| --: | :-: | --: | :-: |
|urboot fixed baud rate|`u7.6 w-u-jpr-.`|0.23 s|0.22 s|
|urboot autobaud|`u7.7 weu-jPrac`|0.27 s|0.26 s|
[...]

PR #1246, opened 2022-12-28
Modify recv timeouts to accommodate urclock sync with ftdi

stefanrueger:
Addresses Issue #1245

Turns out the current receive timeout for synchronising is too short to work with FTDI USB/serial converters. Changed from 10 ms to 25 ms. The secondary change is a calibrated wait so that optiboot v8.x bootloaders work out of the box with `-c urclock`, see PR #1198


mcuee:
This PR is good to go. Tested under Windows (no regression) and Linux (fixed the issue).
* https://github.com/avrdudes/avrdude/issues/1245#issuecomment-1366999795

PR #1247, opened 2022-12-29
Silence conf path detection debug output

mariusgreuel:


mcuee:
I think this should be safe to merge.

PR #1248, opened 2022-12-29
Make WIN32 implementation of ser_recv() fail on time-out

mariusgreuel:
When a time-out occurs during a `ser_recv`, `ReadFile` may return zero or more bytes. As anything less than `buflen` bytes (but zero) cannot be detected by the caller, this PR fails a partial read with -1.

### DIFF
--- a/src/ser_win32.c
+++ b/src/ser_win32.c
@@ -486,12 +486,12 @@ reselect:
 		nfds = select(fd->ifd + 1, &rfds, NULL, NULL, &to2);
 		if (nfds == 0) {
 			if (verbose > 1) {
-				pmsg_notice("ser_recv(): programmer is not responding\n");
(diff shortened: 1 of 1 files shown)


mcuee:
This should be safe to merge.

stefanrueger:
Good catch. Looks good. While you are at it you might as well change the error messages in [Line 489](https://github.com/avrdudes/avrdude/blob/0f956e90eb66e0ba7a4fecdd59e6fab31f346963/src/ser_win32.c#L489) and [Line 494](https://github.com/avrdudes/avrdude/blob/0f956e90eb66e0ba7a4fecdd59e6fab31f346963/src/ser_win32.c#L494) to say `net_recv()` instead of `ser_recv()`.

mariusgreuel:
> While you are at it you might as well change the error messages

Done.

PR #1251, opened 2022-12-30
Guard urclock_getsync() against initial spurious input

stefanrueger:
Fixes Issue #1245

### DIFF
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -1270,7 +1270,7 @@ frame is programmed.
 .It Ar delay=<n>
 Add a <n> ms delay after reset. This can be useful if a board takes a
 particularly long time to exit from external reset. <n> can be negative,
-in which case the default 80 ms delay after issuing reset will be
+in which case the default 120 ms delay after issuing reset will be
 shortened accordingly.
 .It Ar strict
 Urclock has a faster, but slightly different strategy than -c arduino to
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -1270,7 +1270,7 @@ frame is programmed.
 .It Ar delay=<n>
 Add a <n> ms delay after reset. This can be useful if a board takes a
 particularly long time to exit from external reset. <n> can be negative,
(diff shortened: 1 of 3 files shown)


stefanrueger:
Under Linux with urclock boards and legacy bootloaders no `-xdelay=...` is needed. The following bash script tests 20 delays from -15 ms to 80 ms.

``` console
for d in {-15..80..5}; do
  avrdude -curclock -pm328p -xdelay=$d -xshowall >& /dev/null
  echo -n $? | tr 01 .X
done; echo
```
A `.` means the `-xdelay=` value works; an `X` means it fails. Delay 0 (absence of `-xdelay=...`) is the fourth result letter:

| Bootloader | Result | Result with `-xstrict` | Comments |
| --: | --: | --: | :-- |
| `u7.5 we--h-r-.` | `....................` | `....................` | All good |
| `u7.6 w-u-jpr-.` | `....................` | `....................` | All good |
| `u7.7 weu-jPr-c` | `....................` | `....................` | All good |
| `u7.7 weu-jPrac` | `X...................` | `....................` | `-xdelay=n` OK for n >= -10 |
| `o8.3 --s-h-r--` | `....................` | `X............X......` | Some `-xdelay` do not work |
| `o4.4 --s-h-r--` | `....................` | `XXXXXXXXXXXXXXXXXXXX` | opti v4.4 must *not* use `-xstrict` |
| `x0.0 .........` | `XXXXXXXXXXXXXXXXXXXX` | `....................` | picoboot must use `-xstrict` |

stefanrueger:
Decided to lengthen delay after issuing reset for benefit of autobaud bootloaders: it really is the first byte from avrdude that makes autobaud bootloaders set their USART divisor, so we better make sure the board has had enough time to come out of reset. @mcuee had a Nano clone board that needed a bit longer (under Windows) so better to increase the margin. 

Also changed from warning to notice2 the info on libreadline.

So, am going to merge so we are ready for a New Year release :)

PR #1255, opened 2023-01-01
Add a "q" command to the terminal

dl8dtl:
"q" used to be a unique abbrevation for "quit" before the introduction of the "quell" command, and it was an easy way to exit terminal mode.

By adding it as an explicit command (must be before all longer commands starting with "q"), this functionality is regained.

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -85,6 +85,7 @@ static int cmd_quell  (PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]);
 #define _fo(x) offsetof(PROGRAMMER, x)
 
 struct command cmd[] = {
+  { "q",     cmd_quit,  _fo(open),              "abbreviation for 'quit'" },
   { "dump",  cmd_dump,  _fo(read_byte_cached),  "%s <memory> [<addr> <len> | <addr> ... | <addr> | ...]" },
   { "read",  cmd_dump,  _fo(read_byte_cached),  "alias for dump" },
   { "write", cmd_write, _fo(write_byte_cached), "write <memory> <addr> <data>[,] {<data>[,]}" },
--- a/src/term.c
+++ b/src/term.c
@@ -85,7 +85,6 @@ static int cmd_quell  (PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]);
 #define _fo(x) offsetof(PROGRAMMER, x)
 
 struct command cmd[] = {
-  { "q",     cmd_quit,  _fo(open),              "abbreviation for 'quit'" },
   { "dump",  cmd_dump,  _fo(read_byte_cached),  "%s <memory> [<addr> <len> | <addr> ... | <addr> | ...]" },
   { "read",  cmd_dump,  _fo(read_byte_cached),  "alias for dump" },
   { "write", cmd_write, _fo(write_byte_cached), "write <memory> <addr> <data>[,] {<data>[,]}" },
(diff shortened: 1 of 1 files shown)


stefanrueger (commented):
Neat idea and has precedence with ? as abbreviation for help.

I would change the order, so that the help message looks neater:
```
...
  { "?",     cmd_help,  _fo(open),              "same as help" },
  { "help",  cmd_help,  _fo(open),              "show help message" },
  { "q",     cmd_quit,  _fo(open),              "abbreviation for quit" },
  { "quit",  cmd_quit,  _fo(open),              "quit after writing out cache for flash & EEPROM" }
```
(move "?" one up and move "q" just before "quit")
That way the two abbreviations appear just before the command they abbreviate.

Also, I would remove the single quotes around quit in the help text. These days quotes are rarely used, even when referring to titles etc in text (eg, Yesterday I saw 2001: A Space Odyssey in the cinema)

stefanrueger:
> must be before all longer commands starting with "q"

No, works if it appears in between: any command that completely matches with what the user typed will be accepted straight away.

dl8dtl:
> I would change the order, so that the help message looks neater:

Basically agreed, but it must be *before* the "quell" command.

dl8dtl:
I tried appending it to the end of the table, that didn't work.

stefanrueger:
It works in the suggested order, though (between quell and quit), see https://github.com/avrdudes/avrdude/blob/b6d50ef0a9c69e288cbd64482842322881f1335c/src/term.c#L1246-L1247

stefanrueger:
The following pre-existing code is premature though, which complains about ambiguous commands as soon as sth looks ambiguous without allowing for an exact match further below. But q between quell and quit still works

https://github.com/avrdudes/avrdude/blob/b6d50ef0a9c69e288cbd64482842322881f1335c/src/term.c#L1249-L1252

Would be better to count the number of matches and complain about ambiguity *after* the loop, but hey, ho, for what we need now it's OK to leave the code as is.

dl8dtl:
> It works in the suggested order, though (between quell and quit)

Hmm. That's a bit confusing …
I agree that it makes more sense to defer the ambiguity test.

Have to think about it a moment, rather proceeding with the 7.1 release tests now.

stefanrueger:
| Have to think about it a moment, rather proceeding with the 7.1 release tests now.

Let me review the partial matching in AVRDUDE and create a PR. Better use of your time if you test :)

stefanrueger:
With PR #1256 it should be possible to add the q command as the last one in the list (also fixed subtle other errors on locate_mem() functions).

mcuee:
@stefanrueger 

In this case PR #1256 will replace this PR #1255, right?

mcuee:
@stefanrueger and @dl8dtl 

Indeed PR #1256 is good and it makes sense to add `q` after `quit` in the end of the list to be consisent with `?` after `help`.

```
PS C:\work\avr\avrdude_test\avrdude_sr> git diff
diff --git a/src/term.c b/src/term.c
index 63a9464..5f34dea 100644
--- a/src/term.c
+++ b/src/term.c
@@ -108,6 +108,7 @@ struct command cmd[] = {
   { "help",  cmd_help,  _fo(open),              "show help message" },
   { "?",     cmd_help,  _fo(open),              "same as help" },
   { "quit",  cmd_quit,  _fo(open),              "quit after writing out cache for flash & EEPROM" }
+  { "quit",  cmd_quit,  _fo(open),              "Same as quit" }
 };

 #define NCMDS ((int)(sizeof(cmd)/sizeof(struct command)))

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1256_pr1255 -c stk500v2 -P COM6 -p m328p -qqt
avrdude> q
PS C:\work\avr\avrdude_test\avrdude_bin>
```

PR #1260, opened 2023-01-03
Clean up avrdude.conf.in fixing a couple of bugs

stefanrueger:
Whilst trying to clean up avrdude.conf.in for v7.1 I came across a few bugs. Corrected here:

 - Set pin number or pin list rather than adding them
   <br>Pin definitions in config_gram.y fail to clear existing pin definitions when setting them for the second time, eg, through inheritance, For example,
   ```
   programmer parent "ft2232h"
       id                     = "tigard";
       reset                  = 5; # BD5 (GPIOL1)
   ;
   ```
    would add pin 5 to the existing pin 3 for reset from the parent programmer rather than overwriting the reset pin. @mariusgreuel, can you confirm that 5 is the *only* pin that should see the reset signal for the `tigard` programmer?
    This commit clears any pre-existing pin definition first before assigning the new pin number or pin list (for vcc or buff).

 - Change avrdude.conf.in into canonical form
   <br>`avrdude -p*/s -c*/s` outputs how avrdude understands the programmers and parts defined in avrdude.conf.in that syntax (the canonical form). This commit ensures that avrdude.conf.in is in canonical form, typically removing assignments that are redundant owing to inheritance.
 - Also changes a couple of comments and one programmer description string to be more expressive.
 - Fix boot section size of AT90PWM2 and related parts
 - Update avrintel.[ch] fixing boot section sizes
    
 - Fix duplicate `lockbits` memory output of `avrdude -p*/s`
   <br>In order to output memories of a part in the same order, the developer options use an ordered list of all known memories. Owing to the partial match feature of dev_locate_mem_noalias() they would therefore print an existing lockbits memory twice, once for the memory `lock` and once for the memory `lockbits` in the list of all known memory names of all parts.

 - Stop printing redundant `usbpid` and `hvupdi_support` lists for -c*/s
   <br>`avrdude -c*/s` would print the `usbpid` and `hvupdi_support` lists unconditionally even though a programmer might have inherited them and therefore not need to set them if they coincide with the parent's. This PR removes that redundancy from the output of `-c*/s`

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -403,6 +403,7 @@ default_spi        = "@DEFAULT_SPI_PORT@";
 # default_bitclock = 2.5;
 
 @HAVE_PARPORT_BEGIN@
+
 # Parallel port programmers
 
 #------------------------------------------------------------
@@ -845,7 +846,7 @@ programmer parent "ft2232h"
 
 programmer parent "ft2232h"
     id                     = "2232hio";
-    desc                   = "FT2232H based generic programmer";
+    desc                   = "2232hio based on FT2232H with buffer and LEDs";
     buff                   = ~4;
 # LED SIGNALs
     errled                 = ~11;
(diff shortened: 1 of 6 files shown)


stefanrueger:
Without the PR we have
```
$ avrdude -c tigard/s
# tigard

# Tigard - FT2232H based multi-protocol tool for hardware hacking.
# https://github.com/tigard-tools/tigard

programmer parent "ft2232h"
    id                     = "tigard";
    desc                   = "Tigard interface board";
    usbpid                 = 0x6010;
    usbdev                 = "B";
# ISP-signals
    reset                  = 3, 5; # BD5 (GPIOL1)
;
```
Reset has been assigned to two pins, 3 (the pin from the parent programmer) and 5 (the pin requested in avrdude.conf)

With this PR:
```
$ avrdude -c tigard/s
# tigard

# Tigard - FT2232H based multi-protocol tool for hardware hacking
# https://github.com/tigard-tools/tigard

programmer parent "ft2232h"
    id                     = "tigard";
    desc                   = "Tigard interface board";
    usbdev                 = "B";
# Reset is different to the FT2232H; sck, sdo, sdi remain at 0, 1, 2
    reset                  = 5; # BD5 (GPIOL1)
;
```
Reset is assigned correctly.

PR #1266, opened 2023-01-04
Silence page erase in cache code

stefanrueger:
The cache code that is used in the terminal utilises the page erase function if the programmer has one. Some programmers don't follow through, though. The cache code recognises that and still does "the right thing". However, one can get false-alarm error messages such as 
```
avrdude> q
avrdude: synching cache to device ... avrdude jtag3_page_erase() error: page erase not supported

Reading | ################################################## | 100% 0.89 s
Erasing | ################################################## | 100% 0.01 s
Writing | ################################################## | 100% 0.07 s

avrdude done.  Thank you.
```

This PR silences the page erase at the cache level.

### DIFF
--- a/src/avrcache.c
+++ b/src/avrcache.c
@@ -331,6 +331,17 @@ static int guessBootStart(const PROGRAMMER *pgm, const AVRPART *p) {
 }
 
 
+// Page erase but without error messages if it does not work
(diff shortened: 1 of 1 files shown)


mcuee:
@stefanrueger 

I created Issue #1267 for the future work to smoke out all the programmers that advertise their page_erase() but then fail to provide it.

Issue #957, opened 2022-05-10
auto-tools build: missing ws2_32 library under Windows MSYS2 MinGW


mcuee:
Build log under MSYS2 mingw64

```
MINGW64 /c/work/avr/avrdude-7.0
$ make V=1
make  all-recursive
make[1]: Entering directory '/c/work/avr/avrdude-7.0'
Making all in .
make[2]: Entering directory '/c/work/avr/avrdude-7.0'
/bin/sh ./libtool  --tag=CC   --mode=link gcc -Wall -Wno-pointer-sign -g -O2 -DWIN32NATIVE  -static -o avrdude.exe avrdude-main.o avrdude-whereami.o avrdude-term.o ./libavrdude.a -lusb-1.0  -lusb -lftdi1  -lhid -lsetupapi -lelf -lpthread -lm -lhidapi -lreadline -lncurses -ltermcap  -lftdi1
libtool: link: gcc -Wall -Wno-pointer-sign -g -O2 -DWIN32NATIVE -o avrdude.exe avrdude-main.o avrdude-whereami.o avrdude-term.o  ./libavrdude.a -lusb-1.0 -lusb -lhid -lsetupapi -lelf -lpthread -lhidapi -lreadline -lncurses -ltermcap -lftdi1
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: ./libavrdude.a(libavrdude_a-ser_win32.o): in function `net_drain':
C:\work\avr\avrdude-7.0/ser_win32.c:660: undefined reference to `__imp_select'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\work\avr\avrdude-7.0/ser_win32.c:676: undefined reference to `__imp_WSAGetLastError'
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: C:\work\avr\avrdude-7.0/ser_win32.c:696: undefined reference to `__imp_recv'
[...]

PR #1279, opened 2023-01-12
Update configure.ac for MinGW build

mcuee:
Fix avrdudes#957.
Even though we are only supporting CMake build now, the fix is simple enough and not affecting CMake.


mcuee:
Tested here.
* https://github.com/avrdudes/avrdude/issues/957#issuecomment-1379880911

mcuee:
I would not update NEWS as this auto-tools build is no longer supported.

Issue #1298, opened 2023-02-01
Linux build instructions fail on Ubuntu 18.04


jclxx:
Summary
--------------

**The out-of-the-box build instructions for Linux fail on Ubuntu 18.04 LTS (server), because of cmake errors.  It succeeds in 20.04.**

Although 18.04 is approaching end-of-maintenance it remains under security maintenance until 2028.

It may well be there is an easy workaround for those familiar with cmake.

This was discovered in real life as the lab computer (next to the programming equipment) is 18.04 with avrdude 6.3; the actual immediate problem is to program an Atmega 328P**B** which appears in the later avrdude.conf.

How to reproduce
---

The instructions for building under Linux [here](https://github.com/avrdudes/avrdude/wiki/Building-AVRDUDE-for-Linux) say that to build (Linux/Debian variety), you do:

```
sudo apt-get install build-essential git cmake flex bison libelf-dev libusb-dev libhidapi-dev libftdi1-dev libreadline-dev
git clone https://github.com/avrdudes/avrdude.git
cd avrdude
./build.sh
```

Method

* Create fresh AWS Lightsail instances (smallest kind), with Ubuntu 20.04 LTS or 18.04 LTS. 
* After first boot: apt-get update, apt-get upgrade (accept all default questions), reboot.
* Then the instructions above (apt-get deps, git clone, cd avrdude, ./build.sh)
* Git pull got  commit dd7b359d6b5dfcf02374fc697c80dd6d54cd7a81 of 2023-01-31

Reproducing failure in 18.04
---

This fails early in the build script

    18.04.1-LTS$ ./build.sh 
    CMake Error: The source directory "/home/ubuntu/avrdude/build_linux" does not exist.
    Specify --help for usage, or press the help button on the CMake GUI.
    CMake failed.

I am unfamiliar with cmake.  My guess is that there's some flag needed the first time to populate `build_linux`, but this should be in the build.sh script.  (Or I suppose added to the build instructions.)

Versions

```
18.04.1-LTS$ uname -a
Linux ip-172-26-6-124 4.15.0-1021-aws #21-Ubuntu SMP Tue Aug 28 10:23:07 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
18.04.1-LTS$ cmake --version
cmake version 3.10.2
```

Reproducing success in 20.04
---

```
20.04-LTS$ ./build.sh 
-- The C compiler identification is GNU 9.4.0
...
-- Configuration summary:
[...]

PR #1300, opened 2023-02-02
Fix for #1298 (CMake version not checked properly)

jclxx:
Older versions of CMake will exit with error because of the flags used, instead of giving version-too-small error from CMake.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -74,6 +74,14 @@ case "${ostype}" in
 	;;
 esac
 
+# force cmake version check
+echo "Checking CMake version"
+T=/tmp/versioncheck.$$
+mkdir $T
+fgrep cmake_minimum_required CMakeLists.txt > $T/CMakeLists.txt
+(cd $T && cmake -Wno-dev .) > /dev/null || exit 1
+rm -Rf $T
+
 cmake ${build_flags} ${extra_enable} -D CMAKE_BUILD_TYPE=${build_type} -B build_${ostype} ||\
     { echo "CMake failed."; exit 1; }
 cmake --build build_${ostype} ||\
--- a/build.sh
+++ b/build.sh
@@ -74,12 +74,13 @@ case "${ostype}" in
 	;;
 esac
 
-# force cmake version check
+# force cmake version check before anything else
 echo "Checking CMake version"
 T=/tmp/versioncheck.$$
 mkdir $T
-fgrep cmake_minimum_required CMakeLists.txt > $T/CMakeLists.txt
-(cd $T && cmake -Wno-dev .) > /dev/null || exit 1
+grep -i "^cmake_minimum_required" CMakeLists.txt > $T/CMakeLists.txt
+echo "project(test_cmake_version)" >> $T/CMakeLists.txt
+(cd $T && cmake .) || exit 1
 rm -Rf $T
 
 cmake ${build_flags} ${extra_enable} -D CMAKE_BUILD_TYPE=${build_type} -B build_${ostype} ||\


jclxx:
Fixes #1298

jclxx:
Obviously the main thing wrong with this is that it's in the shell script not CMake, which means it will do nothing for the Windows and Macintosh builds.  I haven't got any suitable platforms for testing on those.

mcuee:
@jclxx 

`build.sh` should work under Linux, macOS, BSDs and other non-Windows OS by default. It also works under Windows with MSYS2.

mcuee:
The change is also fine under Windows with MSYS2 mingw64.
```
MINGW64 /c/work/avr/avrdude_test/avrdude_main
$ ./build_pr1300.sh
Checking CMake version
-- Building for: Ninja
-- The C compiler identification is GNU 12.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: C:/msys64/usr/bin/git.exe (found version "2.39.1")
-- Found FLEX: C:/msys64/usr/bin/flex.exe (found version "2.6.4")
-- Found BISON: C:/msys64/usr/bin/bison.exe (found version "3.8.2")
-- Looking for libelf.h
-- Looking for libelf.h - found
-- Looking for libelf/libelf.h
-- Looking for libelf/libelf.h - found
-- Looking for usb.h
-- Looking for usb.h - not found
-- Looking for lusb0_usb.h
-- Looking for lusb0_usb.h - found
-- Looking for libusb.h
-- Looking for libusb.h - not found
-- Looking for libusb-1.0/libusb.h
-- Looking for libusb-1.0/libusb.h - found
-- Looking for hidapi/hidapi.h
-- Looking for hidapi/hidapi.h - found
-- Looking for ftdi_tcioflush
-- Looking for ftdi_tcioflush - found
-- Configuration summary:
-- ----------------------
-- DO HAVE    libelf
-- DO HAVE    libusb
-- DO HAVE    libusb_1_0
-- DO HAVE    libhidapi
-- DON'T HAVE libftdi
-- DO HAVE    libftdi1
-- DO HAVE    libreadline
-- DISABLED   doc
-- DISABLED   parport
-- DISABLED   linuxgpio
-- DISABLED   linuxspi
-- ----------------------
-- Configuring done
-- Generating done
-- Build files have been written to: C:/work/avr/avrdude_test/avrdude_main/build_mingw64_nt-10.0-22621
...

```

mcuee:
@mariusgreuel and @dl8dtl 

Any comments on this PR?

jclxx:
Just a thought on the tags: I'd have said this is a "bug" fix, rather than an "enhancement".  The build code code does try to check the version of cmake, but it's implemented badly and gives incorrect results for some older versions of CMake, which includes the one shipped with a major OS (Ubu 18.04).  It's not like it adds something ("let's check the version"), it fixes something ("let's make the version check actually work as intended").

jclxx:
FYI: I took a look at the source of CMake to see if there was a better way to do this.  The fundamental issue is that the usage of `-B` appears to have subtly changed (whether it makes the dir or not).  If we use a flag wrongly, it will fail before checking the version requirements in `CMakeLists.txt`.  This issue isn't really about an old version of CMake.  It's about making sure AVRDUDE's build process is robust when it takes advantage of later options of the build tools; which may well happen again in the future.

There is no reliable way out of this, as CMake doesn't have a "please-just-check-version" flag. So we have to fabricate one per my pull request.

---
Saves anybody else looking: the CMake version which is failing (3.10.2) is called with `-B build_linux`, and although `-B` exists in the source, it is not covered in the man page until 3.13.  CMake's git log says "Document the previously internal option of '-B' and provide a matching source directory option with '-S'."  Commit [a10d63d5](https://gitlab.kitware.com/cmake/cmake/-/commit/a10d63d5783275f9972aaa86d41071a1ddca7921)

mariusgreuel:
> The fundamental issue is that the usage of `-B`

Yes. Instead of doing a version check in `build.sh`, I suggest to not use `-B` and call older versions of CMake in a compatible way, so the version check can be done by CMake.

mcuee:
> > The fundamental issue is that the usage of `-B`
> 
> Yes. Instead of doing a version check in `build.sh`, I suggest to not use `-B` and call older versions of CMake in a compatible way, so the version check can be done by CMake.

@mariusgreuel 

In this case, we will not adopt this PR, right?

jclxx:
@mariusgreuel @mcuee 

I have a fix which doesn't use `-B`, which is quite simple but in my opinion less reliable that what I've posed here.  Certainly I'll post it here for you to consider, do you prefer a separate pull request? Or should I describe here in comments for discussion.

mcuee:
> I have a fix which doesn't use `-B`, I'll put it here later today after I've tested it a little further.
> 
> It's quite simple but in my opinion it's not as good as what I've posted, but let's see you what you think.

You can describe what you have in mind here. Then let's see if @mariusgreuel likes it or not. He is the CMake expert here.

mariusgreuel:
A simple 'fix' would be using CMakes old out-of-source syntax:

```bash
mkdir -p build
cd build
cmake ..
cmake --build .
```

however, all this gains you is a proper error message, as CMake will stop right at the version check.

I guess I am OK with changing `build.sh`, but I am ambivalent about this PR in general. Supporting older tool-chains, libraries, and OSes costs time, and I am not sure if its worth it. Should we support Ubuntu 18.04 LTS? Then we would need to rework the CMake files a bit. Should we document what the requirements are? Even that is quite a bit of work, and it changes frequently.

@dl8dtl What is your opinion on supporting older OSes?

mcuee:
> Supporting older tool-chains, libraries, and OSes costs time, and I am not sure if its worth it.

I agree with you here. I think we do not need to merge this PR or change any codes. We can document that CMake 3.14 or later version is required in the Wiki.

jclxx:
> however, all this gains you is a proper error message, as CMake will stop right at the version check.

**This is not small**: it took me two days to find out that "all" I needed was a later version of CMake.  Instead of 1 second.

The current PR isn't intended to make AVRDUDE compile on older OS: it's intended to give a proper error message for a condition that isn't intended to be supported, so the user can figure out what to do.  It just adds an invocation of CMake which will check the version number required without doing anything else.

My workaround without `-B` is the same as @mariusgreuel , but it's silly to avoid using modern features of CMake (and difficult to, difficult to test etc).

Nonetheless, thanks for updating the Wiki, that's really helpful.  I regard #1298 as resolved by  wiki commit 8f9b826cff6c02691a0b778a255f3f3ce7a59965

mcuee:
> I didn't mean to close this, I clicked 'close' by mistake and don't see how to undo that.
>
> I really think making the version message appear is worth it.

I have re-opened the issue. Let's hear what @dl8dtl says about this PR.

jclxx:
@mcuee many thanks for reopening.  If we close it, let's close it on purpose!

Just to summarise my view: the current situation is that the build process fails with a very misleading error message under some pretty rare conditions.  A fix is presented which gives good error messages, and will reliably give good error messages in the future.  CMake which is forced to check its version against the requirement in the CMakeLists.txt, instead of forcing the user to do it manually by reading the Wiki.  Why would we make the human do it when we have code for the build process to do it?

A live user in the field had this is a real problem, wasted two days finding how to solved it, and produced the PR.   Please consider adopting the PR.

mariusgreuel:
> it took me two days to find out that "all" I needed was a later version of CMake

Sorry to head that. I guess what is confusing is that the old version of CMake does not complain about the unknown `-B` option and does something unexpected. Apparently, `-B` used to be a hidden option, and the semantics were changed once `-B` was officially introduced.

@jclxx So, lets use the old CMake build syntax in `build.sh`. I can submit the patch.

mariusgreuel:
@jclxx Can you try https://github.com/avrdudes/avrdude/pull/1301 with 18.04, please?

jclxx:
@mariusgreuel Thanks I will test #1301 shortly.

jclxx:
Thanks all for continued attention on this.

@mariusgreuel as anticipated, #1301 behaves exactly correctly on Ubuntu 18.04 with cmake from standard Ubuntu repository, and generates a bad version error.  I put my test results there.  It certainly solves the current issue in giving a correct error message on older systems, and builds correctly on newers ones (I believe you tested this).

I repeat though, that requiring "antique" invocation of CMake is bound to break in the future (especially with no comment saying why we're doing it).  The method here #1298 is a general purpose method to make CMake check its version is suitable before asking it to do anything else, and as such is robust.  Later AVRDUDE developers can use whatever CMake offers, knowing that the version is checked properly even on old systems.  Otherwise, at a certain point, we have to say just don't care how old systems fail, because no one will do the testing. It really seems like a hostage to fortune.

Can I ask why you would prefer #1301 over #1298?

@dl8dtl @mcuee do you have any opinion?

mcuee:
@jclxx 

To be honest, I am thinking that we do not need either PR with the Wiki updates. But I am okay with either #1300 or #1301 in the end,

stefanrueger:
Closing as discussed

PR #1309, opened 2023-03-05
Change Arduino auto-reset via RTS to allow direct RTS-reset connection

### LINKED ISSUE #1262: Change Arduino auto-reset via DTR/RTS/CTS to allow direct reset connection
Currently, the Arduino and urclock programmers attempt to auto-reset the board via the RTS line, to kick it into bootloader mode. On open, the do a 'RTS=0, Sleep, RTS=1', which pulls the reset line low. On close they do a 'RTS=0', which pulls the reset line high again. In other words, during the whole operation the RTS line stays at low-level.

To avoid a continuous reset, this behavior requires a special circuit that turns the low-level signal into a low pulse. For instance, the Arduino Uno has a capacitor in between the RTS line and the reset pin.

For testing with boards that do not have the integrated RTS reset circuitry, I would like to be able to simply connect the chips reset pin to the RTS pin. However, this currently requires a similar circuit, i.e. at least a cap and a diode to clamp the voltage after the cap. I find this inconvenient, unreliable, plus I am in constant fear of damaging the chips by not connecting the diode correctly.

Side story: I had some unreliable reset behavior, so I hooked up a scope to the reset pin just to find that my 100n cap was not big enough to reliably pull down the chips reset pin. Then I saw the voltage jump up close to 8.5V, and found that my diode misplaced...

How about changing the code in `open` to

```c
serial_set_dtr_rts(&pgm->fd, 0); // Discharge the capacitor, if present
usleep(20*1000);
serial_set_dtr_rts(&pgm->fd, 1); // Pull the reset line low
usleep(20*1000);
serial_set_dtr_rts(&pgm->fd, 0); // Pull the reset line high
usleep(20*1000);
```

I think this would work for both the 'rts-capacitor-reset' boards, as well for a direct connection.

Any thoughts?

mariusgreuel:
Closes #1262

@stefanrueger Here is something to look at. I have not grasped the timing logic in urclock, but assuming the magic starts with releasing the reset line in [line 2247](https://github.com/mariusgreuel/avrdude-fork/blob/pr-auto-reset/src/urclock.c#L2247), I figured 120 minus 50 for the extra wait?

### DIFF
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -88,7 +88,13 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
   // This code assumes a negative-logic USB to TTL serial adapter
   // Set RTS/DTR high to discharge the series-capacitor, if present
   serial_set_dtr_rts(&pgm->fd, 0);
-  usleep(50 * 1000);
+  /*
+   * Long wait needed for optiboot: otherwise the second of two bootloader
+   * calls in quick succession fails:
+   *
+   * avrdude -c arduino -qqp m328p -U x.hex; avrdude -c arduino -qqp m328p -U x.hex
+   */
+  usleep(250 * 1000);
   // Pull the RTS/DTR line low to reset AVR
   serial_set_dtr_rts(&pgm->fd, 1);
   usleep(50 * 1000);
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -85,13 +85,16 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
     return -1;
   }
 
-  /* Clear DTR and RTS to unload the RESET capacitor 
-   * (for example in Arduino) */
+  // This code assumes a negative-logic USB to TTL serial adapter
+  // Set RTS/DTR high to discharge the series-capacitor, if present
   serial_set_dtr_rts(&pgm->fd, 0);
-  usleep(250*1000);
-  /* Set DTR and RTS back to high */
+  usleep(50 * 1000);
+  // Pull the RTS/DTR line low to reset AVR
   serial_set_dtr_rts(&pgm->fd, 1);
-  usleep(50*1000);
+  usleep(50 * 1000);
+  // Set the RTS/DTR line back to high
+  serial_set_dtr_rts(&pgm->fd, 0);
+  usleep(50 * 1000);
 
   /*
    * drain any extraneous input
@@ -106,7 +109,6 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
 
 static void arduino_close(PROGRAMMER * pgm)
 {
-  serial_set_dtr_rts(&pgm->fd, 0);
   serial_close(&pgm->fd);
   pgm->fd.ifd = -1;
 }
--- a/src/stk500.c
+++ b/src/stk500.c
@@ -98,15 +98,11 @@ int stk500_getsync(const PROGRAMMER *pgm) {
     // Restart Arduino bootloader for every sync attempt
     if (strcmp(pgm->type, "Arduino") == 0 && attempt > 0) {
       // This code assumes a negative-logic USB to TTL serial adapter
-      // Set RTS/DTR high to discharge the series-capacitor, if present
-      serial_set_dtr_rts(&pgm->fd, 0);
-      usleep(50 * 1000);
-      // Pull the RTS/DTR line low to reset AVR
+      // Pull the RTS/DTR line low to reset AVR: it is still high from open()/last attempt
       serial_set_dtr_rts(&pgm->fd, 1);
-      usleep(50 * 1000);
+      usleep(20*1000);
       // Set the RTS/DTR line back to high
       serial_set_dtr_rts(&pgm->fd, 0);
-      usleep(50 * 1000);
       stk500_drain(pgm, 0);
     }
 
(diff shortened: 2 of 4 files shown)


stefanrueger:
>assuming the magic starts with releasing the reset line in [line 2247](https://github.com/mariusgreuel/avrdude-fork/blob/pr-auto-reset/src/urclock.c#L2247)

Only when directly connected!! When connected via a cap, the magic starts less than a ms after [Line 2245](https://github.com/mariusgreuel/avrdude-fork/blob/b1da0b09222d107f569238db387a51cc72b9f8ef/src/urclock.c#L2245). See

reset

Typical C is 100 nF, typical R is 10k, so we are talking t = RC ~ 1 ms. Really no need to spend 50 ms each!

Why the sleep in [Line 2249]? Time will march on regardless. Can I suggest
```
  // This code assumes a negative-logic USB to TTL serial adapter
  // Set RTS/DTR high to discharge the series-capacitor, if present
  serial_set_dtr_rts(&pgm->fd, 0);
  usleep(20 * 1000);
  // Pull the RTS/DTR line low to reset AVR
  serial_set_dtr_rts(&pgm->fd, 1);
  usleep(20 * 1000);
  // Set the RTS/DTR line back to high
  serial_set_dtr_rts(&pgm->fd, 0);

  if((100+ur.delay) > 0)
    usleep((100+ur.delay)*1000); // Wait until board comes out of reset
```

Similar in stk500.c

If you insist on changing the original 250 ms wait in arduino.c in [Line 91](https://github.com/avrdudes/avrdude/blob/b1da0b09222d107f569238db387a51cc72b9f8ef/src/arduino.c#L91) to 50 ms, I predict there will be trouble when executing two uploads to **optiboot** in quick succession
```
$ avrdude -c arduino -qqp m328p -U blink.hex; avrdude -c arduino -qqp m328p -U blink.hex;
```

Leave that 250 ms be. (Urclock deals with this by sleeping at exit for 200 ms if it had served optiboot).

See my comment [on this](https://github.com/avrdudes/avrdude/issues/1262#issuecomment-1454920629)

stefanrueger:
> Similar in stk500.c

Actually, we know DTR/RTS has been set 0 (ie, high voltage) earlier, all we need to do is
``` C
  // Pull the RTS/DTR line low to reset AVR
  serial_set_dtr_rts(&pgm->fd, 1);
  usleep(20*1000);
  // Set the RTS/DTR line back to high
  serial_set_dtr_rts(&pgm->fd, 0);
```
The sleep after that won't be necessary as draining input itself sleeps and getsync spends time ...

And should we not also improve the other two serial programmers that reset the MCU board by plugging DTR/RTS?

```
wiring.c
xbee.c
```

mcuee:
> And should we not also improve the other two serial programmers that reset the MCU board by plugging DTR/RTS?
> 
> ```
> wiring.c
> xbee.c
> ```

I will say `wiring.c` (stk500v2 based bootloader) is good to have the change as well. 

I am not so sure whether it is a good idea to change `xbee.c` or not. It seems that there is no such thing as `direct connect to reset pin` for the `xbee` programmer.
https://www.arduino.cc/en/uploads/Main/XbeeShieldSchematic.pdf

`serial_set_dtr_rts` is also used by `updi_link.c` but I guess the usage there is different.

stefanrueger:
Last weekend I played with the Falstad simulator. It corroborates my sketch above: below graph pulls DTR low for 20 ms
 on a reset circuit.  After the first millisecond Reset is high enough to be deactivated. Crucially, it is low for at least ~ 2 µs that are required for reset to become active. The simulation shows that 20 ms is sufficient by a country mile.

reset-simulation-label

(The circuit misses external clamp diode protection to Vcc. The chip doesn't have internal protection on the reset pin to allow High Voltage programming. The graph above neatly shows the spike to 2*Vcc when switching the DTR signal to Vcc again. With Vcc > 5V one needs the diode to prevent accidentally entering the HV mode.)

stefanrueger:
Anyway, I pushed my suggested changes onto your PR,  @mariusgreuel, so we can try them out.

If things work out, I'd be keen to integrate this into avrdude because this is great to be able to use DRT/RTS directly on the reset pin! Your idea, @mariusgreuel, is a great improvement.

mcuee:
I will carry out more tests for this to see if there are regressions for STK500 v1 clones.

mcuee:
No regression for "Arduino as ISP" either. 

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1309 -c avrisp -P COM12 -b 19200 -p t88 -U t88_blink.hex

avrdude_pr1309: AVR device initialized and ready to accept instructions
avrdude_pr1309: device signature = 0x1e9311 (probably t88)
avrdude_pr1309: Note: flash memory has been specified, an erase cycle will be performed.
avrdude_pr1309: erasing chip
avrdude_pr1309: reading input file t88_blink.hex for flash
                with 8070 bytes in 1 section within [0, 0x1f85]
                using 127 pages and 58 pad bytes
avrdude_pr1309: writing 8070 bytes flash ...

Writing | ################################################## | 100% 12.59 s

avrdude_pr1309: 8070 bytes of flash written
avrdude_pr1309: verifying flash memory against t88_blink.hex

Reading | ################################################## | 100% 6.59 s

avrdude_pr1309: 8070 bytes of flash verified

avrdude_pr1309 done.  Thank you.

```

stefanrueger:
Made the change for wiring.c as well, so STK500 v2 bootloaders should work with direct reset, too.

Earmarked for the next mergefest (probably in the next days) so any objection speak up now

Issue #1311, opened 2023-03-07
Using serialupdi gives mismached error message


llinjupt:
@mcuee @stefanrueger 

When I was fiddling with serialupdi and tried to speed up with high baudrate, I got this error message.
It has nothing to do with ISP clock when using serialupdi which only supports baudrate setting.
Then I tried -B and it wouldn't take any effect on the uart baudrate.  I am not sure if it's a bug but at least it should show -b entry as well.

```
$ ./avrdude.exe -P COM245 -c serialupdi -p avr128da48 -U flash:r:test.bin:r -b 460800

avrdude error: UPDI link initialization failed
avrdude error: initialization failed, rc=-1
        - double check the connections and try again
        - use -B to set lower ISP clock frequency, e.g. -B 125kHz
        - use -F to override this check


avrdude done.  Thank you.
```

mcuee:
Which USB to TTL adapter are you using?

There are some good info from here. For example, make sure to change the latency timer of FT232R to 1ms in Windows Device Manager.
https://github.com/SpenceKonde/AVR-Guidance/blob/master/UPDI/jtag2updi.md

mcuee:
For good HW design, I think you can use @MCUdude's design can be a reference.
https://d3s5r33r268y59.cloudfront.net/datasheets/27214/2022-06-28-07-47-24/SerialUPDI_R1_Schematic.pdf

mcuee:
@llinjupt 
On the other hand, I tend to agree with you that the error message for serialupdi can be improved to mention to reduce the baud rate.

llinjupt:
> @llinjupt On the other hand, I tend to agree with you that the error message for serialupdi can be improved to mention to reduce the baud rate.

Yes, my point is the error message instead of the circuit. When I set -b 115200/230400, the programmer worked well.

MCUdude:
How about something like this? This will throw relevant errors for `serialupdi` and bitbanged serial programmers like `ponyser`. @stefanrueger I can submit a PR if this is an acceptable solution

```diff
index 68b55d4..1a9c8f9 100644
--- a/src/main.c
+++ b/src/main.c
@@ -1244,11 +1244,16 @@ int main(int argc, char * argv [])
   init_ok = (rc = pgm->initialize(pgm, p)) >= 0;
   if (!init_ok) {
     pmsg_error("initialization failed, rc=%d\n", rc);
     if(rc == -2)
       imsg_error("the programmer ISP clock is too fast for the target\n");
[...]

PR #1317, opened 2023-03-15
Print meaningful error message when using the serialupdi and bitbanged UART ISP programmers

MCUdude:
Resolves issue #1311


stefanrueger:
Earmarked for next mergefest

Issue #1319, opened 2023-03-16
Have Avrdude figure out the end address in terminal write mode


MCUdude:
Currently, there's no way to tell Avrdude that I want to write from a start address to the end of the specified memory when using the `...` ellipsis to fill the specified memory. Currently, you'll have to manually calculate how many bytes there are left in the memory (by looking in the datasheet for that particular AVR). And if you're not writing from start address 0, you'll have to subtract the start address from the total number of bytes the memory has in order to fill the entire memory.

Maybe the ellipsis `...` or asterisk symbol `*` can be used for this?

Below are a few examples for the ATmega1284P, which has 4096 bytes of EEPROM.
```
# write 4096 0x55's to EEPROM
# write eeprom 0 4096 0x55 ...

# Write data to EEPROM starting from address 0x100, and write 0x55's to the rest of the memory
#write eeprom 0x100 3840 "Hello World" 1234 0x55 ...


# Maybe we could use ... or * to specify "end of memory"?


# Fill the entire EEPROM with 0x55's using ... to indicate end of memory instead of length
write eeprom 0 ... 0x55 ...

# With asterisk
write eeprom 0 * 0x55 ...

# Fill the EEPROM memory starting from address 0x100
write eeprom 0x100 ... "Hello World" 1234 0x55 ...

# With asterisk
write eeprom 0x100 ... "Hello World" 1234 0x55 ...
```

Any thoughts?

mcuee:
Sounds like a good enhancement.

Just wondering if we can have a new `from start_addr to `end_addr` command.

mcuee:
The erase function is nice. Hopefully it works for different type of AVRs without erasing other menory types. Initial tests show the results are good.

MCUdude:
> The erase function is nice. Hopefully it works for different type of AVRs without erasing other menory types. Initial tests show the results are good.

Yes, I think this is a good addition to terminal "toolbox". Since it just calls `write <mem> 0 ... 0xff ...`, it should work with all memories the write command supports.

stefanrueger:
I am late to the party. I find the syntax with double ellipses confusing. Could we use negative numbers like in python or perl's substring operations? 

| addr |len | interval | comment
| --: | --: | --: | :-- |
| addr | len | [addr. addr+len-1] | current use |
| 0x1700 | 12 | [0x1700, 0x1700+11] | example |
| -256 | 2 | [size-256, size-255] | negative addr means that many bytes from end |
[...]

PR #1320, opened 2023-03-17
Terminal write and erase improvements

MCUdude:
#1319 related

The first new feature makes it easier to fill a section of a memory with data, as the user doesn't need to keep track of how many bytes there are left in the memory. I decided to use `...` for this, because the read command also uses  `...` for this: `read eeprom 0x100 ...`

```
# Fill the ATmega1284P EEPROM (4096 bytes) with 0x55s starting from address 0x100 and all the way to the end

# Current implementation
write eeprom 0x100 3840 0x55 ...

# New implementation
write eeprom 0x100 ... 0x55 ...
write eeprom 0x18a ... 0xaa ...
```

I've also extended the `erase` command. Now the erase command can be used to _only_ erase the user specified memory.

```
erase eeprom
erase flash
erase userrow
```

I've yet to update the docs, but I'd like to hear your thoughts first!


mcuee:
@MCUdude 

So you decided not to use `*` any more but only `...`, right?

This PR seems to work fine for a quick test.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320 -c pkobn_updi -p avr128da48 -t

                Vtarget         : 3.32 V
                PDI/UPDI clk    : 100 kHz

avrdude_pr1320: AVR device initialized and ready to accept instructions
avrdude_pr1320: device signature = 0x1e9708 (probably avr128da48)

avrdude> dump eeprom 0 0x40

Reading | ################################################## | 100% 0.45 s

0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0010  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write eeprom 0 * "Hello World" 1234 0x55 ...
avrdude_pr1320 error: (write ...) cannot parse length *
avrdude> write eeprom 0 ... "Hello World" 1234 0x55 ...

Caching | ################################################## | 100% 2.15 s

avrdude> flush
avrdude_pr1320: synching cache to device ...
Writing | ################################################## | 100% 15.21 s
avrdude> dump eeprom 0 0x40

Reading | ################################################## | 100% 0.17 s

0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 00 d2 04 55 55  |Hello World...UU|
0010  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|
0020  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|
0030  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|

avrdude> quit

avrdude_pr1320 done.  Thank you.

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320 -c pkobn_updi -p avr128da48 -t

                Vtarget         : 3.32 V
                PDI/UPDI clk    : 100 kHz

avrdude_pr1320: AVR device initialized and ready to accept instructions
avrdude_pr1320: device signature = 0x1e9708 (probably avr128da48)
avrdude> dump eeprom 0 0x40

Reading | ################################################## | 100% 0.45 s

0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 00 d2 04 55 55  |Hello World...UU|
0010  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|
0020  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|
0030  55 55 55 55 55 55 55 55  55 55 55 55 55 55 55 55  |UUUUUUUUUUUUUUUU|

avrdude> erase eeprom

Caching | ################################################## | 100% 2.12 s

avrdude> flush
avrdude_pr1320: synching cache to device ...
Writing | ################################################## | 100% 15.18 s
avrdude> dump eeprom 0 0x40

Reading | ################################################## | 100% 0.16 s

0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0010  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> dump flash 0 0x40

Reading | ################################################## | 100% 0.25 s

00000  0c 94 83 00 0c 94 a2 00  0c 94 a2 00 0c 94 a2 00  | ... ... ... ...|
00010  0c 94 a2 00 0c 94 a2 00  0c 94 a2 00 0c 94 a2 00  | ... ... ... ...|
00020  0c 94 a2 00 0c 94 a2 00  0c 94 a2 00 0c 94 a2 00  | ... ... ... ...|
00030  0c 94 a2 00 0c 94 a2 00  0c 94 a2 00 0c 94 a2 00  | ... ... ... ...|

avrdude> quit

avrdude_pr1320 done.  Thank you.

```

MCUdude:
> So you decided not to use * any more but only ..., right?

Yes, I'll stick with `...` unless there are good arguments to use something else, like `*` or `.`.
`...` means "to the end"/"end of memory" in the read/dump command, so there is a correlation here.

MCUdude:
@stefanrueger I've now added support for negative addresses and length bytes, just like you suggested in #1319.

Please give this PR a try and see if you are able to break anything. It appears to work and do what it should, I can start working on an even more improved erase command that also lets you specify an optional start address and length

mcuee:
Looks good.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320v1 -c usbtiny -p t13a -t

avrdude_pr1320v1: AVR device initialized and ready to accept instructions
avrdude_pr1320v1: device signature = 0x1e9007 (probably t13a)
avrdude> dump flash -256 256

Reading | ################################################## | 100% 0.49 s

0300  11 24 24 b6 14 be 80 e0  52 d0 21 fe 7d c0 8e e0  |.$$.....R.!.}...|
0310  4e d0 b9 9a 34 d0 82 30  71 f4 3f d0 c0 e6 d0 e0  |N...4..0q.?.....|
0320  10 e6 18 0f 2c d0 89 93  1c 13 fc cf 3b d0 a0 e6  |....,.......;...|
0330  b0 e0 44 d0 0e c0 83 30  41 f4 2f d0 c8 2f 32 d0  |..D....0A./../2.|
0340  85 91 0a d0 c1 50 e1 f7  04 c0 82 30 08 f4 ff cf  |.. ..P.....0....|
0350  29 d0 8b e0 01 d0 de cf  2a e0 80 95 08 94 10 f4  |).......*.......|
0360  c1 98 02 c0 c1 9a 00 00  04 d0 86 95 2a 95 b9 f7  |............*...|
0370  08 95 00 d0 96 e0 9a 95  f1 f7 00 c0 08 95 29 e0  |..............).|
0380  b0 99 fe cf f7 df f5 df  88 94 b0 99 08 94 2a 95  |..............*.|
0390  11 f0 87 95 f8 cf a8 95  08 95 f1 df e8 2f ef df  |............./..|
03a0  f8 2f ed cf ec df 80 32  f9 f7 80 e2 d5 cf 98 e1  |./.....2........|
03b0  a8 95 91 bd 81 bd 08 95  fb 01 dc 01 60 e2 6a 0f  |............`.j.|
03c0  9f 01 f3 30 b8 f4 83 e0  10 d0 30 96 29 f4 8f e7  |...0......0.)...|
03d0  8d 93 8f ec 8c 93 11 97  81 e0 0d 90 1d 90 05 d0  |.......... .....|
03e0  32 96 a6 13 fa cf f9 01  85 e0 87 bf e8 95 07 b6  |2...............|
03f0  00 fc fd cf 11 24 08 95  ff ff 08 04 dd cf 26 3f  |.....$........&?|

avrdude> dump flash -512 256

Reading | ################################################## | 100% 0.50 s

0200  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0210  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0220  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0230  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0240  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0250  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0260  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0270  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0280  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0290  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write flash -512 128 0xaa

Caching | ################################################## | 100% 0.01 s

avrdude> flush
avrdude_pr1320v1: synching cache to device ...
Writing | ################################################## | 100% 0.08 s
avrdude> dump flash -512 256

Reading | ################################################## | 100% 0.25 s

0200  80 aa ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0210  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0220  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0230  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0240  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0250  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0260  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0270  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0280  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0290  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write flash -512 128 0xaa ...

Caching | ################################################## | 100% 0.26 s

avrdude> flush
avrdude_pr1320v1: synching cache to device ...
Reading | ################################################## | 100% 0.68 s
Erasing | ################################################## | 100% 0.08 s
Writing | ################################################## | 100% 0.96 s
avrdude> dump flash -512 256

Reading | ################################################## | 100% 0.26 s

0200  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0210  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0220  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0230  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0240  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0250  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0260  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0270  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
0280  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
0290  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
02f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> quit

avrdude_pr1320v1 done.  Thank you.

```

mcuee:
Two missing things.
1) `erase mem` is not documented yet in the terminal `help` command
2) `erase mem add len` proposed by @stefanrueger has not been impelmented.
* https://github.com/avrdudes/avrdude/issues/1319#issuecomment-1474918710

stefanrueger:
Review:

 - This [line](https://github.com/MCUdude/avrdude/blob/f287a5908acd4fe7bb70f2e7bff8845cd3181d49/src/term.c#L323) misses `+ 1` for the computation of the effective length
 - After computing the new effective length the code misses a range check for the new `len`; I predict `read ee -1 -2`  might crash or, if we are lucky, give some weird error message
 - The next line sets len = 1 when the effective length was 0; this used to be innocuous for `read`; I am no longer sure we want this now as write also uses the same technique; do we want `write` to write one byte when the user asks to write zero bytes, so asking read to read zero bytes should perhaps do nothing?
 - This [line](https://github.com/MCUdude/avrdude/blob/f287a5908acd4fe7bb70f2e7bff8845cd3181d49/src/term.c#L503) also miscomputes the effective length; should be `len = maxsize+len - addr + 1`;
 - The next line should have a range check for the now effective length against < 0 and > maxsize-addr and abort with error here "effective address 0x%04x and effective length %d not compatible with memory size %d"; I wonder what happens with `write ee 1023 -1025 0xff ...` using an EEPROM with 1024 bytes (eg, m238p)?
 - There is a regression: `erase` alone no longer works b/c the new code returns an error when `argc < 2` but erase alone will be `argc == 1`
 - The erase code changes are surprisingly many lines... would it be OK to do something like
   ```
     if(argc > 4 || argc == 3) {
       msg_error("Usage: erase [<memory> [<addr> <len>]]\n");
       return -1;
     }
      
     if(argc > 1) {
       // @MCUdude: Insert code here for erase memory and erase memory addr len
       
       return 0; // @MCUdude: ensure code returns from this if block
     }
   ```
  This would ensure the existing code could stay on unchanged.

Note that `strtol()` is far more difficult to handle than `strtoul()` b/c it needs range checking with `errno`, see its manual page; you /might/ get away with not checking `errno` here (b/c the numbers involved are normally only a few 100,000s not in the billions when you see this effect) but it is not good practice to cut the corners (not checking errno). However, I have a plan to replace all atoi(), strtol() etc functions with a dedicated avr_str2num() function (generic conversion will be able to handle binary etc), so don't worry about that now. I will replace these anyway later on.

MCUdude:
Thanks for the review @stefanrueger! I had already figured out the issues pointed out in bullet point 1 and two, but I hadn't pushed the code yet.

I'll spend some time to see if I can get things right

MCUdude:
@stefanrueger I've fixed the issues you pointed out, and I've also re-implemented the erase command to prevent modifying existing code. Feel free to review again!

src/term.c
```diff
@@ -307,14 +313,22 @@ static int cmd_dump(PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]) {
...
-      if (ul == 0 || ul > INT_MAX) // Not positive if used as int, make it 1
-        ul = 1;
-      read_mem[i].len = (int) ul;
+      // Turn negative len value (no. bytes from top of memory) into an actual length number
+      if (len < 0)
+        len = maxsize + len + 1 - read_mem[i].addr;
+
+      if (len == 0)
+        return 0;
+      else if (len < 0) {
```
stefanrueger, line 327: This is not wrong, but I normally refrain from elsing a return

src/term.c
```diff
@@ -307,14 +313,22 @@ static int cmd_dump(PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]) {
...
+        pmsg_error("(dump) invalid length %d\n", len);
```
stefanrueger, line 328: Could be called "invalid effective length" rather than "invalid length" b/c if the user requests "read ee -1 -3" the length they set is -3, but the effective length would be -1, ie, memsize-2-(memsize-1)+1...

src/term.c
```diff
@@ -287,17 +287,23 @@ static int cmd_dump(PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]) {
...
-        mem->size > 0x10000? 5: 4, maxsize-1);
+
+    // Turn negative addr value (counting from top and down) into an actual memory address
+    if (addr < 0)
+      addr = maxsize + addr;
+
+    if (addr < 0 || addr >= maxsize) {
+      int digits = mem->size > 0x10000? 5: 4;
+      pmsg_error("(dump) %s address %s is out of range [-0x%0*x, 0x%0*x]\n",
+        mem->desc, argv[2], digits, maxsize, digits, maxsize-1);
```
stefanrueger, line 303: I like that the valid address range now reads [-maxsize, maxsize-1]; I hadn't thought of that, good.

src/term.c
```diff
@@ -479,12 +499,22 @@ static int cmd_write(PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]) {
...
     }
+    // Turn negative len value (no. bytes from top of memory) into an actual length number
+    if (len < 0)
+      len = maxsize + len - addr + 1;
+    if (len == 0)
+      return 0;
+    if (len < 0 || len > maxsize - addr) {
+      pmsg_error("(write) effective %s start address 0x%0*x and effective length %d not compatible with memory size %d\n",
+        mem->desc, maxsize > 0x10000? 5: 4, addr, len, maxsize);
+      return -1;
```
stefanrueger, line 516: Good to mention which memory. I thought mentioning (write ...) not (write) is better (as it's the write fill mode here)

src/term.c
```diff
@@ -797,6 +827,32 @@ static int cmd_send(PROGRAMMER *pgm, AVRPART *p, int argc, char *argv[]) {
...
+  }
+
+  if (argc > 1) {
+    char *memtype = argv[1];
+    AVRMEM *mem = avr_locate_mem(p, memtype);
+    if (mem == NULL) {
+      pmsg_error("(erase) %s memory type not defined for part %s\n", argv[1], p->desc);
+      return -1;
+    }
+    char *args[6] = {"write", memtype, "", "", "0xff", "..."};
```
stefanrueger, line 842: This is not wrong, but tradition has it that argv lists are terminated by NULL, so just for superstition, I'd write `char *args[] = {"write", memtype, "", "", "0xff", "...", NULL};` Admittedly, it's unlikely that the write code ever looks for the terminating NULL.

stefanrueger:
earmarked for the next mergefest

stefanrueger:
Ahh, I forgot: Maybe the documentation of the read ellipses could be downgraded to simplify documentation, eg, removing the ellipses from the terminal help menus for read, from avrdude.1 and avrdude.texi; in the latter two I'd add a sentence in the read description to the effect of `read ...` returns the full memory as shortcut and `read addr ...` returns the memory from `addr` to the end. Then erase [memory [start len]] ought to be documented, as should the use of negative addresses and length in read, fill write and erase.

src/term.c
MCUdude: Thanks for your input, I wasn't aware of this so if I add a NULL to the args array, should I pass 6 or 7 as the argc number to cmd_write?

src/term.c
MCUdude: Yes , that's a good idea!

src/term.c
MCUdude: Sorry, that's a mistake on my side. I'll have it fixed

src/term.c
stefanrueger: argc must be 6 (no change)

mcuee: https://en.cppreference.com/w/cpp/language/main_function

argc	-	Non-negative value representing the number of arguments passed to the program from the environment in which the program is run.

argv	-	Pointer to the first element of an array of argc + 1 pointers, of which the last one is null and the previous ones, if any, point to null-terminated multibyte strings that represent the arguments passed to the program from the execution environment. If argv[0] is not a null pointer (or, equivalently, if argc > 0), it points to a string that represents the name used to invoke the program, or to an empty string.

stefanrueger:
Would a table with formulae and examples be OK? Sth like

The  `<addr>` and `<len>` parameters of the read, write fill and erase commands can be negative with the same syntax as substring() computations in perl or python. The table below details their meaning with respect to a memory of `size=0x800`:

| addr |len | memory interval | comment
| --: | --: | --: | :-- |
| *positive* | *positive* | `[addr, addr+len-1]` | conventional use (note len = end - start +1) |
| 0x700 | 12 | `[0x700, 0x711]` | example |
| *positive* | *negative* | `[addr, size+len]` | end point is that many bytes below size|
| 1024 | -257 | `[0x400, 0x6ff]` | size of memory is `2048` or `0x800` |
| *negative* | *positive* | `[size+addr, size+addr+len-1]` | start point is that many bytes below size |
| -512 | 512 | `[0x600, 0x7ff]` | last 512 bytes |
| *negative* | *negative* | `[size+addr, size+len]` | |
| -512 | -512 | `[0x600, 0x600]` | one byte 512 bytes from top memory |
| -256 | -1 | `[0x700, 0x7ff]` | the last 256 bytes |
| *any*  | *zero* | empty set | no action |
| *zero*  | *positive* | `[0, len]` | first len bytes |
| *zero* | *negative* | `[0, addr+len]` | all but the last `\|len+1\|` bytes |
| 0 | -1 | `[0, 0x7ff]` | all memory without knowing its size |

Please double check examples!

src/term.c
stefanrueger, line 1: Also the overall help message needs changing from `"%s <memory> [<addr> <len> | <addr> ... | <addr> | ...]"` to `"%s <memory> [<addr> | <addr> <len>]"`

MCUdude:
Thanks for the table @stefanrueger! Was the idea to add this table to both avrdude.1 and avrdude.texi?

stefanrueger:
Prob only in avrdude.texti. It's sufficient to explain the key rows of the table in avrdude.1

src/term.c
stefanrueger: I had a look at the `tokenize()` function that provides the argc, argv pair for all the commands in the terminal. Indeed, that uses a terminating NULL (as convention has it). So better to change that line to `char *args[] = {"write", memtype, "", "", "0xff", "...", NULL};`

stefanrueger:
Cool updates.

The table in avrdude.1 could do with explaining negative addr. The table in avrdude.texi might need some fiddling (shortening columns, comments etc) to look good:

Screenshot from 2023-03-22 20-55-27

What are you going to do to resolve the three conflicts with the base?

stefanrueger:
> How about removing all references to specific hardware?

They don't worry me. The text before says "The terminal commands below may only be implemented on some specific programmers", so in terms of examples that's prob OK.

MCUdude:
> The table in avrdude.texi might need some fiddling (shortening columns, comments etc) to look good:

I haven't been able to export the text file to PDF. The formatting problem is probably caused by the following line:
```
@multitable @columnfractions .16 .10 .35 .48
```

Can you try to tweak these numbers until you get a table that looks right?

> What are you going to do to resolve the three conflicts with the base?

Rebase, merge, and resolve the merge conflict.

stefanrueger:
> tweak these numbers

Might be good for these to add up to 1 or slightly below and change text to be crisper? End is `|len|` below size rather than End address is `|len|` below memory size; Note: len  = end-start +1; that kind of thing. You did similar things to the avrdude.1 table, eg 0/pos, neg, etc

src/avrdude.1
```diff
@@ -848,7 +848,12 @@ so previously entered command lines can be recalled and edited.
 .Pp
 The <addr> and <len> parameters of the read, write and erase commands can be
 negative with the same syntax as substring() computations in perl or python.
-The table below details their meaning with respect to a memory of size=0x800:
+The table below details their meaning with respect to a memory of size=0x800.
+Note that
+.Ar pos
+means positive, and
```
stefanrueger, line 854: Readers will be able to guess what pos or neg might mean, but the table below lacks the row with the general case when `addr` is negative, so they'd have to work that back from the couple of examples. 


| `addr` |`len` | Memory interval | Comment |
| --: | --: | --: | :-- |
| negative | positive | [size+addr, size+addr+len-1] | Start address is \|addr\| bytes below memory size |

MCUdude: The problem is that I'm having a _real hard time_ getting everything to fit within 75 characters (80 - 5 indents).
Can some of the memory intervals or examples be re-written or reformatted to fir within 75 characters? I'm all ears!

```
                                             75 character limit ends here!
                                                                         |
                                                                         |
addr   len  Memory interval             Comment                          v
--------------------------------------------------------------------------
0/pos  pos  [addr, addr+len-1]          Conventional use, len=end-start+1
0/pos  neg  [addr, size+len]            End addr |len| bytes below mem size
neg    pos  [size+addr,size+addr+len-1] Start addr is |addr| bytes below mem size
neg    neg  [size+addr, size+len]       Combining above two cases
0x700  12   [0x700, 0x711]              Example
1024   -257 [0x400, 0x6ff]              Here, size of mem is 2048 or 0x800
-512   512  [0x600, 0x7ff]              Last 512 bytes
-256   -1   [0x700, 0x7ff]              Last 256 bytes
0      49   [0, 48]                     First 49 bytes
0      -1   [0, 0x7ff]                  Entire mem without knowing its size
```

MCUdude:
> Might be good for these to add up to 1 or slightly below and change text to be crisper? End is |len| below size rather than End address is |len| below memory size; Note: len = end-start +1; that kind of thing. You did similar things to the avrdude.1 table, eg 0/pos, neg, etc

The problem is that I know nothing about Texinfo, and I don't have a tool installed that let me see what the end result looks like, except for a VSCode plugin that's nothing like a "real" PDF output.

When I try to export avrdude.texi to pdf, I'm getting an error saying that there are no TeX binary installed.

```
$ texi2dvi --pdf doc/avrdude.texi 
You don't have a working TeX binary (tex) installed anywhere in
your PATH, and texi2dvi cannot proceed without one.  If you want to use
this script, you'll need to install TeX (if you don't have it) or change
your PATH or TEX environment variable (if you do).  See the --help
output for more details.

For information about obtaining TeX, please see http://tug.org/texlive,
or do a web search for TeX and your operating system or distro.
```

However, when I try to install it, I'm getting an error saying that my macOS version is too old, and I'm not ready for a clean install of macOS just yet.

```
$ brew install texlive
Warning: You are using macOS 10.14.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
Any opened issues will be immediately closed without response.
Do not ask for help from MacHomebrew on Twitter.
You may ask for help in Homebrew's discussions but are unlikely to receive a response.
Try to figure out the problem yourself and submit a fix as a pull request.
We will review it but may or may not accept it.

openjdk: A full installation of Xcode.app is required to compile
this software. Installing just the Command Line Tools is not sufficient.

Xcode can be installed from the App Store.
openjdk: macOS Catalina or newer is required for this software.
Error: texlive: Unsatisfied requirements failed this build.
```

So I'm asking kindly; can you @stefanrueger see if you can figure out what needs to be done in order to get the table looking OK? Thanks!

mcuee:
> I know humans can live without food for a couple of weeks, without water for a couple of days, but without TeX?!?

Haha, engineers can usually live without TeX. :-)

Last time when disk space is a constraint, TeX was a space hogger and often got unstalled first by me. Now it will probbaly be LibreOffice under Linux.

MCUdude:
> OK, not a prob. Have pushed a commit. See how you like it. 72 chars and fitting TeX table

The 72-character table looks alright, but is the `sz` supposed to be there?

```
0/pos   neg  [addr, sz+len]      End is |len| bytes below memory size sz
```

TeX table looks neat! 

> I know humans can live without food for a couple of weeks, and without water for a couple of days, but without TeX?!?

I know, I'm a freak. We actually never learned to use TeX at the university but instead pushed Microsoft software instead. And now I don't really have an excuse to spend time learning TeX when I rarely (ok, never) write academic papers anymore.

mcuee:
> However, when I try to install it, I'm getting an error saying that my macOS version is too old, and I'm not ready for a clean install of macOS just yet.

@MCUdude 

You may want to get MacTeX 2023 a try as it still support macOS 10.14. It is a huge package though.
 https://tug.org/mactex/mactex-download.html

They do have a smaller package as well.
https://tug.org/mactex/morepackages.html

stefanrueger:
|is the sz supposed to be there?

It's a reminder of naming the memory size sz. Feel free to remove it if you think it's clear enough from the last sentence before the table... Other than that we're getting in the territory of diminishing returns. Happy to merge during the next mergefest unless @mcuee can break it.

mcuee:
@MCUdude and @stefanrueger 

I have done some simple tests and so far so good. Give me a few days more so that I can carry out more tests.

```
PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320v1 -c usbasp -p m128a -t

avrdude_pr1320v1: AVR device initialized and ready to accept instructions
avrdude_pr1320v1: device signature = 0x1e9702 (probably m128a)
avrdude> help
Valid commands:
  dump    : dump <memory> [<addr> | <addr> <len>]
  read    : alias for dump
  write   : write <memory> <addr> <data>[,] {<data>[,]}
          : write <memory> <addr> <len> <data>[,] {<data>[,]} ...
  flush   : synchronise flash & EEPROM writes with the device
  abort   : abort flash & EEPROM writes (reset the r/w cache)
  erase   : perform a chip erase
  sig     : display device signature bytes
  part    : display the current part information
  send    : send a raw command: send <b1> <b2> <b3> <b4>
  sck     : set <SCK period>
  verbose : change verbosity
  quell   : set quell level for progress bars
  help    : show help message
  ?       : same as help
  quit    : quit after writing out cache for flash & EEPROM
  q       : abbreviation for quit

Note that not all programmer derivatives support all commands. Flash and
EEPROM type memories are normally read and written using a cache via paged
read and write access; the cache is synchronised on quit or flush commands.
The part command displays valid memory types for use with dump and write.

avrdude> dump flash 0x200 0x100

Reading | ################################################## | 100% 0.30 s

00200  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00210  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00220  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00230  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00240  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00250  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00260  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00270  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00280  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00290  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002a0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002b0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002e0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
002f0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> write flash 0x200 0x100 0xaa ...

Caching | ################################################## | 100% 0.25 s

avrdude> flush
avrdude_pr1320v1: synching cache to device ...
Writing | ################################################## | 100% 0.09 s
avrdude> dump flash 0x200 0x100

Reading | ################################################## | 100% 0.26 s

00200  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00210  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00220  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00230  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00240  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00250  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00260  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00270  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00280  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
00290  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002a0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002b0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002c0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002d0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002e0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
002f0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|

avrdude> dump flash -256 -1

Reading | ################################################## | 100% 0.30 s

1ff00  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff10  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff20  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff30  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff40  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff50  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff60  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff70  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff80  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ff90  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ffa0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ffb0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ffc0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ffd0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1ffe0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
1fff0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> dump flash -256 0 (Note: do nothing as expected)
avrdude> dump flash -256 1

Reading | ################################################## | 100% 0.00 s

1ff00  ff                                                |.               |

avrdude> write flash -0x100 0x100 0xaa ...

Caching | ################################################## | 100% 0.25 s

avrdude> flush
avrdude_pr1320v1: synching cache to device ...
Writing | ################################################## | 100% 0.09 s
avrdude> dump flash -256 0x100

Reading | ################################################## | 100% 0.25 s

1ff00  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff10  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff20  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff30  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff40  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff50  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff60  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff70  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff80  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ff90  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ffa0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ffb0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ffc0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ffd0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1ffe0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|
1fff0  aa aa aa aa aa aa aa aa  aa aa aa aa aa aa aa aa  |................|

avrdude> dump flash 0 0x10

Reading | ################################################## | 100% 0.09 s

00000  54 68 65 20 71 75 69 63  6b 20 62 72 6f 77 6e 20  |The quick brown |

avrdude> quit

avrdude_pr1320v1 done.  Thank you.

```

MCUdude:
While we still have a terminal mode PR open.
I've noticed that some commands print an extra newline character when the command is finished, while other doesn't. Is this deliberate, or should we do something about this? In the output below, you can see that `erase`, `flush`, `abort`, `sck`, `verbose`, and `quell` doesn't print an extra line, but there may be other commands as well.

```
$ ./avrdude -cusbasp -patmega328p -t

avrdude: device signature = 0x1e950f (probably m328p)
avrdude> ?
Valid commands:
  dump    : dump <memory> [<addr> | <addr> <len>]
  read    : alias for dump
  write   : write <memory> <addr> <data>[,] {<data>[,]}
          : write <memory> <addr> <len> <data>[,] {<data>[,]} ...
  flush   : synchronise flash & EEPROM writes with the device
  abort   : abort flash & EEPROM writes (reset the r/w cache)
  erase   : perform a chip erase
  sig     : display device signature bytes
  part    : display the current part information
  send    : send a raw command: send <b1> <b2> <b3> <b4>
  sck     : set <SCK period>
  verbose : change verbosity
  quell   : set quell level for progress bars
  help    : show help message
  ?       : same as help
  quit    : quit after writing out cache for flash & EEPROM
  q       : abbreviation for quit

Note that not all programmer derivatives support all commands. Flash and
EEPROM type memories are normally read and written using a cache via paged
read and write access; the cache is synchronised on quit or flush commands.
The part command displays valid memory types for use with dump and write.

avrdude> read eeprom 0 16

Reading | ################################################## | 100% 0.01 s 

0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|

avrdude> sig
Device signature = 0x1e950f

avrdude> erase
erasing chip ...
avrdude> part

AVR Part                      : ATmega328P
Chip Erase delay              : 9000 us
PAGEL                         : PD7
BS2                           : PC2
RESET disposition             : possible i/o
Timeout                       : 200
StabDelay                     : 100
CmdexeDelay                   : 25
SyncLoops                     : 32
PollIndex                     : 3
PollValue                     : 0x53

  Memory Type Alias    Mode Delay Size  Indx Paged  Size   Size #Pages MinW  MaxW   ReadBack
  ----------- -------- ---- ----- ----- ---- ------ ------ ---- ------ ----- ----- ---------
  eeprom                 65    20     4    0 no       1024    4      0  3600  3600 0x00 0x00
  flash                  65    10   128    0 yes     32768  128    256  4500  4500 0x00 0x00
  lfuse                   0     0     0    0 no          1    1      0  4500  4500 0x00 0x00
  hfuse                   0     0     0    0 no          1    1      0  4500  4500 0x00 0x00
  efuse                   0     0     0    0 no          1    1      0  4500  4500 0x00 0x00
  lock                    0     0     0    0 no          1    1      0  4500  4500 0x00 0x00
  signature               0     0     0    0 no          3    1      0     0     0 0x00 0x00
  calibration             0     0     0    0 no          1    1      0     0     0 0x00 0x00

avrdude> flush
avrdude> write eeprom 0 "Hello World!"

Caching | ################################################## | 100% 0.00 s 

avrdude> flush
avrdude: synching cache to device ... 
Writing | ################################################## | 100% 0.17 s 
avrdude> write eeprom 16 "Ja vi elsker"

Caching | ################################################## | 100% 0.01 s 

avrdude> abort
avrdude> sck 1
avrdude: set SCK frequency to 750000 Hz
avrdude> quell 2
New quell level: 2
avrdude> verbose 2
New verbosity level: 2
avrdude> quit

avrdude done.  Thank you.
```

stefanrueger:
Good catch, some commands have that extra newline, most notably read. Some of the extra newlines may be hard to find: the interaction of verbose progress reports, potential errors and normal output can be hard to understand/manage. But when called with `-qq` (my favourite AVRDUDE option!) there probably shouldn't be any unwarranted newlines (my tuppence).

mcuee:
> While we still have a terminal mode PR open. I've noticed that some commands print an extra newline character when the command is finished, while other doesn't. Is this deliberate, or should we do something about this? In the output below, you can see that `erase`, `flush`, `abort`, `sck`, `verbose`, and `quell` doesn't print an extra line, but there may be other commands as well.

It will be good to clean up and minimize the differences. 

I will suggest we remove most of the new lines. For example, I think we should remove the new line immediately after the `dump` and `write` command. Same for `quit` command.

mcuee:
> Good catch, some commands have that extra newline, most notably read. Some of the extra newlines may be hard to find: the interaction of verbose progress reports, potential errors and normal output can be hard to understand/manage. But when called with `-qq` (my favourite AVRDUDE option!) there probably shouldn't be any unwarranted newlines (my tuppence).

I think we just need to reduce the number of new lines. Line breaks are good in some places (eg: verbose reports, errors report). But there are simply too many new lines right now.

As for `-qq`, I find it useful.

But I am reducing the usage of `-qq` in my test report now as it is omitting some useful information like the time takes to finish the read and write command.

mcuee:
> I have done some simple tests and so far so good. Give me a few days more so that I can carry out more tests.

@MCUdude and @stefanrueger 

I have done more tests under Windows and Linux and this PR is good to go.

We can deal with the new line improvement in another PR.

MCUdude:
> We can deal with the new line improvement in another PR.

Yes, most of the commands are trivial to fix, but the dump and write progress bars would need some work

stefanrueger:
> progress bar would need some work

Indeed, it does. I have pushed a commit onto this PR that addresses double newlines. In a lot of cases a newline is printed to ensure whatever the output at the screen the next thing is written on a fresh line. That often causes extra newlines in case the output on screen was already terminated by a newline earlier. Naturally the function wanting to start output on a new line has no idea what's the state of the screen.

 My idea was to hijack a vertical tab `\v` when it's the first character in a `msg_xyz(format, args, ...)` format string and
 - Print nothing if the output is already at column 0
 - Print a newline `\n` if the output is not at column 0
 
Effectively this allows `msg_info("\v");` to print a *conditional* newline. I have then used this technique (gratuitously?) in `term.c` to ensure that there are newlines where there should be newlines, but also to avoid a second (unnecessary) newline. As this affects progress reporting *everywhere* (not only the terminal) this means that a lot of paragraph breaks around progress reporting melt away. I personally *really like that*, but I am also sure that other people have a different taste and *want* progress reporting surrounded by empty space (double newlines). Give it a go, and see how you like that @mcuee, @MCUdude, @dl8dtl, @mariusgreuel.

MCUdude:
It's great that lots of unnecessary double new lines are now gone. I slightly prefer double newlines before and after the progress bars, but It's not a deal breaker. Let's hear what the other dudes have to say!

mcuee:
@stefanrueger and @MCUdude 
I am thinking we may want to remove the new line after `sig` command, and maybe `quit` command as well.

Latest testing results: 

```
C:\work\avr\avrdude_test\avrdude_bin> .\avrdude -c usbasp -p m328p
 -U .\urboot_uno_autobaud_ee_led+b5_fr_ce_ur_vbl.hex -qq && echo OK
OK
C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320v2 -C .\avrdude_pr1320v2.conf
 -c urclock -P COM4 -p m328p -qqt
avrdude> read eeprom 0 16
0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
avrdude> sig
Device signature = 0x1e950f

avrdude> erase
erasing chip ...
avrdude> write eeprom 0 "Hello World!"
avrdude> flush
avrdude> read eeprom 0 16
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 ff ff ff  |Hello World!....|
avrdude> quit

C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_pr1320v2 -C .\avrdude_pr1320v2.conf
 -c urclock -P COM4 -p m328p -t

avrdude_pr1320v2: AVR device initialized and ready to accept instructions
avrdude_pr1320v2: device signature = 0x1e950f (probably m328p)
avrdude> read eeprom 0 16
Reading | ################################################## | 100% 0.08 s
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 ff ff ff  |Hello World!....|
avrdude> sig
Device signature = 0x1e950f

avrdude> erase
erasing chip ...
avrdude> read eeprom 0 16
Reading | ################################################## | 100% 0.05 s
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 ff ff ff  |Hello World!....|
avrdude> write eeprom 0 "Hello World!"
Caching | ################################################## | 100% 0.04 s
avrdude> flush
avrdude> read eeprom 0 16
Reading | ################################################## | 100% 0.05 s
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 ff ff ff  |Hello World!....|
avrdude> erase eeprom
Caching | ################################################## | 100% 1.12 s
avrdude> flush
avrdude_pr1320v2: synching cache to device ...
Writing | ################################################## | 100% 0.07 s
avrdude> read eeprom 0 16
Reading | ################################################## | 100% 0.05 s
0000  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
avrdude> write eeprom 0 "Hello World!"
Caching | ################################################## | 100% 0.04 s
avrdude> flush
avrdude_pr1320v2: synching cache to device ...
Writing | ################################################## | 100% 0.08 s
avrdude> read eeprom 0 16
Reading | ################################################## | 100% 0.05 s
0000  48 65 6c 6c 6f 20 57 6f  72 6c 64 21 00 ff ff ff  |Hello World!....|
avrdude> sck
avrdude_pr1320v2 error: (cmd) command sck is invalid
avrdude> quit

avrdude_pr1320v2 done.  Thank you.
```

stefanrueger:
> remove the new line after `sig`

@mcuee Good catch; done, and also for the `send` command. Re `quit`: the double newline is brought about by main
https://github.com/avrdudes/avrdude/blob/7a8d257676ca858f4c00e61c09885234c64a880f/src/main.c#L1474

Not sure we can touch this. It's something that has probably been like this for a couple of decades.

mcuee:
> > remove the new line after `sig`
> 
> @mcuee Good catch; done, and also for the `send` command. Re `quit`: the double newline is brought about by main
> 
> https://github.com/avrdudes/avrdude/blob/7a8d257676ca858f4c00e61c09885234c64a880f/src/main.c#L1474
> 
> Not sure we can touch this. It's something that has probably been like this for a couple of decades.

@stefanrueger 

I see. No problem. We can keep the behavior for `quit`.

MCUdude:
I've just updated the terminal examples to reflect the "new" style where excess newlines were printed. However, We'll have to update a few examples in avrdude.texi as well, under the "Example Command Line Invocations" section.

MCUdude:
> As it's meant to be examples, it might (ever so slightly) improve clarity if we pretended (just for the examples) that the prompt was \navrdude> and not avrdude>, ie, leave a newline before every prompt in the examples. I think journalists call this technique "truth, well told", ie, a slight deviation from reality with the intent to show the substance more vividly (here, the commands and the effect they have).

You're right, and I just rebased the previous commit. I added a few new lines; the examples are easier to figure out just by looking at them.

BTW I've now updated the "regular" CLI examples, so this PR should be good to go now.

stefanrueger:
> BTW I've now updated the "regular" CLI examples, so this PR should be good to go now.

Excellent, that's a substantial improvement of the terminal now. Planning to merge in the next few days hopefully along with the ch341a programmer and three smaller PRs.

PR #1341, opened 2023-04-09
Move avrintel into libavrdude

### LINKED ISSUE #1334: Build for Bob3
I hope I am stepping not on anybodies toes here....
My wife works in STEM education and uses the Bob3 "robot" for basic programming introduction.
They have a Linux version of their SW at https://www.bob3.org/de/linux .  This uses the library part of avrdude to do the actual transfer of data to the MCU. Not sure if this was/is intended by the project to be used this way. 
Anyhow building avrdude for this purpose I found two issues and propose the following changes.

a) Move the avrintel functions into the library (as opposed to directly linking it into the executable
```
--- src/CMakeLists.txt.orig     2023-03-03 20:05:29.274700258 +0100
+++ src/CMakeLists.txt  2023-03-05 12:44:47.583333162 +0100
@@ -145,6 +145,8 @@
     avrftdi_private.h
     avrftdi_tpi.c
     avrftdi_tpi.h
+    avrintel.c
+    avrintel.h
     avrpart.c
     bitbang.c
     bitbang.h
@@ -278,8 +280,6 @@
     main.c
     term.c
     term.h
-    avrintel.c
-    avrintel.h
     developer_opts.c
     developer_opts.h
     developer_opts_private.h
```

b) Would you be willing to accept a patch to build.sh that makes build_flags a parameter that can be passed to the skript? I am generating an RPM for avrdude and want to install to /usr (resp. /etc for the config file).

stefanrueger:
Fixes #1334

Avrintel should be in the library as one would get unresolved references otherwise when using some functionality in libavrdude. It also provides an instant knowledge base about parts (and programmers) for any project linking to libavrdude without having to manage a relatively complex `avrdude.conf`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -151,6 +151,8 @@ add_library(libavrdude
     avrftdi_private.h
     avrftdi_tpi.c
     avrftdi_tpi.h
+    avrintel.c
+    avrintel.h
     avrpart.c
     bitbang.c
     bitbang.h
@@ -288,8 +290,6 @@ add_executable(avrdude
     main.c
     term.c
     term.h
-    avrintel.c
-    avrintel.h
     developer_opts.c
     developer_opts.h
     developer_opts_private.h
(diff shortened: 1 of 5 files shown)


mcuee (approved):
This should be safe to merge.

PR #1346, opened 2023-04-20
Add case-insensitive string comparison and wildcard match to str_*() ...

stefanrueger:
…routines and apply new str_*() functions in `urclock.c` and `developer_opts.c`

Should be relatively uncontroversial but needs a pair of eyes (@MCUdude?)

### DIFF
--- a/src/libavrdude.h
+++ b/src/libavrdude.h
@@ -403,9 +403,7 @@ typedef void (*walk_avrparts_cb)(const char *name, const char *desc,
 void walk_avrparts(LISTID avrparts, walk_avrparts_cb cb, void *cookie);
 void sort_avrparts(LISTID avrparts);
 
-int strcase_eq(const char *str1, const char *str2);
-int part_match(const char *pattern, const char *string);
-// cmp can be, eg, strcase_eq or part_match
+// cmp can be, eg, str_caseeq or str_casematch
 int part_eq(AVRPART *p, const char *string, int (*cmp)(const char *, const char *));
 
 int compare_memory_masked(AVRMEM * m, uint8_t buf1, uint8_t buf2);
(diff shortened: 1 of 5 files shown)


src/libavrdude.h
```diff
@@ -1161,6 +1159,11 @@ int str_eq(const char *str1, const char *str2);
 int str_contains(const char *str, const char *substr);
 int str_ends(const char *str, const char *ends);
```
MCUdude, line 1161: Since all `str*` functions "belongs together", maybe you can remove the blank like here?  Then it looks more like these belong together, which they absolutely do; they are all string-checking functions.

src/strutil.c
```diff
@@ -41,3 +42,142 @@ int str_ends(const char *str, const char *ends) {
 
   return str_eq(str + str_len - ends_len, ends);
 }
+
```
MCUdude, line 45: Perhaps we should have a single-line comment above the new functions to describe what they do quickly.
I didn't remember what `strncasecmp` did, so I had to look it up. And [drumroll], it compares strings without looking at the casing.

How about something like this?
```
// Check if [str] starts with [starts] without case sensitivity
```

stefanrueger:
Great shouts @MCUdude Followed both suggestions in new commit

src/strutil.c
MCUdude: Excellent, thanks!

src/strutil.c
```diff
@@ -63,9 +70,11 @@ int str_caseends(const char *str, const char *ends) {
```
MCUdude, line 70: This is really knit-picking, but you can remove a blank line here as well so that there is a single blank line between the functions.

stefanrueger: > single blank line between the functions

Uhhhm, actually, I follow a different style in that I break groups of functions doing similar things with *two* blank lines (also if they are longer). In that respect it's a bit like breaking up text into paragraphs in ordinary writing. There is no clear rule other than "look and feel" and a sense of what belongs together...

stefanrueger: I am more worried about having wrongly replaced a strcmp() with str_eq() in the other files of this PR (b/c that might break things).

stefanrueger: > There is no clear rule other than "look and feel" and a sense of what belongs together...

And having said that... if you feel str_casematch() "belongs" to the other functions feel free to remove that line, just push onto this commit (I don't mind).

mcuee:
This PR looks good to me.

stefanrueger:
OK, merging soon

PR #1265, opened 2023-01-04
Ensure full words are loaded, low-byte first, for ISP programming 

stefanrueger:
Addresses Issue #1263 

Although the ISP interface for classic parts loads pages byte for byte (not word for word) into a buffer on chip, the data sheets say that *the low byte must be written before the high byte for programming to work*. This PR ensures that. The problem with the current code is that perfectly legal hex files can just set high bytes from time to time (odd start address of blocks etc).

PR #1376, opened 2023-06-01
Remove pgm->page_erase() function when unable to erase page

### LINKED ISSUE #1267: To smoke out all the programmers that advertise their `page_erase()` but then fail to provide it
@stefanrueger created PR #1266 to silence potential false-alarm error messages.

* https://github.com/avrdudes/avrdude/pull/1266#issuecomment-1371315416
> That is the second-best option, though. It would be better to smoke out all the programmers that advertise their page_erase() but then fail to provide it. Anyway, that is for another day.

stefanrueger:
Fixes #1267 

AVRDUDE treats `pgm->page_erase()` as an optional function that should *only* be set if the programmer/part actually can erase pages. jtagmkII, jtag3 and stk500v2 programmers offer page_erase() functions that, in some circumstances, cannot erase pages. This PR attempts to correct this behaviour.

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -1054,7 +1054,7 @@ int avr_write_mem(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m, int 
 
       if (need_write) {
         rc = 0;
-        if (auto_erase)
+        if (auto_erase && pgm->page_erase)
           rc = pgm->page_erase(pgm, p, cm, pageaddr);
         if (rc >= 0)
           rc = pgm->paged_write(pgm, p, cm, cm->page_size, pageaddr, cm->page_size);
--- a/src/avrcache.c
+++ b/src/avrcache.c
@@ -549,7 +549,7 @@ int avr_flush_cache(const PROGRAMMER *pgm, const AVRPART *p) {
 
       for(int iwr = 0, pgno = 0, n = 0; n < cp->size; pgno++, n += cp->page_size) {
         if(cp->iscached[pgno] && memcmp(cp->copy + n, cp->cont + n, cp->page_size)) {
-          if(!chiperase && mems[i].pgerase)
+          if(!chiperase && mems[i].pgerase && pgm->page_erase)
             pgm->page_erase(pgm, p, mem, n);
           if(writeCachePage(cp, pgm, p, mem, n, 1) < 0)
             return LIBAVRDUDE_GENERAL_FAILURE;
--- a/src/jtag3.c
+++ b/src/jtag3.c
@@ -1474,6 +1474,9 @@ static void jtag3_disable(const PROGRAMMER *pgm) {
 }
 
 static void jtag3_enable(PROGRAMMER *pgm, const AVRPART *p) {
+  if(!(p->prog_modes & (PM_PDI | PM_UPDI)))
+    pgm->page_erase = NULL;
+
   return;
 }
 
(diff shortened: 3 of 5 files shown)


mcuee:
@stefanrueger 

Just wondering how to test this. I have the STK500V2 based programmer. Supposedly #1266 has already silent the warnings.
* https://github.com/avrdudes/avrdude/pull/1266

mcuee:
Test results.

1) git main without PR #1265 (not exactly revert but similar to revert): there is an error message

```
PS C:\work\avr\avrdude_test\avrdude_main> git diff
diff --git a/src/avrcache.c b/src/avrcache.c
index da40c7a7..0ddfa1ad 100644
--- a/src/avrcache.c
+++ b/src/avrcache.c
@@ -333,10 +333,10 @@ static int guessBootStart(const PROGRAMMER *pgm, const AVRPART *p) {

 // Page erase but without error messages if it does not work
 static int silent_page_erase(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m, unsigned int a) {
-  int bakverb = verbose;
-  verbose = -123;
+//  int bakverb = verbose;
+//  verbose = -123;
   int ret = pgm->page_erase? pgm->page_erase(pgm, p, m, a): -1;
-  verbose = bakverb;
+//  verbose = bakverb;

   return ret;
 }

PS C:\work\avr\avrdude_test\avrdude_bin> .\avrdude_git_no_pr1266 -c stk500v2 -P COM3 -p m2560 -t
avrdude_git_no_pr1266: AVR device initialized and ready to accept instructions
avrdude_git_no_pr1266: device signature = 0x1e9801 (probably m2560)
avrdude> dump flash 0x400 0x20
Reading | ################################################## | 100% 0.17 s
00400  00 00 00 00 ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
00410  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
avrdude> write flash 0x400 0xaa 0x55 0xaa 0x55 0xaa 0x55
Caching | ################################################## | 100% 0.02 s
avrdude> flush
avrdude_git_no_pr1266: synching cache to device ...
avrdude_git_no_pr1266 error: this function must never be called
Reading | ################################################## | 100% 93.42 s
Erasing | ################################################## | 100% 0.06 s
Writing | ################################################## | 100% 5.13 s
avrdude> dump flash 0x400 0x20
Reading | ################################################## | 100% 0.08 s
00400  aa 55 aa 55 aa 55 ff ff  ff ff ff ff ff ff ff ff  |.U.U.U..........|
00410  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
avrdude> quit

avrdude_git_no_pr1266 done.  Thank you.
```

2) git main

```

```

mcuee:
Wiring bootloader testing results.

```
(to be updated)
```

stefanrueger:
@mcuee: well done for testing this. You can also see the difference to git main in the terminal: The jtag2, jtagmkII and stk500v2 programmer variants should not offer the `pgerase` command on parts or programmer variants that cannot do a page erase. In git main, when page erase is offered for a classic part in an stk500v2 programmer, the page erase should fail with an unusual(!) message "this function must never be called".

mcuee:
I think this PR is safe to merge now. I will check the Wiring bootloader issue later. In any case, I think wiring bootloader is not popular (only used by Arduino Mega2560 board) and the Arduino bootloader FW is not that good (very old optiboot version and also old STK500v2 bootloader with EEPROM bug not fixed).

PR #1378, opened 2023-06-03
Write fuse only when terminal config needs value to change

stefanrueger:
Simple PR that saves writing to fuses/lock bits when no change is needed.


mcuee:
This should be safe to meger. It is a simple but useful improvement.

Issue #1384, opened 2023-06-05
Improve programmer documentation


stefanrueger:
AVRDUDE supports some 155 programmers, and growing, but in many cases there is little documentation about the supported programmer. This [list](https://github.com/avrdudes/avrdude/discussions/1019#discussioncomment-4146009) is a good start for collecting URLs and background info. The aim of this issue would be to, at the very least provide each programmer with an URL with some background (where to buy, how to build that programmer, its data sheet or product sheet or similar) as comment between title and programmer definition in `avrdude.conf`; I am sure the project would be happy to receive PRs improving this.

mcuee:
Reference discussion here:

* https://github.com/avrdudes/avrdude/discussions/1019

mcuee:
Early draft in #1019.

Draft: 15-Nov-2022
Programmer info URL list.

Name | Description | URL
-- | -- | --
2232HIO | FT2232H based generic programmer |  
4232h | FT4232H based generic programmer |  
adafruit_gemma | Arduino Gemma bootloader disguised as   USBtiny |  
arduino | Arduino skeleton STK500v1 compatible   bootloader |  
arduino_gemma | Adafruit Trinket Gemma bootloader disguised   as USBtiny |  
arduino-ft232r | Arduino: FT232R connected to ISP |  
arduinoisp | Arduino ISP Programmer |  
arduinoisporg | Arduino ISP Programmer |  
atmelice | Atmel-ICE (ARM/AVR) in JTAG mode |  
atmelice_dw | Atmel-ICE (ARM/AVR) in debugWIRE mode |  
atmelice_isp | Atmel-ICE (ARM/AVR) in ISP mode |  
atmelice_pdi | Atmel-ICE (ARM/AVR) in PDI mode |  
atmelice_updi | Atmel-ICE (ARM/AVR) in UPDI mode |  
avr109 | Atmel AppNote AVR109 Boot Loader |  
avr910 | Atmel Low Cost Serial Programmer |  
avr911 | Atmel AppNote AVR911 AVROSP |  
avrftdi | FT2232D based generic programmer |  
avrisp | Atmel AVR ISP |  
avrisp2 | Atmel AVR ISP mkII |  
avrispmkII | Atmel AVR ISP mkII |  
avrispv2 | Atmel AVR ISP V2 |  
buspirate | The Bus Pirate |  
buspirate_bb | The Bus Pirate (bitbang interface, supports   TPI) |  
[...]

PR #1393, opened 2023-06-10
Add more documentation for urclock programmer

stefanrueger:
It occurred to me that using the developer option `-c programmer/s` one sees relatively easily the comments from `avrdude.conf` about a programmer. This makes the config file a great location to add more documentation for particular programmers.

Partially addresses #1384 

This PR gives an example:

```
$ avrdude -c ur'*'/s
# urclock

# Serves urboot and optiboot bootloaders
#  - Reads/writes flash/EEPROM via the MCU bootloader and a serial connection
#  - Implements urprotocol, a skeleton version of STK500v1
#  - Supports vector bootloaders by patching interrupt vectors during upload:
#     + Vector bootloaders run on all parts, no need for a HW boot section
#     + Can be much smaller than the smallest HW boot section of a part, eg,
#       256 bytes for ATmega2560 (smallest HW boot section is 1024 bytes)
#  - Checks sizes of applications so they don't overwrite the bootloader
#  - Keeps the bootloader alive during interactive terminal sessions
#  - Provides a 4-byte metadata interface in top flash for
#     + Allowing applications to utilise unused flash similar to EEPROM
#     + Storing in top flash the file name and its last-modified date
#     + Displaying file name and date of the last uploaded application
#
# See https://github.com/stefanrueger/urboot

programmer
    id                     = "urclock";
    desc                   = "Urclock programmer for urboot bootloaders using urprotocol";
    type                   = "urclock";
    prog_modes             = PM_SPM;
    connection_type        = serial;
;
```


mcuee:
Looks good to me.

PR #1386, opened 2023-06-06
Fix stk500v2.c's write_byte() for old parts

stefanrueger:
A stab in the grey for Issue #433 and #509. For more details see [here](https://github.com/avrdudes/avrdude/issues/433#issuecomment-1578755276).

PR #1401, opened 2023-06-13
Fix memory leak in avrcache.c

stefanrueger:
Something I found whilst reviewing avrcache.c for PR #1386 

Not a biggie, but feel free to check the diffs for sanity.


mcuee:
Looks good to me.

Issue #1398, opened 2023-06-13
Mute missing USB device warning when programmer is present with another USB PID


MCUdude:
Some programmers, especially older Atmel ones, tend to have changed their USB PID during their lifespan.
The JTAGICE3 for instance, started out with PID 0x2110 but was later changed to 0x2140 when the never CMCIS-DSP firmware was released.

However, the way usb_hidapi.c works, Avrdude prints a warning for every possible USB PID that's not present.
Here's an example of a JTAGICE3 that's running the latest firmware.

```
$ ./avrdude -cjtag3updi -patmega4808
avrdude warning: USB device with VID: 0x03eb and PID: 0x2110 not found
avrdude: AVR device initialized and ready to accept instructions
avrdude: device signature = 0x1e9650 (probably m4808)

avrdude done.  Thank you.
```

Perhaps we should remove the "USB device with VID" warning completely, or only show it in -vv, -vvv or -vvvv mode? If no programmer is connected, we'll get a "no device found matching VID" error anyways, which is just as useful.

```
$ ./avrdude -cjtag3updi -patmega4808
avrdude warning: USB device with VID: 0x03eb and PID: 0x2110 not found
avrdude warning: USB device with VID: 0x03eb and PID: 0x2140 not found
avrdude error: no device found matching VID 0x03eb and PID list: 0x2110, 0x2140
avrdude error: unable to open programmer jtag3updi on port usb

avrdude done.  Thank you.
```

stefanrueger:
Good idea. Perhaps still print with `-vv`?

MCUdude:
> Perhaps still print with -vv?

I'm file with this! I'll create a PR this evening

PR #1403, opened 2023-06-14
Mute missing USB device warning for lower verbosity levels

MCUdude:
See #1398 for details.

Now the warning will only be displayed if the verbosity level is two or greater.

### DIFF
--- a/src/usb_hidapi.c
+++ b/src/usb_hidapi.c
@@ -123,10 +123,8 @@ static int usbhid_open(const char *port, union pinfo pinfo, union filedescriptor
     dev = hid_open(pinfo.usbinfo.vid, pinfo.usbinfo.pid, NULL);
     if (dev == NULL)
     {
-      if (verbose > 1) {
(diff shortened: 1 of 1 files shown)


stefanrueger:
Good. Why use `verbose`? It's not wrong, but `pmsg_notice()` or `pmsg_notice2()` depending on the level you'd like to set would have had a similar (though not identical) effect. But either is cool.

mcuee:
> Good. Why use `verbose`? It's not wrong, but `pmsg_notice()` or `pmsg_notice2()` depending on the level you'd like to set would have had a similar (though not identical) effect. But either is cool.

Indeed this should be a better way. I think `pmsg_notice()` is good enough,

MCUdude:
The reason for using `verbose > 1` is to keep the `avrdude warning` prefix.
But if you think it's not necessary, I can use `pmsg_notice2` instead.

stefanrueger:
All good, either way is fine. More depends on whether you think this is a running commentary on something that is expected or a genuine warning about something that is not quite right.

PR #1418, opened 2023-06-21
Update ft2232h programmer info and URLs

stefanrueger:


stefanrueger:
The comment to the ft2232h programmer says:

```
# The drivers will look for a specific device and use the first one found.
# If you have mulitple devices, then look for unique information (like SN)
# and fill that in here.
```

What does it mean to look for unique information (like SN) and fill that in here? How would a user do that?

mcuee:
> The comment to the ft2232h programmer says:
> 
> ```
> # The drivers will look for a specific device and use the first one found.
> # If you have mulitple devices, then look for unique information (like SN)
> # and fill that in here.
> ```
> 
> What does it mean to look for unique information (like SN) and fill that in here? How would a user do that?

The user will need to put the USB Serial Number in the avrdude configure file.

The code has also the place holder for using the unique descriptor or index to identify the programmers, but not implemented yet.

https://github.com/avrdudes/avrdude/blob/9269102ba5dee7706cc4432a528635293e27f05b/src/avrftdi.c#L668-L677

stefanrueger:
@mcuee Thanks. I have added the following:
```
# found. If you have multiple devices, and they give out serial
# numbers, a different entry for each of them can be created in a
# personal ~/.avrduderc or avrdude.rc entry and the usbsn = "...";
# field added to distinguish between them.
```

The `usbsn` field is used.

mcuee:
Good. Now this PR should be good to go.

PR #1422, opened 2023-06-21
Update usbasp programmer info

stefanrueger:


MCUdude:
We should try this with various USBasp firmware before merging this. IIRC, there is firmware out there that check whenever the fischl.de name is correct:

https://github.com/nerdralph/usbasp/blob/6b9b85f13d0434f8e4c0ed434c692392488e1fd0/firmware/usbconfig.h#L222

mcuee:
This PR should be okay.

I will put in the Wiki about improved firmware like the following.
1) https://github.com/mariusgreuel/USBasp (Windows OS descriptor to load WinUSB driver automatically)
2) https://github.com/nerdralph/usbasp 1)+ (features a default SCK clock of 1.5Mhz and automatic SCK slowing if target does not respond. PORTD is left as input, so this firmware also works [with USBISP modules](https://www.sciencetronics.com/greenphotons/?p=938).
3) https://github.com/dioannidis/usbasp 2)+, with extra firmware feature like HID UART, setting unique serial number and AT89S51/AT89S52 support.

stefanrueger:
> improved firmware

Made a note in avrdude.conf, too. It's something a user sees when typing `avrdude -cusbasp/s`

mcuee:
Now this PR should be good to go.

Issue #1408, opened 2023-06-20
Add `-xrtsdtr=low,high` to the jtag2updi programmer


MCUdude:
the `-xrtsdtr=low,high` extended parameter is used on the SerialUPDI programmer to control the state of the DTR/RTS line when Avrdude is communicating with the target. This is especially useful for programmers such as the [SerialUPDI programmer I sell on Tindie](https://www.tindie.com/products/mcudude/serialupdi-programmer/) because it can automatically switch between "UPDI mode" and "UART mode".

This is something that would be just as useful for programmer hardware that's based on the jtag2updi project.

(SerialUPDI schematic that utilizes the RTS line to switch between UPDI and UART mode)
![image](https://github.com/avrdudes/avrdude/assets/3238730/94621b96-2b34-44a5-be6e-9be359cf4666)

PR #1424, opened 2023-06-21
Enable -c programmers to access -c string

stefanrueger:
Enabling `-c` programmers to see the `-c` string chosen by the user is helpful to distinguish between different versions of (essentially) the same programmer.

The programmer id parameter is a list of programmer strings. Whenever code wants to refer to the name of the programmer selected by the user with `-c` it normally uses `(char *) ldata(lfirst(pgm->id))`. This is incorrect if the chosen programmer happens to be the *second* or later in a list of programmer ids. 

To reproduce, create the following two new programmers in `~/.avrduderc`
```
programmer parent "urclock"
    id                     = "urclock1", "urclock2";
;
```

Then execute
``` console
$ avrdude -c urclock2 -xhelp
avrdude -c urclock1 extended options:
  -xshowall         Show all info for connected part and exit
[...]
```

Notice the `urclock1` response whereas the user selected `urclock2`. 

This PR requires careful review as not all `(char *) ldata(lfirst(pgm->id))` can be replaced with `pgmid`. @MCUdude?

PR #1425, opened 2023-06-21
Add jtag2updi -xrtsdtr extended parameter

MCUdude:
This PR lets the user control the RTS/DTR line on the UART line when using the jtag2updi programmer. This may be useful for controlling external hardware, such as switching between UPDI programming and UART mode. See #1408 for details

### DIFF
--- a/src/doc/avrdude.texi
+++ b/src/doc/avrdude.texi
@@ -1456,10 +1456,12 @@ the MCU's ‘TXD’ line.
 Show help menu and exit.
 @end table
 
+@cindex @code{-x} jtag2updi
 @cindex @code{-x} serialupdi
+@item jtag2updi
 @item serialupdi
 
-The serialupdi programmer type accepts the following extended parameter:
+The jtag2updi and serialupdi programmer types accepts the following extended parameters:
 @table @code
 @item @samp{rtsdtr=low,high}
 Forces RTS/DTR lines to assume low or high state during the whole
--- a/src/doc/avrdude.texi
+++ b/src/doc/avrdude.texi
@@ -1461,7 +1461,7 @@ Show help menu and exit.
 @item jtag2updi
 @item serialupdi
 
-The jtag2updi and serialupdi programmer types accepts the following extended parameters:
+The jtag2updi and serialupdi programmer types accept the following extended parameters:
 @table @code
 @item @samp{rtsdtr=low,high}
 Forces RTS/DTR lines to assume low or high state during the whole
--- a/src/jtagmkII.c
+++ b/src/jtagmkII.c
@@ -1417,7 +1417,7 @@ static int jtagmkII_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) 
         if (str_caseeq(rts_mode, "low")) {
           PDATA(pgm)->rts_mode = RTS_MODE_LOW;
         } else if (str_caseeq(rts_mode, "high")) {
-          PDATA(pgm)->rts_mode = RTS_MODE_LOW;
+          PDATA(pgm)->rts_mode = RTS_MODE_HIGH;
         } else {
           pmsg_error("RTS/DTR mode must be LOW or HIGH\n");
           return -1;
--- a/src/jtagmkII.c
+++ b/src/jtagmkII.c
@@ -1220,7 +1220,7 @@ static int jtagmkII_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   const char *ifname;
 
   if (PDATA(pgm)->rts_mode != RTS_MODE_DEFAULT) {
-    pmsg_warning("forcing serial DTR/RTS handshake lines %s\n",
+    pmsg_info("forcing serial DTR/RTS handshake lines %s\n",
       PDATA(pgm)->rts_mode == RTS_MODE_LOW ? "LOW" : "HIGH");
   }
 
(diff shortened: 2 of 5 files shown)


src/jtagmkII.c
```diff
@@ -1215,6 +1219,11 @@ static int jtagmkII_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   int ok;
   const char *ifname;
 
+  if (PDATA(pgm)->rts_mode != RTS_MODE_DEFAULT) {
+    pmsg_warning("forcing serial DTR/RTS handshake lines %s\n",
```
stefanrueger, line 1223: Isn't this a running commentary `pmsg_info()` rather than a warning?

src/doc/avrdude.texi
```diff
@@ -1456,10 +1456,12 @@ the MCU's ‘TXD’ line.
 Show help menu and exit.
 @end table
 
+@cindex @code{-x} jtag2updi
 @cindex @code{-x} serialupdi
+@item jtag2updi
 @item serialupdi
 
-The serialupdi programmer type accepts the following extended parameter:
+The jtag2updi and serialupdi programmer types accepts the following extended parameters:
```
stefanrueger, line 1464: types accept (not accepts)

MCUdude: Thanks! It's these things that always get me

src/jtagmkII.c
MCUdude: I ended up using `pmsg_warning` because that's what the SerialUPDI programmer does. But it would make sense to use `pmsg_info` instead for both the jtag2updi and serialupdi programmer

stefanrueger:
Looks good though cannot test functionality.  I suspect there are going to be a number of conflicts with the previous PR #1324, which is an unfortunate coincidence, but lets cross that bridge when we come to it. I normally merge in batches but within batches in historic order (first PR gets merged first).

MCUdude:
I assume you mean PR #1424? Merge that first, and I'll rebase upon that, no worries

MCUdude:
@mcuee do you have a chance to test this PR with your hardware as well? You can use a multimeter or an oscilloscope and measure directly on the RTS or DTR line. It should stay high or low during the entire programming session (terminal mode as well), depending on the user-provided `-xrtsdtr` parameter.

mcuee:
@MCUdude 

It seems to me I do not have the right hardware to carry out the test.

Just one question about your setup of jtag2updi and target for this PR.

1) For on-board jtag2updi programmer (Arduino Nano Every, or Nano 4808, with buggy jtag2updi implementation), I think we do not have the RTS/DTR lines available on the pins.

2) For Arduino Uno (official or clone) based implementation, it seems to me DTR/RTS lines are not available for use.

Official Uno Rev 3:
https://content.arduino.cc/assets/UNO-TH_Rev3e_sch.pdf

Typical CH340x based clone
http://electronoobs.com/images/Arduino/tut_31/arduino_uno_scheamtic_ch340.png

arduino_uno_scheamtic_ch340

MCUdude:
> For on-board jtag2updi programmer (Arduino Nano Every, or Nano 4808, with buggy jtag2updi implementation), I think we do not have the RTS/DTR lines available on the pins.

> For Arduino Uno (official or clone) based implementation, it seems to me DTR/RTS lines are not available for use.

You're absolutely correct. In order to utilize this functionality, one would have to design a dedicated jtag2updi programmer and construct it similar to the way the [SerialUPDI programmer](https://www.tindie.com/products/mcudude/serialupdi-programmer) is constructed. Even though I'm not planning to sell a jtag2updi programmer that has this feature, it could be useful for future programmers. For instance, the Nano Every could have been using this feature to switch between UART and UPDI mode instead of the pesky 1200bps touch hack.

If you want to check if this works on your UNO, trace down the other side of the auto-reset capacitor (which is connected to the DTR or RTS pin) and see how it behaves.

mcuee:
> If you want to check if this works on your UNO, trace down the other side of the auto-reset capacitor (which is connected to the DTR or RTS pin) and see how it behaves.

I will test this next Monday when I have the access to the equipment.

stefanrueger:
OK, so will leave this PR unmerged when I merge the other PRs, probably this (European) evening

mcuee:
@stefanrueger and @MCUdude 

Somehow I can not test this properly on my Uno clone, as it may not use CH340G IC and the pin-out is totally different.

Anyway this PR needs to be updated to resolve the conflicts. After that, this PR can be merged after the review from @stefanrueger.

mcuee:
@MCUdude 

So I get another Arduino Clone which follows the official design (using ATmeag16U2 as the USB to Serial chip).
https://content.arduino.cc/assets/UNO-TH_Rev3e_sch.pdf

Then I see why my testing is not good, the design is using CTS plus capacitor for the reset pin. not DTR or RTS. The RTS pin is not exposed so I can not probe it.

In the end, I do not have the right HW to carry out the test.

mcuee:
Anyway this PR needs to be updated to resolve the conflicts. After that, maybe this PR can be merged if the test results from @MCUdude is positive and the review from @stefanrueger is positive.

MCUdude:
Merge conflict resolved 👍

stefanrueger:
Looks good: ready to be merged. Also probably the last merge of code before v7.2 apart from fixes arising from tests.

PR #1450, opened 2023-06-27
Update info on bsd programmer

stefanrueger:
... and update Brian Dean's email address. The wayback machine has unearthed some interesting historical details of the [AVRDUDE project in 2003](http://web.archive.org/web/20030217002616/http://www.bsdhome.com/avrdude/) when it was renamed from the [2001 AVRPROG project](http://web.archive.org/web/20010810054108/http://www.bsdhome.com/avrprog/) that apparently had its first commit on <s>2000-05-08</s> 2000-08-05.

It is rare for any computer program to survive a couple of decades and much rarer for one that essentially keeps its structure along the way!

This one really is for @dl8dtl to review :wink:


mcuee:
Looks like a good PR to merge. The typo fix and email address update are anyway good to go already.

bd34n:
Looks good to me. Thanks @stefanrueger for cleaning up those dead links and adding a bit of the history. On the comment above, I believe the date is 2000-08-05. The next commit for the Makefile, revision 1.2, was on 2000-08-06. That's from my old CVS repository. "avrprog.c" was committed at the same time, which is why the commit log comment mentions it as well. Both, along with other files, were initially checked into CVS on 2000-08-05.

```c
----------------------------
revision 1.2
date: 2000/08/06 03:53:06;  author: bsd;  state: Exp;  lines: +8 -1
Makefile: Add an install target.

avrprog.c:

  Add license.

  Document the header a bit better.

  Add capability to read out and display the device signature bytes.

  Add capability to power the device from the parallel port.

  Eliminate debug print facility.

  Provide 'avr_cmd()' function.

  When memory locations don't program, generate a newline so that the
  information is not overwritten and lost.

  Don't print out the message about needing to specify a file if the
  user is not requesting an operation that requires the file.
----------------------------
revision 1.1
date: 2000/08/05 05:12:50;  author: bsd;  state: Exp;
branches:  1.1.1;
Initial revision
----------------------------
revision 1.1.1.1
date: 2000/08/05 05:12:50;  author: bsd;  state: Exp;  lines: +0 -0
Initial check-in
=============================================================================

```

PR #1451, opened 2023-06-27
Add datasheet link comments to parts

stefanrueger:
Not sure we want this. There is a risk Microchip move the data sheets elsewhere and we have dead links. And there are a lot of parts too so it it's not clear we will be able to get good coverage. Comment welcome

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -2819,6 +2819,14 @@ programmer
 #------------------------------------------------------------
 
 # This is an HVSP-only device.
+# ATtiny 11 and 12 seem to be a predecessors of the ATtiny13:
+# https://ww1.microchip.com/downloads/en/Appnotes/doc4297.pdf
+#
+# Only the summary data-sheet appears to be availible:
+# https://ww1.microchip.com/downloads/en/DeviceDoc/1006s.pdf
+#
+# Errata:
+# http://ww1.microchip.com/downloads/en/DeviceDoc/DOC1607.PDF
 
 part
     desc                   = "ATtiny11";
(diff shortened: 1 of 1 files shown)


mcuee:
I am think this is hard to maintain and it does not help most of the people. Usually people will use the search engine for the datasheet (and then save it if necessary).

mcuee:
And for outdated parts like ATtiny11, you may not find datasheet link from Microchip with the search engines.

I just used Google to search `ATtiny11 datasheet` and the first hit is from Mouser Singapore with info for the ATtiny11. 
https://www.mouser.sg/datasheet/2/268/1006s-1315538.pdf

The second is indeed a Microchip datasheet but for the ATtiny25/45/85 and not ATtiny11. Google can not find the link you mentioned.

mcuee:
And for ATtiny12, there is a Microchip website with quite some information, except the datasheet. And it mentions: `Status: New Device Entry`, which seems to indicate they have recently resurrected the part website but forgot to attach the datasheet.
https://www.microchip.com/en-us/product/ATTINY12

mcuee:
@stefanrueger 

I think this is a good idea.

Documentation from Microchip
1) Microchip programmer documentation, including protocol and bootloader information (AN)
2) Microchip common manual (eg: instruction set manual)
3) Microchip datasheet for AVR chips

Other things
1) Microchip ATDF files (Microchip Studio)
2) Microchip xml files (older version of Atmel Studio)

MCUdude:
The only issue I see is that URLs tends to change from time to time, which may lead to a bunch of broken links when the release is a few years old. Wouldn't it be better to create a new repo where we host everything, and just link to each document in this repo instead?

mcuee:
In this case, we can close this PR and convert this to a discussion.

PR #1454, opened 2023-06-29
Suppress overzealous compiler warnings

stefanrueger:
This PR replaces `-1UL` with `~0UL` (and my heart bleeds a little).

Some compilers are overzealous when the unary minus operates on an unsigned number, and issue a warning. In fact, `-u` is a perfectly valid unsigned for any unsigned `u`, and always has been from K&R C to C++20. `-u` is the same as `~u + 1`. C/C++ unsigned numbers form a group, where `-u` is the unique unsigned number that when added to `u` results in `0`. Neither `u` nor `-u` can be negative because no unsigned number is.

### DIFF
--- a/src/stk500v2.c
+++ b/src/stk500v2.c
@@ -2725,7 +2725,7 @@ static int stk500v2_paged_write(const PROGRAMMER *pgm, const AVRPART *p, const A
   // determine which command is to be used
   if (strcmp(m->desc, "flash") == 0) {
     addrshift = 1;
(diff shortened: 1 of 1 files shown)


mcuee:
@stefanrueger 

Alternative solution is to disable the warning. I am okay with either this PR or just disable the warning.

```
#ifdef _MSC_VER
/* disable: "warning C4146: unary minus operator applied to unsigned type, result still unsigned" */
#pragma warning(disable:4146)
#endif
```

Issue #1330, opened 2023-03-28
avrdude memory type enhancement


stefanrueger:
There isn't really a good authority how the memory should be called that houses the lock bits. Neither the data sheets nor the atdf files have a systematic way of naming. AVRDUDE calls these memories `lock` unless it's a `TPI` part (t4, t5, ..., t40, t102, t104), in which case it's `lockbits`. 

*Or* and that's the surprising bit: if it's the ATmega406 then this memory is *also* called `lockbits`.

Any rhyme or reason?

dl8dtl:
No idea, and actually, I've never even held an ATmega406 in my fingers. It's a quite exotic part – even though it's still in production these days.

The datasheets of the non-Xmega AVRs call it "lock bit byte", the Xmegas call it "lock bit register".

stefanrueger:
And the ATDFs call that register variously,  `lockbit`  (241x), `lockbits` (45x, Xmegas) and (... wait for it ...) `key` (37x, Dx). 

How about calling the ATmega406's memory `lock` in `avrdude.conf.in` based on the strength of the datasheet that also calls that byte `lock bit byte`. It doesn't terribly matter, but it's kinda a bit more systematic. What actually should happen is that there ought to be an int parameter in each memory section whose bits tell us what the function of that memory is. This saves "AVRDUDE" from faffing around with quite a few `strcmp()` to figure that out, eg, `MEM_IS_FLASH` would be set for apptable, application, boot and flash, whilst the various fuses would have `MEM_IS_CONFIG` or similar. Mucho eleganto.

![Screenshot from 2023-03-28 21-01-18](https://user-images.githubusercontent.com/14282046/228354148-4f520f19-61cd-4a49-bede-18150f71c092.png)

dl8dtl:
> How about calling the ATmega406's memory `lock`

I am always hesitant to change things that have alway^H^H^H^H^Hfor a long time been the way they are. You never know whom you might shoot into the foot with renaming things. Frankly, I don't even have the slightest idea whether there's even a single ATmega406 AVRDUDE user – but I am sure, if you gratuitously rename things (even if for a good reason), you might suddenly learn how many active users of it have already been out there. :-)

That's opensource life: if something works, you'll (almost) never hear about it from the users.

stefanrueger:
> for a long time been the way they are

Fair point as to external representation.

[...]

PR #1482, opened 2023-08-06
Replace string comparison functions

MCUdude:
This PR replaces the `strcmp`/`strncmp`/`strcasecmp`/`strncmp` string comparison functions with Avrdude's own `str_eq`/`str_starts`/`str_caseeq` etc.

I've so far only modified `jtag3.c`, `stk500v2.c`, `jtagmkii.c` and `jtagmki.c`, but I can go through the entire code base if that's OK. When grepping for `strcmp` or `strncmp`, I realize there aren't that many occurrences left in the codebase.

I've also slightly simplified the way jtag3/jtagmkii/jtagmki/stk500v2 deals with the various fuses.

### DIFF
--- a/src/avr910.c
+++ b/src/avr910.c
@@ -403,7 +403,7 @@ static int avr910_write_byte(const PROGRAMMER *pgm, const AVRPART *p, const AVRM
 {
   char cmd[2];
 
-  if (strcmp(m->desc, "flash") == 0) {
+  if (str_eq(m->desc, "flash")) {
     if (addr & 0x01) {
       cmd[0] = 'C';             /* Write Program Mem high byte */
     }
@@ -413,7 +413,7 @@ static int avr910_write_byte(const PROGRAMMER *pgm, const AVRPART *p, const AVRM
 
     addr >>= 1;
   }
-  else if (strcmp(m->desc, "eeprom") == 0) {
+  else if (str_eq(m->desc, "eeprom")) {
     cmd[0] = 'D';
   }
   else {
@@ -468,11 +468,11 @@ static int avr910_read_byte_eeprom(const PROGRAMMER *pgm, const AVRPART *p, cons
 static int avr910_read_byte(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m,
                             unsigned long addr, unsigned char * value)
 {
-  if (strcmp(m->desc, "flash") == 0) {
+  if (str_eq(m->desc, "flash")) {
     return avr910_read_byte_flash(pgm, p, m, addr, value);
   }
 
-  if (strcmp(m->desc, "eeprom") == 0) {
+  if (str_eq(m->desc, "eeprom")) {
     return avr910_read_byte_eeprom(pgm, p, m, addr, value);
   }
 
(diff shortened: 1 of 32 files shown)


stefanrueger:
Thanks for the PR, @MCUdude. I have been meaning to replace most string comparisons with memories with testing bits in a new `mem_type` descriptor, see https://github.com/avrdudes/avrdude/issues/1330#issuecomment-1487579845 So, as an advance warning, some of the work that this PR does might be undone at some point if I get round to deal with #1330

MCUdude:
That's totally fine @stefanrueger! Several other string comparisons are still done "the old way" that could be replaced. If you think you'll accept the PR once I'm finished, I can continue replacing the rest of the string comparisons in the other files as well, and deal with irregular formatting if I come across it.

stefanrueger:
Thanks @MCUdude. Yes, the PR looks fine. I might have chosen a less permissive variant instead of `str_contains(m->desc, "fuse")` sth along the lines of
```
  strlen(m->desc) <= 5 && (str_ends(m->desc, "fuse") || str_begins(m->desc, "fuse"))
```
that entertains the likes of "fuse", "fuse0" and "lfuse" but it's also OK as is. Unlikely Micochip will invent a memory `pwroffuse` or similar that contains "fuse" but isn't one.

MCUdude:
> I might have chosen a less permissive variant instead of `str_contains(m->desc, "fuse")` sth along the lines of

Good idea! I ended up using the following code instead, that's a little more compact but still does the same thing:
```c
else if (str_contains(mem->desc, "fuse") && strlen(mem->desc) <= 5) 
```

I've updated the PR and replaced (almost) all strcmp/strncmp/strstr function calls. There are still a few left that I didn't dare to touch in case I screwed up something.

stefanrueger:
> a little more compact but still does the same thing

Well spotted!

Best to leave `avrintel.c` alone: it's written by a program that creates the relevant tables from .atdf files and other sources of knowledge incl errata.

stefanrueger:
> There are still a few [strcmp] left

Yes, some are actually needed to facilitate sorting, ie, provide an int that can be negative, postitive or zero.

MCUdude:
> Best to leave avrintel.c alone: it's written by a program that creates the relevant tables from .atdf files and other sources of knowledge incl errata.

Yes, I realized that shortly after pushing the commit. Fixed!
BTW, would it be a good idea to upload the `mkavrintel.pl` script into the `tools` folder in this repo? Using that script is the only sensible way to update the avrintel.c/h files.

> Yes, some are actually needed to facilitate sorting, ie, provide an int that can be negative, postitive or zero.

I've tried to be as careful as I can and left strcmp/strncmp where ever non-boolean value (or a returned pointer) is needed.

stefanrueger:
> upload the mkavrintel.pl script into the tools folder

It's a complex, grown, haphazard project on its own with a lot of source files (incl avrlib, .atdf files) hand-curated errata lists in, ahem, perl that famously is one of the few languages that is easier to write than to read. `mkavrintel.c` relies on the output of other meta-programmes. The world is a better place with this program hidden away. It's better for my streed-cred as a programmer, too. ;)

Issue #907, opened 2022-03-21
Support COM port discovery via USB VID/PID


MCUdude:
Apparently, @mariusgreuel's Avrdude fork for windows supports COM port discovery via USB VID/PID. This is actually very neat if you're using a UART-based programmer that has a tendency to bump the COM port number. I have a similar issue on my mac as well; The serial port (/dev/cu.*) gets a different name depending on which USB port I connect it to.

Would it be possible, within a reasonable amount of time, to make this work on non-windows systems as well?

mariusgreuel:
The crucial part is figuring out the link between the USB device and the serial device. I never looked into non-Windows implementations, but on Windows there are two parts of metadata one can use: 1) You can enumerate through the device tree, and find the parent of the serial device, which will be the USB device. That is tricky if the USB device is a composite device. 2) Windows actually maintains a bunch of properties for every device, and [Portname](https://github.com/mariusgreuel/avrdude/blob/91be5272654e062fc5bf1b2e69deb5544e33520f/src/msvc/usb_com_locator.h#L624) does what I needed.

dl8dtl:
For a completely different example, on FreeBSD, they can only be examined by `sysctl`.
Here's the example of the tty port of a curiosity nano board:
```
# sysctl dev.umodem
dev.umodem.0.ttyports: 1
dev.umodem.0.ttyname: U3
dev.umodem.0.%parent: uhub12
dev.umodem.0.%pnpinfo: vendor=0x03eb product=0x2175 devclass=0xef devsubclass=0x02 devproto=0x01 sernum="MCHP3372031800002359" release=0x0100 mode=host intclass=0x02 intsubclass=0x02 intprotocol=0x01 ttyname=U3 ttyports=1
dev.umodem.0.%location: bus=4 hubaddr=4 port=4 devaddr=5 interface=1 ugen=ugen4.5
dev.umodem.0.%driver: umodem
dev.umodem.0.%desc: Microchip Technology Incorporated nEDBG CMSIS-DAP, class 239/2, rev 2.00/1.00, addr 5
dev.umodem.%parent: 
```
These attach to the `umodem` driver (aka. CDC device), and the respective device here is `/dev/cuaU3`. Alas, one would have to walk across all possible drivers in addition to `umodem` (FTDI, PL230x, CH3xx).

So question is to find out how to get this information on MacOS.

Downside: it adds a lot of per-OS code to AVRDUDE.

MCUdude:
[...]

PR #1497, opened 2023-08-28
Provide serialadapter entity in avrdude.conf

stefanrueger:
This work-in-progress PR provides a `serialadapter` entity in `avrdude.conf.in` as a specific programmer, namely one that only defines USB parameters. In contrast to the [ser-auto branch](https://github.com/MCUdude/avrdude/tree/ser-auto) this PR intends to
 - Allow inheritance of `serialadapter` entities, eg, you can have in your `~/.avrduderc` file sth like
   ```
   serialadapter parent "ch340"
       id                     = "mysuperduperboard";
       usbsn                  = "08/15";
   ;
   ```
 - Allow any programmer with usbpid definitions to be used as a serial adapter, eg, the `ft232r` programmer is also a serial adapter
 - Have `usbpid` as a list instead of a single value
 - Make all functions in `serialadapter.c` of the [ser-auto branch](https://github.com/MCUdude/avrdude/tree/ser-auto) superfluous as `serialadapter` *is* a programmer, so the original programmer functions will do the job
 - Print `serialadapter` entities with developer options `-c*/s`

This PR also provides a `SERIALADAPTER` structure: actually, behind the scenes, it is the same as `PROGRAMMER`, but this will allow writing self-documenting code.

For the time being, this PR utilises the `baudrate` component of `PROGRAMMER` as `default_baud` but it would be vvv simple to have both `baudrate` and `default_baud` if needed. Here is a code example of how this PR can be utilised to retrieve a particular serial adapter (or programmer with USB definitions) from the config files:

```
  SERIALADAPTER *sea = locate_programmer(programmers, sea_id);
  if (sea == NULL || !is_serialadapter(sea)) {
    serialadapter_not_found(sea_id);
    exit(1);
  }
```

@MCUdude: Please check out this PR and see whether this is useful for what you try to achieve. I suspect @dl8dtl has reservations about viewing a serial adapter as a special unfinished programmer entry, but I am hoping that the simplicity of this PR will be somewhat persuasive.

It's possible to carry on with the [ser-auto branch](https://github.com/MCUdude/avrdude/tree/ser-auto) but adding above functionality will cost quite some work, in particular, the developer options that print AVRDUDE's understanding of its internal entities.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -214,6 +214,7 @@ add_library(libavrdude
     ser_avrdoper.c
     ser_posix.c
     ser_win32.c
+    serialadapter.c
     serialupdi.c
     serialupdi.h
     solaris_ecpp.h
@@ -286,6 +287,7 @@ target_link_libraries(libavrdude
     ${LIB_LIBFTDI}
     ${LIB_LIBFTDI1}
     ${LIB_LIBREADLINE}
+    ${LIB_LIBSERIALPORT}
     ${LIB_NCURSES}
     ${LIB_LIBGPIOD}
     ${EXTRA_WINDOWS_LIBRARIES}
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -163,6 +163,7 @@ libavrdude_a_SOURCES = \
 	ser_avrdoper.c \
 	ser_posix.c \
 	ser_win32.c \
+	serialadapter.c \
 	solaris_ecpp.h \
 	stk500.c \
 	stk500.h \
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -4,11 +4,11 @@
 #
 # This file contains configuration data used by AVRDUDE which describes
 # the programming hardware pinouts and also provides part definitions.
-# AVRDUDE's "-C" command line option specifies the location of the
-# configuration file.  The "-c" option names the programmer configuration
-# which must match one of the entry's "id" parameter.  The "-p" option
+# AVRDUDE's -C command line option specifies the location of the
+# configuration file.  The -c option names the programmer configuration
+# which must match one of the entry's id parameter.  The -p option
 # identifies which part AVRDUDE is going to be programming and must match
-# one of the parts' "id" parameter.
+# one of the parts' id parameters.
 #
 # DO NOT MODIFY THIS FILE.  Modifications will be overwritten the next
 # time a "make install" is run.  For user-specific additions, use the
@@ -21,12 +21,12 @@
 #       id       = <id1> [, <id2> ... ] ;         # <idN> are quoted strings
 #       desc     = <description> ;                # quoted string
 #       type     = <type>;                        # programmer type, quoted string
-#                                                 # supported types can be listed by "-c ?type"
+#                                                 # list known types with -c ?type
 #       prog_modes = PM_<i/f> {| PM_<i/f>}        # interfaces, eg, PM_SPM|PM_PDI (1)
-#       extra_features = HAS_<fea> {| HAS_<fea>}  # extra features, eg, HAS_SUFFER|HAS_VTARG_ADJ (2)
+#       extra_features = HAS_<fea> {| HAS_<fea>}  # extra features, eg, HAS_SUFFER (2)
 #       connection_type = parallel | serial | usb | spi
 #       baudrate = <num> ;                        # baudrate for avr910-programmer
-#       vcc      = <pin1> [, <pin2> ... ] ;       # pin number(s)
+#       vcc      = <pin1> [, <pin2> ... ] ;       # pin number(s) (3)
 #       buff     = <pin1> [, <pin2> ... ] ;       # pin number(s)
 #       reset    = <pin> ;                        # pin number
 #       sck      = <pin> ;                        # pin number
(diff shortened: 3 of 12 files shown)


MCUdude:
Thanks for the PR @stefanrueger! This approach looks much less complicated than mine, and I agree that the small tradeoffs is weighed up by the fact that the code is easier to understand, and we can utilize the developer options.

> both baudrate and default_baud if needed

> Allow inheritance of serialadapter entities, eg, you can have in your ~/.avrduderc file sth like


Inheritance is really neat. And I'm absolutely fine with `baudrate`. And I'd argue that it would make more sense to have a user-specified default baud rate in avrduderc rather than in avrdude.conf.

```
serialadapter parent "ch340"
    id                     = "mysuperduperboard";
    usbsn                  = "08/15";
    baudrate               = 250000;
;
```

> Allow any programmer with usbpid definitions to be used as a serial adapter, eg, the ft232r programmer is also a serial adapter

I didn't think of this. Neat!

> It's possible to carry on with the [ser-auto branch](https://github.com/MCUdude/avrdude/tree/ser-auto) but adding above functionality will cost quite some work, in particular, the developer options that print AVRDUDE's understanding of its internal entities.

IMO there is no doubt. We should continue working on this rather than my branch!

MCUdude:
@stefanrueger where do we continue from here? From what I can read from the code, what's missing now is the libserialport part, where it's given a USB vid, pid, and an optional serial number, and a COM port number/dev path is returned.

[The libserialport example code I wrote earlier this summer](https://github.com/avrdudes/avrdude/issues/907#issuecomment-1613575301) contains the essence of what we want to use the libserialport library for. What the example currently doesn't support is `[vid]:[pid]` or `[vid]:[pid]:[sn]` as its second argument.

MCUdude:
Perhaps we can do some basic regression testing against this PR, and then merge it as it is? I prefer to have it merged into main before continuing the work, to make my git life easier.

dl8dtl:
> I suspect @dl8dtl has reservations about viewing a serial adapter as a special unfinished programmer entry, but I am hoping that the simplicity of this PR will be somewhat persuasive.

Well, yes, I think viewing a serialadapter as just another programmer is mildly confusing. What convinces is the quality of the patch though :), in particular all the documentation updates.
I'm not sure whether there's any much value in being able to consider any programmer with a PID as a serialadapter, but maybe I'm too much used to the old Atmel tools where that really doesn't make any sense.
Regarding the future implementation, if possible, it would make sense to handle the string "usb" as a kind of pre-defined serialadapter. That would allow for unifying the current syntax that allows for e.g. `… -p jtag2 -P usb:1234` (trailing serial number digits) into a generic `… -p something -P serialadapter:serno` one, regardless of the actual programmer type. So far, I believe that `-P usb` is always handled per-programmer.

stefanrueger:
>  value in being able to consider any programmer with a PID as a serialadapter

There are a number of existing programmers, eg, ft232r, ft232h, ft2232h etc that you might want as serial adapters and that already have all the relevant info... There will be other programmers that are not serial adapters (by any stretch of imagination) but then allowing them is no harm as it really means I have usbvid and usbpid and want to be looked up as serial interface.

stefanrueger:
@MCUdude @dl8dtl There is also the possibility to add either sth like `IS_SERIALADAPTER` to `extra_features` or create a boolean `is_serialadapter` in the programmer syntax to explicitly say which of the existing programmers double up as serial adapters, for use with the likes of `ft232r`. That would be an easy addition to the current state.

MCUdude:
>  There is also the possibility to add either sth like IS_SERIALADAPTER to extra_features or create a boolean is_serialadapter in the programmer syntax to explicitly say which of the existing programmers double up as serial adapters, for use with the likes of ft232r. 

That sounds like a good idea. Then one can look at the programmers in avrdude.conf and figure out that this `ft232rl` programmer can also serve as a generic serial adapter.

stefanrueger:
@MCUdude Which programmers should this PR also treat as generic serial adapters? I assume `ft232r`, `ft2232h`, `2232hio`, `ft4232h`, `ft232h`, `c232hm`, `ft245r`,`ttl232r` and `ch341a`. Some will need usbvid/usbpid entries. What else?

However, the derived `avrftdi`, `tigard`, `avrisp-u`, `ft2232h_jtag`, `ft232h_jtag`, `um232h`, `tumpa`, etc probably not?

Note there is no entity called `ft232rl`.

MCUdude:
I tried adding the following code to main.c, just to see if I could make it print out the information from avrdude.conf. For some reason it segfaults if -P isn't `ch340`.

```c
  const char *seradapter;
  SERIALADAPTER *ser = locate_programmer_set(programmers, port, &seradapter);
  if (is_serialadapter(ser))
    msg_info("port id: %s, vid: 0x%04x, pid: 0x%04x\n", (const char *)ldata(lfirst(ser->id)), ser->usbvid, *(int *)ldata(lfirst(ser->usbpid)));
  else if (is_programmer(ser))
    msg_error("programmer incorrectly specified using -P %s\n", port);
  else if (port && !seradapter)
    msg_info("Port %s specified\n", port);
```

But why does the above code fail when I use something else than `-P ch340`? Is it `is_programmer` and `is_serialport` that don't properly handle null pointers?

stefanrueger:
It's a philosophical question of whether the caller or callee is supposed to check against NULL. Before using a function it's best to read its man page or, failing that, the function itself.

> guarding against ser being NULL. But I don't see why this would be necessary 

Try passing NULL to strlen(), str_eq(), fprintf(), ... and watch the world explode. More often than not, C-functions assume it's the caller who does the checking. Makes sense, too. You open a file once and then use the file pointer one million times to print something. So, it's expected that the caller checks *once* after opening and not burden the callee 1 million times with unnecessary checks against NULL.

But yes, is_programmer() and is_serialadapter() could be hardened against NULL.

stefanrueger:
OK, implemented `is_serialadapter = <yes|no>` component. Now

```
Valid serial adapters are:
  ch340             = [usbvid 0x1a86, usbpid 0x7523]
  ch341a            = [usbvid 0x1a86, usbpid 0x5512]
  ft2232h           = [usbvid 0x0403, usbpid 0x6010]
  ft232h            = [usbvid 0x0403, usbpid 0x6014]
  ft232r            = [usbvid 0x0403, usbpid 0x6001]
  ft4232h           = [usbvid 0x0403, usbpid 0x6011]
```

MCUdude:
> implemented is_serialadapter = <yes|no> component.

Excellent! I think this PR is ready to be merged then

PR #1512, opened 2023-10-10
libserialport github action

mcuee:
This pull request will add libserialport into Linux, macOS and MinGW github action build


mcuee:
Need to use "Squash and merge".

mcuee:
I think this should be safe to merge.

Issue #1514, opened 2023-10-12
pgm_led not activated for flash write or eeprom write - only for chip erase


Ho-Ro:
During preparing a patch for my ScratchMonkey programmer (based on Arduino Nano) that can handle its 4 status LEDs `rdy_led`, `err_led`, `pgm_led` and `vfy_led` I detected that the `pgm_led` is only activated during chip erase, but not during the flash or eeprom write. This is caused by missing calls to `pgm->pgm_led()` from `avrdude`.

My working ScratchMonkey implementation is located in my branch [ScrMon](https://github.com/Ho-Ro/avrdude/tree/ScrMon).

**How to detect this issue**
For debugging the pgm_led behaviour I've created the branch [pgm_led](https://github.com/Ho-Ro/avrdude/tree/pgm_led) that just prints the led status changes to `stderr`.

**Flash programming**
```sh
$ avrdude_Ho-Ro -pt85 -U flash:w:t85_flash.hex -q -c scratchmonkey
LED RDY ON
avrdude_Ho-Ro: AVR device initialized and ready to accept instructions
avrdude_Ho-Ro: device signature = 0x1e930b (probably t85)
avrdude_Ho-Ro: Note: flash memory has been specified, an erase cycle will be performed.
               To disable this feature, specify the -D option.
avrdude_Ho-Ro: erasing chip
LED PGM ON
LED PGM OFF

avrdude_Ho-Ro: processing -U flash:w:t85_flash.hex:i
avrdude_Ho-Ro: reading input file t85_flash.hex for flash
               with 8168 bytes in 1 section within [0, 0x1fe7]
               using 128 pages and 24 pad bytes
avrdude_Ho-Ro: writing 8168 bytes flash ...
avrdude_Ho-Ro: 8168 bytes of flash written
LED VFY ON
avrdude_Ho-Ro: verifying flash memory against t85_flash.hex
avrdude_Ho-Ro: 8168 bytes of flash verified
LED VFY OFF
LED RDY OFF

avrdude_Ho-Ro done.  Thank you.
```

**EEPROM programming**
```sh
$ avrdude_Ho-Ro -pt85 -U eeprom:w:t85_eeprom.hex -q -c scratchmonkey
LED RDY ON
avrdude_Ho-Ro: AVR device initialized and ready to accept instructions
avrdude_Ho-Ro: device signature = 0x1e930b (probably t85)

avrdude_Ho-Ro: processing -U eeprom:w:t85_eeprom.hex:i
[...]

PR #1515, opened 2023-10-12
make scratchmonkey an own programmer type that can use status LEDs

Ho-Ro:
All 4 status LEDs can be activated during programming and verifying process. Due to a probable issue #1514 from `avrdude` the `pgm_led` is only active during chip erase, but not during flash or eeprom write.


stefanrueger:
@Ho-Ro Martin, thanks for the PR!

I have had a quick look, and think we need changes
 - Changes to `.gitignore` should not be part of this PR; if you want to change .gitignore I suggest a separate PR
 - Formatting changes are discouraged unless the lines are touched anyway to keep PRs crisp, clean and focussed
 - Introducing a new programmer type should only happen when a new *programming paradigm* is used; this PR introduces *three* new types, but all that happens is a (minor) extension of STK500v2 to support LEDs; in order to avoid a proliferation of programmer types I suggest not introducing a new programmer type
 - Static variables are discouraged as in the long run we want `libavrdude` to become a re-entrant library; I realise you need to keep the state of the LEDS (b/c the scratchmonkey LED command expects the status of all LEDs) but that should happen in allocated private programmer data *per programmer*; that way if an application (different from AVRDUDE) wants to serve three connected programmers two of which are scratchmonkeys that application can  use `libavrdude` and open two scratchmonkeys and have two *separate* LED state variables

I have no doubt that this PR works for you, but I think it needs a major rewrite so it fits in with the AVRDUDE philosophy. Let me think about the best way to add LED functions to STK500v2 during runtime once the programmer has identified itself as scratchmonkey.

Ho-Ro:
> @Ho-Ro Martin, thanks for the PR!
I have had a quick look, and think we need changes

That was the reason I didn't wanted to present the PR unless asked by @mcuee in #1514

> Looks like an avrdude bug to me. Can you create a PR to address this?

So I just provided my private SMo development branch, even if I knew that it is not mature enough for production. It is just a proof of concept working on my desk with my nano equipped with 4 LEDs runing SMo software. And when the LEDs are in place they should light accordingly. :)

@stefanrueger Thx for your fast response - I will check immediately.

Ho-Ro:
Closed due to #1517

stefanrueger:
@Ho-Ro Thanks for reaching out and for providing a PR that explains in C rather than prose how AVRDUDE might be improved. So that has been useful. Yes, `extra_features` would probably have worked as well (didn't think of this).

PR #1539, opened 2023-10-28
To support MinGW cross compiler

mcuee:
This is the same as avrdude-packing patch.
https://github.com/arduino/avrdude-packing/blob/main/patches/0007-Cmake-fix-build-not-working-with-mingw-toolchain.patch


mcuee:
This should be safe to merge.

PR #1540, opened 2023-10-28
Add Arduino avrdude_packing github action for git man

mcuee:
I think this needs some review and changes to be correct.

This is a modified version of the following github action script from Arduino avrdude-packing project.
https://github.com/arduino/avrdude-packing/blob/main/.github/workflows/release.yml

I have used it for a while in my fork here.
https://github.com/mcuee/avrdude-packing

With the merge of PR #1539, no extra patches are needed to use the above script.

### DIFF
--- a/.github/workflows/arduino_packing.yml
+++ b/.github/workflows/arduino_packing.yml
@@ -1,4 +1,4 @@
-name: release
+name: avrdude_packing
 
 env:
   # The name of the project
@@ -8,7 +8,6 @@ env:
 
 on:
   push:
-  pull_request:
 
 jobs:
   build:
--- a/.github/workflows/arduino_packing.yml
+++ b/.github/workflows/arduino_packing.yml
@@ -7,7 +7,12 @@ env:
   ARTIFACT_NAME: dist
 
 on:
+  label:
+    types:
+      - created
   push:
+    branches:
+      - main
 
 jobs:
   build:
--- a/.github/workflows/arduino_packing.yml
+++ b/.github/workflows/arduino_packing.yml
@@ -44,12 +44,6 @@ jobs:
       image: ghcr.io/arduino/crossbuild:0.2.2
     
     steps:
-    
-    # this repo should contain only the patches that could not be upstreamed and the release CI nothing else
-    - name: Checkout avrdude-packing repository
-      uses: actions/checkout@v3
-      with:
-        path: avrdude-packing
 
     - name: Checkout avrdude repository
       uses: actions/checkout@v3
--- a/.github/workflows/arduino_packing.yml
+++ b/.github/workflows/arduino_packing.yml
@@ -1,0 +1,94 @@
+name: release
+
+env:
+  # The name of the project
+  PROJECT_NAME: avrdude
+  DIST_DIR: dist
+  ARTIFACT_NAME: dist
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    name: build (${{ matrix.config.os }}, ${{ matrix.config.arch }})
+    runs-on:
+      ubuntu-latest
+    strategy:
+      matrix:
+        config:
+          - os: Linux
+            arch: 64bit
+            cross_compile: x86_64-ubuntu16.04-linux-gnu
+          - os: Linux
+            arch: 32bit
+            cross_compile: i686-ubuntu16.04-linux-gnu
+          - os: Linux
+            arch: ARMv6
+            cross_compile: arm-linux-gnueabihf
+          - os: Linux
+            arch: ARM64
+            cross_compile: aarch64-linux-gnu
+          - os: macOS
+            arch: 64bit
+            cross_compile: x86_64-apple-darwin13
+            cross_compiler: o64-clang
+            ar: /opt/osxcross/target/bin/x86_64-apple-darwin13-ar # we have to manually specify the full path otherwise it's not found for some reason
+            ld: /opt/osxcross/target/bin/x86_64-apple-darwin13-ld
+          - os: Windows
+            arch: 32bit
+            cross_compile: i686-w64-mingw32
+            extension: .exe
+            
+    container:
+      image: ghcr.io/arduino/crossbuild:0.2.2
+    
+    steps:
+    
+    # this repo should contain only the patches that could not be upstreamed and the release CI nothing else
+    - name: Checkout avrdude-packing repository
+      uses: actions/checkout@v3
+      with:
+        path: avrdude-packing
+
+    - name: Checkout avrdude repository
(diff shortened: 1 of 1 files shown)


mcuee:
@umbynos

I am not so sure if you like to see this one in the main avrdude project or not. Please comment. Thanks.

mcuee:
@MCUdude, @stefanrueger and @dl8dtl 

The good thing about this is that we can provide static-linked avrdude binary for Linux, macOS. But I am not so sure if we want to support this or not.

mcuee:
It seems to be correct now.

The reason to run only on git push to git main and not for PR is that this build is a bit slow (using containers) compared to normal run.

And the script is anyway not correct for PR.

mcuee:
BTW, no rush for this PR. It is not essential as of now.

MCUdude:
If we can release statically linked binaries with every Avrdude release, that's a good thing for those who bundle Avrdude with other things, like with the Arduino cores I'm maintaining. Previously I had to wait three month for Arduino to release their 7.2 statically built binaries.

However, we should recommend users to install though their package manager or build from source if possible.

stefanrueger:
I have no real opinion on this and am happy to merge this provided we have the expertise in the team to debug problems of packaging.

mcuee:
> If we can release statically linked binaries with every Avrdude release, that's a good thing for those who bundle Avrdude with other things, like with the Arduino cores I'm maintaining. Previously I had to wait three month for Arduino to release their 7.2 statically built binaries.

With this PR, we will be able to generate statically linked binaries with every push to git main. 

> However, we should recommend users to install though their package manager or build from source if possible.

Yes I agree.

mcuee:
> I have no real opinion on this and am happy to merge this provided we have the expertise in the team to debug problems of packaging.

Yes that is the worry. I do not have the expertise myself.

On the other hand, the code is pretty much the same as the following file maintained by @umbynos (just some minor simplification). 
https://github.com/arduino/avrdude-packing/blob/main/.github/workflows/release.yml

 The script depends on the container by Arduino crossbuild project, again maintained by @umbrynos.
https://github.com/arduino/crossbuild/pkgs/container/crossbuild

That is why I'd really like to hear the opinion from @@umbrynos.

mcuee:
@umbynos

Please help here. Thanks.
1) Please state your opinion to see if it is good to have this PR in avrude or not.
2) If yes, please also look at the script (basically your script with minor modification) to see if it is correct or not and if it can be simplified or improved. I have tested this myself and it seems to work. But I think it can be simplified since it is now part of the avrdude git main repo.
2) If no, I think I may have to drop this PR.

.github/workflows/arduino_packing.yml
```diff
@@ -0,0 +1,93 @@
+name: avrdude_packing
+
+env:
+  # The name of the project
+  PROJECT_NAME: avrdude
+  DIST_DIR: dist
+  ARTIFACT_NAME: dist
+
+on:
+  push:
```
umbynos, line 10: here you can specify a branch that will trigger the action (main in your case)

.github/workflows/arduino_packing.yml
```diff
@@ -0,0 +1,93 @@
...
+    container:
+      image: ghcr.io/arduino/crossbuild:0.2.2
+    
+    steps:
+    
+    # this repo should contain only the patches that could not be upstreamed and the release CI nothing else
+    - name: Checkout avrdude-packing repository
+      uses: actions/checkout@v3
+      with:
+        path: avrdude-packing
```
umbynos, line 52: I don't think you need this, since the patches are not needed
```suggestion
```

umbynos (commented):
Hello,

Firstly, it would be an honor to see some of my work in the main avrdude repo. 
Overall IMHO looks good, I've made some minor suggestions.

> Previously I had to wait three month for Arduino to release their 7.2 statically built binaries.

I would like to be more responsive on this, but I'm not the one deciding priorities in Arduino... Integrating this would mean no waiting is required :wink: 

The only "problem" I see is this one:
> The script depends on the container by Arduino crossbuild project, again maintained by @umbynos. We will rely on the improvement of the container to add more functionalites sometimes (but not often, unless we add more dependancies like libserialport in the future).

Of course improvements will be made, but I cannot make promises on the timeline... Of course libraries can be installed/compiled insiede the container at runtime, so I don't see that as a big problem. Overall I see this situation as an improvement

.github/workflows/arduino_packing.yml
mcuee: Done. I also added the label part. Please help to check. Thanks.

MCUdude:
> I would like to be more responsive on this, but I'm not the one deciding priorities in Arduino... Integrating this would mean no waiting is required 😉

@umbynos sorry, that was not meant to sound like critique! I'm very grateful that you found some time to release 7.2 as statically linked binaries. I don't _expect_ you or Arduino to always have time to spare when a new version of Avrdude gets released, but it has been very helpful in the past.  But as you pointed out, if this PR gets merged, we (and other 3rd parties) can just use official statically linked binaries instead.

mcuee:
Just to play safe, I have created a temporary branch here to test this out. It seems to work.
https://github.com/avrdudes/avrdude/blob/test_pr1540/.github/workflows/arduino_packing.yml

I use the test branch to test and that is why it is upon push to the test_pr1547 branch.
You can see github action here.
https://github.com/avrdudes/avrdude/actions/runs/6692557883

@umbynos 
Please carry out one more review for my latest changes. Thanks a lot for the help!

mcuee:
> Of course libraries can be installed/compiled insiede the container at runtime, so I don't see that as a big problem. 

Good point. Thanks.

mcuee:
@stefanrueger 

I think this PR is ready to be merged. But we can wait for your next merge window.

It is probably better to use "squash and merge" when you merge this PR. Thanks.

MCUdude:
> I think this PR is ready to be merged. But we can wait for your next merge window.

I'm all for it, as it makes my job of maintaining 3rd part Arduino cores much easier when I want to bundle the latest Arduino version. However, we should give @dl8dtl time to share his thoughts as well.

mcuee:
> > I think this PR is ready to be merged. But we can wait for your next merge window.
> 
> I'm all for it, as it makes my job of maintaining 3rd part Arduino cores much easier when I want to bundle the latest Arduino version. However, we should give @dl8dtl time to share his thoughts as well.

That is a good point.

@dl8dtl 
Please share your thoughts. Thanks.

mcuee:
Just another test to confirm that it is working, using the test_pr1540 branch (which will be deleted after this PR gets merged).
https://github.com/avrdudes/avrdude/actions/runs/6715032983

PR #1545, opened 2023-10-29
Add Avrdude version number to avrdude.conf using autoconf

MCUdude:
Resolves #1475.

@stefanrueger I had to comment out `avrdude_version`, because the grammar complaints:

```
$ ./avrdude -cpkobn_updi -pavr16eb32
avrdude error: syntax error [/Users/hans/Downloads/avrdude/src/avrdude.conf:435]
avrdude error: unable to process system wide configuration file /Users/hans/Downloads/avrdude/src/avrdude.conf
```

It would be great if you could push a tiny grammar change to this PR so that this line doesn't have to be commented out. And maybe it can be utilized by Avrdude as well? Potentially warn the user that they're using an old/untested version of avrdude.conf?

PR #1547, opened 2023-10-30
Add avrdude_conf_version to grammar

stefanrueger:
As [requested](https://github.com/avrdudes/avrdude/pull/1545#issue-1967132995)

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -2,13 +2,17 @@
 #
 # AVRDUDE Configuration File
 #
-# This file contains configuration data used by AVRDUDE which describes
-# the programming hardware pinouts and also provides part definitions.
+
+avrdude_conf_version = "@AVRDUDE_FULL_VERSION@";
+
+# This file contains AVRDUDE's default configuration data describing
+# programming hardware pinouts and part definitions. Normally, it
+# requires AVRDUDE be of the same or higher version than above string.
 # AVRDUDE's -C command line option specifies the location of the
-# configuration file.  The -c option names the programmer configuration
-# which must match one of the entry's id parameter.  The -p option
-# identifies which part AVRDUDE is going to be programming and must match
-# one of the parts' id parameters.
+# configuration file. The -c option names the programmer configuration
+# which must match one of the entry's id parameter. The -p option
+# identifies which part AVRDUDE is going to be programming and must
+# match one of the parts' id parameters.
 #
 # DO NOT MODIFY THIS FILE.  Modifications will be overwritten the next
 # time a "make install" is run.  For user-specific additions, use the
@@ -431,9 +435,6 @@
 # ATmega16      0x74
 # ATmega8       0x76
 # ATmega169     0x78
-
-# Buildtime configuration
-avrdude_version = "@AVRDUDE_FULL_VERSION@";
 
 #
 # Overall avrdude defaults; suitable for ~/.config/avrdude/avrdude.rc
--- a/src/config.c
+++ b/src/config.c
@@ -36,6 +36,8 @@
 
 #include "config_gram.h"
 
+const char *avrdude_conf_version;
+
 const char *default_programmer;
 const char *default_parallel;
 const char *default_serial;
--- a/src/config_gram.y
+++ b/src/config_gram.y
@@ -74,6 +74,7 @@ static int pin_name;
 
 %token K_ALIAS
 %token K_ALLOW_SUBSHELLS
+%token K_AVRDUDE_CONF_VERSION
 %token K_BUFF
 %token K_CONNTYPE
 %token K_DEDICATED
@@ -212,6 +213,11 @@ def :
   prog_def TKN_SEMI |
 
   part_def TKN_SEMI |
+
+  K_AVRDUDE_CONF_VERSION TKN_EQUAL TKN_STRING TKN_SEMI {
+    avrdude_conf_version = cache_string($3->value.string);
+    free_token($3);
+  } |
 
   K_DEFAULT_PROGRAMMER TKN_EQUAL TKN_STRING TKN_SEMI {
     default_programmer = cache_string($3->value.string);
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -543,8 +543,9 @@ AC_CONFIG_FILES([
        Makefile
 ])
 
-# Pass into avrdude.conf.in
-AVRDUDE_FULL_VERSION='$(VERSION)'
+# Pass version number into avrdude.conf
+AVRDUDE_FULL_VERSION=$PACKAGE_VERSION
+AC_SUBST(AVRDUDE_FULL_VERSION, $AVRDUDE_FULL_VERSION)
 
 # The procedure to create avrdude.conf involves two steps.  First,
 # normal autoconf substitution will be applied, resulting in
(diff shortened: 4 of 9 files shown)


mcuee:
Great. This fixed the issue for git main.

```
$ ./avrdude_pr1547
Usage: avrdude_pr1547 [options]
Options:
  -p <partno>            Specify AVR device; -p ? lists all known parts
  -p <wildcard>/<flags>  Run developer options for matched AVR devices,
                         e.g., -p ATmega328P/s or /S for part definition
  -b <baudrate>          Override RS-232 baud rate
  -B <bitclock>          Specify bit clock period (us)
  -C <config-file>       Specify location of configuration file
  -c <programmer>        Specify programmer; -c ? and -c ?type list all
  -c <wildcard>/<flags>  Run developer options for matched programmers,
                         e.g., -c 'ur*'/s for programmer info/definition
  -A                     Disable trailing-0xff removal for file/AVR read
  -D                     Disable auto erase for flash memory; implies -A
  -i <delay>             ISP Clock Delay [in microseconds]
  -P <port>              Connection; -P ?s or -P ?sa lists serial ones
  -r                     Reconnect to -P port after "touching" it; wait
                         400 ms for each -r; needed for some USB boards
  -F                     Override invalid signature or initial checks
  -e                     Perform a chip erase
  -O                     Perform RC oscillator calibration (see AVR053)
  -t                     Run an interactive terminal when it is its turn
  -T <terminal cmd line> Run terminal line when it is its turn
  -U <memtype>:r|w|v:<filename>[:format]
                         Carry out memory operation when it is its turn
                         Multiple -t, -T and -U options can be specified
  -n                     Do not write to the device whilst processing -U
  -V                     Do not automatically verify during -U
  -E <exitsp>[,<exitsp>] List programmer exit specifications
  -x <extended_param>    Pass <extended_param> to programmer, see -xhelp
  -v                     Verbose output; -v -v for more
  -q                     Quell progress output; -q -q for less
  -l logfile             Use logfile rather than stderr for diagnostics
  -?                     Display this usage

avrdude version 7.2-20231030 (2736663b), https://github.com/avrdudes/avrdude

$ cat avrdude_pr1547.conf | head -n 10
# $Id$ -*- text -*-
#
# AVRDUDE Configuration File
#

avrdude_conf_version = "7.2-20231030 (2736663b)";

# This file contains AVRDUDE's default configuration data describing
# programming hardware pinouts and part definitions. Normally, it
# requires AVRDUDE be of the same or higher version than above string.

$ ./avrdude_pr1547 -C ./avrdude_pr1547.conf -c urclock -P ch340 -p m328p
avrdude_pr1547: AVR device initialized and ready to accept instructions
avrdude_pr1547: device signature = 0x1e950f (probably m328p)

avrdude_pr1547 done.  Thank you.
```

mcuee:
As mentioned in the following comment, we still need to fix `configure.ac`.
* https://github.com/avrdudes/avrdude/pull/1545#issuecomment-1784420284

mcuee:
The following patch mentioned by @Youw seems to work.  The only thing is that it does not contains the git commit ID (but consistent with the existing avrdude). It is better to have the commit ID since we may have several commits on the same date.

However, I think we can live with this for now with the auto-tools build.

```
/c/work/avr/avrdude_test/avrdude_pr1547
$ git diff
diff --git a/src/configure.ac b/src/configure.ac
index a67dac28..9f4f0cb1 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -543,8 +543,9 @@ AC_CONFIG_FILES([
        Makefile
 ])

-# Pass into avrdude.conf.in
-AVRDUDE_FULL_VERSION='$(VERSION)'
+# Pass version number into avrdude.conf
+AVRDUDE_FULL_VERSION=$PACKAGE_VERSION
+AC_SUBST(AVRDUDE_FULL_VERSION, $AVRDUDE_FULL_VERSION)

 # The procedure to create avrdude.conf involves two steps.  First,
 # normal autoconf substitution will be applied, resulting in

$ cd src/

$ head -n 10 avrdude.conf
# $Id$ -*- text -*-
#
# AVRDUDE Configuration File
#

avrdude_conf_version = "7.2-20230720";

# This file contains AVRDUDE's default configuration data describing
# programming hardware pinouts and part definitions. Normally, it
# requires AVRDUDE be of the same or higher version than above string.
```

MCUdude:
@stefanrueger Please add this to configure.ac, as previously done in #1545

```
# Pass version number into avrdude.conf
AVRDUDE_FULL_VERSION=$PACKAGE_VERSION
AC_SUBST(AVRDUDE_FULL_VERSION, $AVRDUDE_FULL_VERSION)
```

stefanrueger:
@MCUdude Like so?
``` diff
diff --git a/src/configure.ac b/src/configure.ac
index a67dac28..9f4f0cb1 100644
--- a/src/configure.ac
+++ b/src/configure.ac
@@ -543,8 +543,9 @@ AC_CONFIG_FILES([
        Makefile
 ])
 
-# Pass into avrdude.conf.in
-AVRDUDE_FULL_VERSION='$(VERSION)'
+# Pass version number into avrdude.conf
+AVRDUDE_FULL_VERSION=$PACKAGE_VERSION
+AC_SUBST(AVRDUDE_FULL_VERSION, $AVRDUDE_FULL_VERSION)
 
 # The procedure to create avrdude.conf involves two steps.  First,
 # normal autoconf substitution will be applied, resulting in
```

I don't understand this change, as I don't know where PACKAGE_VERSION is coming from... Your PR #1545 did another change that I now cannot remember that conflicted with git main.

MCUdude:
Yes, just like so.

`$PACKAGE_VERSION` is the version number passed to `AC_INIT`, and is a built-in feature of autoconf (I think).

https://github.com/avrdudes/avrdude/blob/9dfb61646a0f5ce4e4a55e16b075fe1a739a9a1e/src/configure.ac#L26

autoconf `AC_INIT`:
https://www.gnu.org/software/autoconf/manual/autoconf-2.67/html_node/Initializing-configure.html

#1545 also [did some changes to CMakeLists.txt](https://github.com/avrdudes/avrdude/pull/1545/commits/18587338d16e662ef1a03f2c66a6f8df71c50fee), but this wasn't intentional, and git refused to remove this commit from the PR. So the only thing you need to deal with is to add `AVRDUDE_FULL_VERSION` to configure.ac

stefanrueger:
Brilliant. Thanks for the explanation. Seeing that this PR will reverse the breakage of git main I suggest merging out of order soon. I can see the benefit of checking the independent version strings for AVRDUDE and avrdude.conf to issue a warning, but as this check is a bit more involved I will leave this for later, ideally for someone else (@MCUdude?)

@mcuee We have all pushed commits that break git main; letting a PR hang around for a couple of days is a great way for identifying teething problems :wink:

mcuee:
> @mcuee We have all pushed commits that break git main; letting a PR hang around for a couple of days is a great way for identifying teething problems 😉

Noted and thanks.

MCUdude:
> I can see the benefit of checking the independent version strings for AVRDUDE and avrdude.conf to issue a warning, but as this check is a bit more involved I will leave this for later, ideally for someone else (@MCUdude?)

Good! I'll see what I can do. Feel free to merge!

Issue #1249, opened 2022-12-29
Harmonize WIN32 implementation of serial time-outs with Posix one


mariusgreuel:
The Posix time-out mechanism for serial reads is specified as the total time for any 1024 byte chunk of data.

The WIN32 time-out is specified as `ReadTotalTimeoutConstant + buflen * ReadTotalTimeoutMultiplier`, which makes it potentially much longer than the Posix one (i.e. at least double). `ReadIntervalTimeout` is also set to `timeout`, which essentially renders the interval timeout ineffective, so IMHO, we should explicitly disable that feature.

We should probably change it to something like this, or even read the data in 1024 byte chunks, as the Posix one does.

```c
static BOOL serial_w32SetTimeOut(HANDLE hComPort, DWORD timeout) // in ms
{
	COMMTIMEOUTS ctmo;
	ZeroMemory (&ctmo, sizeof(COMMTIMEOUTS));
	ctmo.ReadIntervalTimeout = 0;
	ctmo.ReadTotalTimeoutMultiplier = 0;
	ctmo.ReadTotalTimeoutConstant = timeout;

	return SetCommTimeouts(hComPort, &ctmo);
}
```

However, this probably needs some extensive testing for those programmers where the time-out is carefully chosen to match the requirements.

See <https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddser/ns-ntddser-_serial_timeouts>

mcuee:
> However, this probably needs some extensive testing for those programmers where the time-out is carefully chosen to match the requirements.

So I think this can be postponed after 7.1 release.

stefanrueger:
@mariusgreuel I broadly agree with your analysis and had similar thoughts when reading the documentation. On the issue of reading in blocks of 1024 characters, I wondered whether it's OK to actually read byte by byte. It's serial comms after all and the overhead involved in bytewise read is minimal given today's computing power, even on a RPi.

mcuee:
@mariusgreuel 

Just wondering if you have some time to look at this. Thanks.

mcuee:
@stefanrueger 

Looks like @mariusgreuel is busy. Just wondering if you want to take a stab at this one.

mcuee:
Here is the first test.

```
PS C:\work\avr\avrdude_test\avrdude_main> git diff
diff --git a/src/ser_win32.c b/src/ser_win32.c
index b7480ac2..c5b9fafa 100644
--- a/src/ser_win32.c
[...]

PR #1550, opened 2023-11-01
Harmonize WIN32 implementation of serial time-outs with Posix

mcuee:
This is from @mariusgreuel in Issue #1249.
* https://github.com/avrdudes/avrdude/issues/1249

> The WIN32 time-out is specified as `ReadTotalTimeoutConstant + buflen * ReadTotalTimeoutMultiplier`, which makes it potentially much longer than the POSIX one (i.e. at least double). `ReadIntervalTimeout` is also set to timeout, which essentially renders the interval timeout ineffective, so IMHO, we should explicitly disable that feature.

### DIFF
--- a/src/ser_win32.c
+++ b/src/ser_win32.c
@@ -88,8 +88,8 @@ static BOOL serial_w32SetTimeOut(HANDLE hComPort, DWORD timeout) // in ms
 {
 	COMMTIMEOUTS ctmo;
 	ZeroMemory (&ctmo, sizeof(COMMTIMEOUTS));
-	ctmo.ReadIntervalTimeout = timeout;
-	ctmo.ReadTotalTimeoutMultiplier = timeout;
+	ctmo.ReadIntervalTimeout = 0;
+	ctmo.ReadTotalTimeoutMultiplier = 0;
 	ctmo.ReadTotalTimeoutConstant = timeout;
 
 	return SetCommTimeouts(hComPort, &ctmo);


mcuee:
I've carried some tests and it seems to work. However, as mentioned by @mariusgreuel, this needs more tests.

https://github.com/avrdudes/avrdude/issues/1249#issue-1513806121
> However, this probably needs some extensive testing for those programmers where the time-out is carefully chosen to match the requirements.
>
> See https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntddser/ns-ntddser-_serial_timeouts

mcuee:
```
COMMTIMEOUTS structure (winbase.h)
https://learn.microsoft.com/en-us/windows/win32/api/winbase/ns-winbase-commtimeouts

typedef struct _COMMTIMEOUTS {
  DWORD ReadIntervalTimeout;
  DWORD ReadTotalTimeoutMultiplier;
  DWORD ReadTotalTimeoutConstant;
  DWORD WriteTotalTimeoutMultiplier;
  DWORD WriteTotalTimeoutConstant;
} COMMTIMEOUTS, *LPCOMMTIMEOUTS;
```

stefanrueger:
I agree with this PR. After reading up on the COMMITTIMEOUTS structure had come to the same conclusion as @mariusgreuel. Quite a few users and applications will be affected by the change. Let's cross our fingers that this is not one of those situations where a particular application *relied* on the larger timeout (without AVRDUDE knowing about it). But yes, it's better to make this change now than 10 seconds before the release of 7.3 as, if it breaks something, we might hear back before 7.3

mcuee:
@stefanrueger 

Glad to hear that you agree with this PR. I agree with you that it is better to merge this soon so that issues will be reported within the next two months.

PR #1556, opened 2023-11-03
Remove paged eeprom property from ATtiny43U

stefanrueger:
The `paged` should only be used for flash of classic parts. It may lead to problems when used for `eeprom` memories.


MCUdude:
I don't have an ATtiny43U, but the PR looks OK!

mcuee:
> I don't have an ATtiny43U, but the PR looks OK!

Same here. This PR is good to go.

PR #1563, opened 2023-11-09
Always print error if part and programmer doesn't have a common programming interface

MCUdude:
See #1561 for details. OP used `-F` and didn't get a warning that the part and programmer doesn't share a common programming interface.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1111,10 +1111,11 @@ int main(int argc, char * argv [])
   }
 
   if(partdesc && (p = locate_part(part_list, partdesc)) && !(p->prog_modes & pgm->prog_modes)) {
-    pmsg_error("programmer %s cannot program part %s as they\n", pgmid, p->desc);
-    imsg_error("lack a common programming mode; use -F to override this check\n");
-    if(!ovsigck)
+    pmsg_error("-c %s cannot program %s for lack of a common programming mode\n", pgmid, p->desc);
+    if(!ovsigck) {
+      imsg_error("use -F to override this check\n");
       exit(1);
+    }
   }
 
   if (pgm->initpgm) {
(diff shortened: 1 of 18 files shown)


src/main.c
```diff
@@ -1110,10 +1110,11 @@ int main(int argc, char * argv [])
     exit(1);
   }
 
-  if(!ovsigck && partdesc && (p = locate_part(part_list, partdesc)) && !(p->prog_modes & pgm->prog_modes)) {
+  if(partdesc && (p = locate_part(part_list, partdesc)) && !(p->prog_modes & pgm->prog_modes)) {
```
stefanrueger, line 1113: Should we print use -F to override this check when -F was specified?

stefanrueger:
I'd suggest something like 
``` diff
diff --git a/src/main.c b/src/main.c
index 2d0b3fa6..19b41413 100644
--- a/src/main.c
+++ b/src/main.c
@@ -1110,10 +1110,12 @@ int main(int argc, char * argv [])
     exit(1);
   }
 
-  if(!ovsigck && partdesc && (p = locate_part(part_list, partdesc)) && !(p->prog_modes & pgm->prog_modes)) {
-    pmsg_error("programmer %s cannot program part %s as they\n", pgmid, p->desc);
-    imsg_error("lack a common programming mode; use -F to override this check\n");
-    exit(1);
+  if(partdesc && (p = locate_part(part_list, partdesc)) && !(p->prog_modes & pgm->prog_modes)) {
+    pmsg_error("-c %s cannot program %s for lack of a common programming mode\n", pgmid, p->desc);
+    if(!ovsigck) {
+      imsg_error("use -F to override this check\n");
+      exit(1);
+    }
   }
 
   if (pgm->initpgm) {
```

Issue #1562, opened 2023-11-09
Print warning when avrdude and avrdude.conf versions doesn't match


MCUdude:
Now that #1547 has been merged, we can utilize this to print a warning if the user is using an avrdude.conf file that are not up to date.

PR #1564, opened 2023-11-09
Print warning when avrdude and avrdude.conf versions doesn't match

MCUdude:
With this PR Avrdude will print a warning if the avrdude.conf file version doesn't match the avrdude executable version.

```
$ cat avrdude.conf | grep avrdude_conf_version
avrdude_conf_version = "7.1";

$ ./avrdude -cdryrun -patmega328p -PUSB 
avrdude warning: System wide configuration file version (7.1)
        does not match Avrdude build version (7.2-20230720)
avrdude: device signature = 0x1e950f (probably m328p)

avrdude done.  Thank you.
```

Resolves issue #1562

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -995,6 +995,11 @@ int main(int argc, char * argv [])
     }
   }
 
+  if(!str_starts(avrdude_conf_version, version)) {
+    pmsg_warning("System wide configuration file version (%s)\n", avrdude_conf_version);
+    imsg_warning("does not match Avrdude build version (%s)\n", version);
+  }
+
   if (lsize(additional_config_files) > 0) {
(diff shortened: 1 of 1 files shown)


mcuee:
Looks fine to me.

stefanrueger:
I'd rather warn always if the version numbers do not coincide. Technically, AVRDUDE should be able to cope with older configuration files, but in practice you might expect malfunctions. So, I suggest to test `str_eq()` rather than `str_starts()`.

MCUdude:
> I'd rather warn always if the version numbers do not coincide.

Sure! I've replaced str_starts with str_eq now

PR #1565, opened 2023-11-10
Support for STK500 v1 devices with STK500_XTAL != 7372800

Ho-Ro:
As discussed in #1560 - please check

```
$ avrdude_Ho-Ro -pt85 -cstk500v1 -x xtal=16M -v

avrdude_Ho-Ro: Version 7.2-20231110 (9e41b00)

...
               Programmer Type : STK500
               Description     : Atmel STK500 version 1.x firmware
               Hardware Version: 2
               Firmware Version: 1.28
               Vtarget         : 4.8 V
               Varef           : 0.0 V
               Oscillator      : 1.000 MHz
               SCK period      : 0.5 us
               XTAL frequency  : 16.000 MHz
avrdude_Ho-Ro: AVR device initialized and ready to accept instructions
avrdude_Ho-Ro: device signature = 0x1e930b (probably t85)

avrdude_Ho-Ro done.  Thank you.
```

PR #1566, opened 2023-11-10
Support for STK500 v1 and v2 devices with STK500_XTAL != 7372800

Ho-Ro:
This PR replaces #1565 because it introduced a lot of whitespace-only changes. Sorry for the confusion.
1st part fixed some errors while 2nd commit does the real work.

### DIFF
--- a/src/stk500.c
+++ b/src/stk500.c
@@ -862,6 +862,11 @@ static int stk500_open(PROGRAMMER *pgm, const char *port) {
   if (stk500_getsync(pgm) < 0)
     return -1;
 
+  if (pgm->bitclock != 0.0) {
+    if (pgm->set_sck_period(pgm, pgm->bitclock) != 0)
+      return -1;
+  }
+
   return 0;
 }
 
--- a/src/stk500.c
+++ b/src/stk500.c
@@ -635,25 +635,22 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
     }
 
     else if (str_starts(extended_param, "vtarg")) {
-      if (pgm->extra_features & HAS_VTARG_ADJ) {
+      if ((pgm->extra_features & HAS_VTARG_ADJ) && (str_starts(extended_param, "vtarg=")))  {
         // Set target voltage
-        if (str_starts(extended_param, "vtarg=") ) {
-          double vtarg_set_val = 0;
-          int sscanf_success = sscanf(extended_param, "vtarg=%lf", &vtarg_set_val);
-          PDATA(pgm)->vtarg_data = (double)((int)(vtarg_set_val * 100 + .5)) / 100;
-          if (sscanf_success < 1 || vtarg_set_val < 0) {
-            pmsg_error("invalid vtarg value '%s'\n", extended_param);
-            rv = -1;
-            break;
-          }
-          PDATA(pgm)->vtarg_set = true;
-          continue;
+        double vtarg_set_val = 0;
+        int sscanf_success = sscanf(extended_param, "vtarg=%lf", &vtarg_set_val);
+        PDATA(pgm)->vtarg_data = (double)((int)(vtarg_set_val * 100 + .5)) / 100;
+        if (sscanf_success < 1 || vtarg_set_val < 0) {
+          pmsg_error("invalid vtarg value '%s'\n", extended_param);
+          rv = -1;
+          break;
         }
+        PDATA(pgm)->vtarg_set = true;
+        continue;
+      } else if ((pgm->extra_features & HAS_VTARG_READ) && str_eq(extended_param, "vtarg")) {
         // Get target voltage
-        else if(str_eq(extended_param, "vtarg")) {
-          PDATA(pgm)->vtarg_get = true;
-          continue;
-        }
+        PDATA(pgm)->vtarg_get = true;
+        continue;
       }
     }
 
@@ -691,7 +688,7 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
     }
 
     else if (str_starts(extended_param, "fosc")) {
-      if (pgm->extra_features & HAS_VAREF_ADJ) {
+      if (pgm->extra_features & HAS_FOSC_ADJ) {
         // Set clock generator frequency
         if (str_starts(extended_param, "fosc=")) {
           char fosc_str[16] = {0};
@@ -730,8 +727,10 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
     else if (str_eq(extended_param, "help")) {
       msg_error("%s -c %s extended options:\n", progname, pgmid);
       msg_error("  -xattempts=<arg>      Specify no. connection retry attempts\n");
-      if (pgm->extra_features & HAS_VTARG_ADJ) {
+      if (pgm->extra_features & HAS_VTARG_READ) {
         msg_error("  -xvtarg               Read target supply voltage\n");
+      }
+      if (pgm->extra_features & HAS_VTARG_ADJ) {
         msg_error("  -xvtarg=<arg>         Set target supply voltage\n");
       }
       if (pgm->extra_features & HAS_VAREF_ADJ) {
--- a/src/stk500.c
+++ b/src/stk500.c
@@ -401,12 +401,15 @@ static int stk500_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   unsigned char buf[32];
   AVRMEM * m;
   int tries;
-  unsigned maj, min;
+  unsigned maj = 0, min = 0;
   int rc;
   int n_extparms;
 
-  stk500_getparm(pgm, Parm_STK_SW_MAJOR, &maj);
-  stk500_getparm(pgm, Parm_STK_SW_MINOR, &min);
+  if ((rc = stk500_getparm(pgm, Parm_STK_SW_MAJOR, &maj)) < 0
+     || (rc = stk500_getparm(pgm, Parm_STK_SW_MINOR, &min)) < 0 ) {
+    pmsg_error("cannot obtain SW version\n");
+    return rc;
+  }
 
   // MIB510 does not need extparams
   if (str_eq(pgmid, "mib510"))
@@ -559,8 +562,11 @@ static int stk500_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   // Read or write target voltage
   if (PDATA(pgm)->vtarg_get || PDATA(pgm)->vtarg_set) {
     // Read current target voltage set value
-    unsigned int vtarg_read;
-    stk500_getparm(pgm, Parm_STK_VTARGET, &vtarg_read);
+    unsigned int vtarg_read = 0;
+    if ((rc = stk500_getparm(pgm, Parm_STK_VTARGET, &vtarg_read)) < 0) {
+      pmsg_error("cannot obtain V[target]\n");
+      return rc;
+    }
     if (PDATA(pgm)->vtarg_get)
       msg_info("Target voltage value read as %.2f V\n", (vtarg_read / 10.0));
     // Write target voltage value
@@ -574,8 +580,11 @@ static int stk500_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   // Read or write analog reference voltage
   if (PDATA(pgm)->varef_get || PDATA(pgm)->varef_set) {
     // Read current analog reference voltage
-    unsigned int varef_read;
-    stk500_getparm(pgm, Parm_STK_VADJUST, &varef_read);
+    unsigned int varef_read = 0;
+    if ((rc = stk500_getparm(pgm, Parm_STK_VADJUST, &varef_read)) < 0) {
+      pmsg_error("cannot obtain V[aref]\n");
+      return rc;
+    }
     if (PDATA(pgm)->varef_get)
       msg_info("Analog reference voltage value read as %.2f V\n", (varef_read / 10.0));
     // Write analog reference voltage
@@ -590,12 +599,15 @@ static int stk500_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
   // Read or write clock generator frequency
   if (PDATA(pgm)->fosc_get || PDATA(pgm)->fosc_set) {
     // Read current target voltage set value
-    unsigned int osc_pscale;
-    unsigned int osc_cmatch;
+    unsigned int osc_pscale = 0;
+    unsigned int osc_cmatch = 0;
     const char *unit_get = {"Hz"};
     double f_get = 0.0;
-    stk500_getparm(pgm, Parm_STK_OSC_PSCALE, &osc_pscale);
-    stk500_getparm(pgm, Parm_STK_OSC_CMATCH, &osc_cmatch);
+    if ((rc = stk500_getparm(pgm, Parm_STK_OSC_PSCALE, &osc_pscale)) < 0
+      || (rc = stk500_getparm(pgm, Parm_STK_OSC_CMATCH, &osc_cmatch) < 0)) {
+      pmsg_error("cannot obtain fosc values\n");
+      return rc;
+    }
     if(osc_pscale) {
       int prescale = 1;
       f_get = PDATA(pgm)->xtal / 2;
@@ -1165,19 +1177,19 @@ static int stk500_paged_load(const PROGRAMMER *pgm, const AVRPART *p, const AVRM
 
 
 static int stk500_set_vtarget(const PROGRAMMER *pgm, double v) {
-  unsigned uaref, utarg;
+  unsigned uaref = 0;
+  unsigned utarg = (unsigned)((v + 0.049) * 10);
+  int rc = 0;
 
-  utarg = (unsigned)((v + 0.049) * 10);
-
-  if (stk500_getparm(pgm, Parm_STK_VADJUST, &uaref) != 0) {
+  if ((rc = stk500_getparm(pgm, Parm_STK_VADJUST, &uaref)) != 0) {
     pmsg_error("cannot obtain V[aref]\n");
-    return -1;
+    return rc;
   }
 
   if (uaref > utarg) {
     pmsg_warning("reducing V[aref] from %.1f to %.1f\n", uaref / 10.0, v);
-    if (stk500_setparm(pgm, Parm_STK_VADJUST, utarg) != 0)
-      return -1;
+    if ((rc = stk500_setparm(pgm, Parm_STK_VADJUST, utarg)) != 0)
+      return rc;
   }
   return stk500_setparm(pgm, Parm_STK_VTARGET, utarg);
 }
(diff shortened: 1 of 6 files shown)


Ho-Ro:
STK500 v2 takes some more time. The source code is more confusing because so many different programmers are involved and the crystal frequency has also been coded into many different "magic constants".

stefanrueger:
> STK500 v2 takes some more time.

Thanks for looking into this! I think it will be worth our while. Let us know once you are finished with code and documentation in avrdude.1 and doc/avrdude.texi; then the team will be able to review and test.

Ho-Ro:
stk500v2 support for `-x xtal` is done, also `avrdude.1` and `doc/avrdude.texi` - please review.

src/stk500.c
```diff
@@ -707,7 +704,7 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
             if (str_eq(fosc_str, "off"))
               PDATA(pgm)->fosc_data = 0.0;
             else {
-              pmsg_error("cannot parse fosc value %s\n", fosc_str);
+              pmsg_error("invalid fosc value '%s'\n", fosc_str);
```
stefanrueger, line 707: ever so slightly prefer %s without single quotes in common with other messages

src/stk500.c
```diff
@@ -590,7 +590,7 @@ static int stk500_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
     stk500_getparm(pgm, Parm_STK_OSC_CMATCH, &osc_cmatch);
     if(osc_pscale) {
       int prescale = 1;
-      f_get = STK500_XTAL / 2;
+      f_get = PDATA(pgm)->xtal / 2;
```
stefanrueger, line 593: better divide by 2.0 (otherwise lsb is lost)

src/stk500.c
```diff
@@ -727,11 +726,45 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
       }
     }
 
+    else if (str_starts(extended_param, "xtal")) {
+      // Set clock generator frequency
+      if (str_starts(extended_param, "xtal=")) {
+        char xtal_str[16] = {0};
+        int sscanf_success = sscanf(extended_param, "xtal=%10s", xtal_str);
+        if (sscanf_success < 1) {
+          pmsg_error("invalid xtal value '%s'\n", extended_param);
```
stefanrueger, line 735: no single quotes please in message

src/stk500.c
```diff
@@ -727,11 +726,45 @@ static int stk500_parseextparms(const PROGRAMMER *pgm, const LISTID extparms)
...
+          rv = -1;
+          break;
+        }
+        char *endp;
+        double v = strtod(xtal_str, &endp);
+        if (endp == xtal_str){
```
stefanrueger, line 741: Would be neat to allow space between number and unit; eg, with code to the effect of
```
  if(*endp == ' ')
    endp++;
```

src/stk500.c
```diff
@@ -1167,31 +1206,31 @@ static int stk500_set_fosc(const PROGRAMMER *pgm, double v) {
 
   prescale = cmatch = 0;
   if (v > 0.0) {
-    if (v > STK500_XTAL / 2) {
+    if (v > PDATA(pgm)->xtal / 2) {
```
stefanrueger, line 1209: 2.0 (multiple times)

src/stk500.c
```diff
@@ -1415,7 +1454,19 @@ static void stk500_print_parms1(const PROGRAMMER *pgm, const char *p, FILE *fp)
...
+  double f = PDATA(pgm)->xtal;
+  if (f >= 1e6) {
+    f /= 1e6;
+    unit = "MHz";
+  } else if (f >= 1e3) {
+    f /= 1000;
+    unit = "kHz";
+  } else
+    unit = "Hz";
+  fmsg_out(fp, "%sXTAL frequency  : %.3f %s\n", p, f, unit);
```
stefanrueger, line 1469: Is .3 sufficient? Might .5 or .6 be better?

src/stk500.h
```diff
@@ -57,6 +57,9 @@ struct pdata {
   bool fosc_get;
   bool fosc_set;
   double fosc_data;
+
```
stefanrueger, line 60: Perhaps neater to have inline comment at column 32 without empty line? So adding a component is one line, not three.

src/stk500v2.c
```diff
@@ -3710,6 +3750,20 @@ static void stk500v2_print_parms1(const PROGRAMMER *pgm, const char *p, FILE *fp
...
+  double f = PDATA(pgm)->xtal;
+  if (f >= 1e6) {
+    f /= 1e6;
+    unit = "MHz";
+  } else if (f >= 1e3) {
+    f /= 1000;
+    unit = "kHz";
+  } else
+    unit = "Hz";
+  fmsg_out(fp, "%sXTAL frequency  : %.3f %s\n", p, f, unit);
```
stefanrueger, line 3766: No space before colon

MCUdude: He may have added that space so that all colons lines up when running the `parms` command in terminal mode.

stefanrueger (commented):
Code looks vvv good; I only have minor (minor) suggestions, which need applying multiple times; still needs to be tested independently, but all looks vvv good

src/stk500v2.c
stefanrueger: Of course, I missed that. Thanks for reviewing the review

src/stk500.c
Ho-Ro: Good POLA point, I will check.

src/stk500.c
Ho-Ro: Yes, I changed to variable decimal count 6 for MHz, 3 for kHz, 0 for Hz

Ho-Ro:
@stefanrueger et. al. What's the next step, should I add another commit with all requested changes and you "squash and merge" them as one commit?
Or should I tidy up on my side?

mcuee:
Unfortunately my original STK500 was dead so I can not carry out real test. I have a few STK500 clones which only implement the programming portion (including one with HVPP capability) but not other features.

@MCUdude 
Just wondering if you can carry out the real test.

MCUdude:
> Unfortunately my original STK500 was dead so I can not carry out real test. I have a few STK500 clones which only implement the programming portion (including one with HVPP capability) but not other features.
@MCUdude
Just wondering if you can carry out the real test.

Sure! I'll give it a test tonight.

"How dead" are your STK500 board? I accidentally bricked one of mine, but I was able to get it working again by re-flashing the firmware for the ATtiny2313 (running AVR910) and the ATmega8515 (running STK500 FW). I think I have the correct hex files somewhere if that's your issue.

mcuee:
> "How dead" are your STK500 board? I accidentally bricked one of mine, but I was able to get it working again by re-flashing the firmware for the ATtiny2313 (running AVR910) and the ATmega8515 (running STK500 FW). I think I have the correct hex files somewhere if that's your issue.

I have the necessary STK500 V1/V2 FW hex files -- that is why I can switch between STK500 V1 and STK500 V2 for two clones using ATmega8535 MCU.

Unfortunately for my original STK500 it is HW failure and it is hard to repair now due to broken traces. It was a very old unit (probably made in 2005/2006) and I got from my colleague a few years ago.

MCUdude:
I just tested this PR on two separate STK500 boards running V1 and v2 firmware, and when I changed -xxtal to 14.7456 MHz, I got the reported `fosc` was twice the frequency that I measured. All good then!

Two things:
*`-xxtal` is not an option for `-c stk500` even though it probably should be.
* Could it be an idea to add `xtal` as a command in terminal mode?

```
avrdude> xtal
Syntax: xtal <value>[M|k]
avrdude> xtal 16M
```

```
$ ./avrdude -cstk500 -xhelp
avrdude -c stk500 extended options:
  -xhelp    Show this help menu and exit
```

MCUdude:
> Could it be an idea to add xtal as a command in terminal mode?

When thinking a bit; probably not, as it doesn't change anything on the board other than a constant used for calculation. However, it should probably not be present when running the `parms` command.

Ho-Ro:
> However, it should probably not be present when running the parms command.

I agree, as it cannot be changed from terminal cmd.

src/stk500.c
stefanrueger: Users are unlikely to use quotes but scripts that extract the frequency from another AVRDUDE or otherwise call might have a frequency as `1.23 MHz` and find it easy to use quotes but marginally harder to remove the space between 3 and MHz. It is always neat if the output of AVRDUDE (correctly done with space between number and unit) can be used as input without mods.

src/stk500.c
```diff
@@ -1402,20 +1450,19 @@ static void stk500_print_parms1(const PROGRAMMER *pgm, const char *p, FILE *fp)
...
-        unit = "MHz";
-      } else if (f > 1e3) {
-        f /= 1000;
-        unit = "kHz";
-      } else
-        unit = "Hz";
-      fmsg_out(fp, "%.3f %s\n", f, unit);
+      decimals = get_decimals(f);
+      f = f_to_kHz_MHz(f, &unit);
+      fmsg_out(fp, "%.*f %s\n", f, decimals, unit);
```
stefanrueger, line 1455: Must be `fmsg_out(fp, "%.*f %s\n", decimals, f, unit);` ie, decimals first.

src/stk500v2.c
```diff
@@ -3706,11 +3755,20 @@ static void stk500v2_print_parms1(const PROGRAMMER *pgm, const char *p, FILE *fp
         }
         f /= prescale;
         f /= (osc_cmatch + 1);
+        decimals = get_decimals(f);
         f = f_to_kHz_MHz(f, &unit);
-        fmsg_out(fp, "%.3f %s\n", f, unit);
+        fmsg_out(fp, "%.*f %s\n", f, decimals, unit);
```
stefanrueger, line 3760: change order of f and decimals arguments (multiple times)

src/stk500v2.c
```diff
@@ -1306,7 +1314,7 @@ static int stk500v2_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
       stk500v2_getparm(pgm, PARAM_OSC_CMATCH, &osc_cmatch);
       if(osc_pscale) {
         int prescale = 1;
-        f_get = STK500V2_XTAL / 2;
+        f_get = PDATA(pgm)->xtal / 2;
```
stefanrueger, line 1317: Not a fault of this PR, but `osc_pscale` and `osc_cmatch` ought to be set to 0 above b/c `stk500v2_getparm()` might not set them! Ideally, *also* the return value of `stk500v2_getparm()` calls ought to be checked to guard against not setting the `osc_...` variables.

src/stk500v2.c
```diff
@@ -1620,7 +1628,7 @@ static int stk500hv_initialize(const PROGRAMMER *pgm, const AVRPART *p, enum hvm
...
```
stefanrueger, line 1631: Same as above:

Not a fault of this PR, but osc_pscale and osc_cmatch ought to be set to 0 above b/c stk500v2_getparm() might not set them! Ideally, also the return value of stk500v2_getparm() calls ought to be checked to guard against not setting the osc_... variables.

src/stk500v2.c
```diff
@@ -1270,10 +1278,10 @@ static int stk500v2_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
       unsigned char varef_read;
       stk500v2_getparm(pgm, PARAM_VADJUST, &varef_read);
       if (PDATA(pgm)->varef_get)
-        msg_info("Analog reference voltage value read as %.2fV\n", (varef_read / 10.0));
+        msg_info("Analog reference voltage value read as %.2f V\n", (varef_read / 10.0));
       // STK500: Write analog reference voltage
       else {
-        msg_info("Changing analog reference voltage from %.2f to %.2fV\n",
+        msg_info("Changing analog reference voltage from %.2f V to %.2f V\n",
```
stefanrueger, line 1284: `unsigned char varef_read` above should be initialised to 0 (same reason as below, four times in this code)

src/stk500v2.c
```diff
@@ -1254,10 +1262,10 @@ static int stk500v2_initialize(const PROGRAMMER *pgm, const AVRPART *p) {
     unsigned char vtarg_read;
     stk500v2_getparm(pgm, PARAM_VTARGET, &vtarg_read);
     if (PDATA(pgm)->vtarg_get)
-      msg_info("Target voltage value read as %.2fV\n", (vtarg_read / 10.0));
+      msg_info("Target voltage value read as %.2f V\n", (vtarg_read / 10.0));
     // Write target voltage value
     else {
-      msg_info("Changing target voltage from %.2f to %.2fV\n", (vtarg_read / 10.0), PDATA(pgm)->vtarg_data);
+      msg_info("Changing target voltage from %.2f V to %.2f V\n", (vtarg_read / 10.0), PDATA(pgm)->vtarg_data);
```
stefanrueger, line 1268: `unsigned char vtarg_read;` should be initialised with 0 (same reason as below, needs to be done twice in this file).

stefanrueger:
> What's the next step

Normally someone reviews *all* changes and also the surrounding code (sometimes there are problems that are outside the PR but should be fixed "while at it"). Then things are tested. Then the PR hangs around for a few days (can be weeks) until a bunch of  PRs is merged. I sometimes squash and merge (eg, when a few ideas were tested) or merge when it's not problematic that a few fixes were committed. If it's important to you that a PR goes in with a particular pristine commit sequence it is normally better to recreate  a new PR. Force pushes are not really encouraged as they can go wrong.

src/stk500.c
Ho-Ro: Right, star comes before f in the string - I know I checked it later with the C standard, but only corrected line 1465 below and forgot the other occurrences here and in v2.
Funny thing is that gcc doesn't complain when I do it the wrong way round and does *"what I mean"* - I don't know what clang does though.

src/stk500v2.c
Ho-Ro: yep, 2 times

src/stk500v2.c
Ho-Ro: This will be a bigger separate commit, because `stk500v2_getparm()` (as well as `stk500_getparm()` in `stk500.c`) is used quite often w/o checking the return value. Nevertheless I can *adopt* this issue and fix it in the course of my stk500* overhaul.

Ho-Ro:
>  I sometimes squash and merge 

I think it would then be ok to stack some more commits  e.g. the check of `*_getparm()` on top of my previuous work.

src/stk500.c
stefanrueger: > does *"what I mean"*

A fluke maybe? Maybe doubles are passed in a different frame to functions than smaller data types via the stack?

stefanrueger: > gcc doesn't complain

gcc does not realise that the called function resolves to or calls printf()

src/stk500v2.c
stefanrueger: Thanks for cleaning this in your overhaul of stk500* - much appreciated.

src/stk500.c
Ho-Ro: ok, c983e30 allows spaces in extra parameter, e.g. `-x "fosc= .1e3 kHz"`

stefanrueger:
Code of new commits looks good. Thank you, @Ho-Ro. 

@mcuee @MCUdude Once this is thoroughly tested I think it can be merged

stefanrueger:
@mcuee @MCUdude please test the `-xxtal` functionality and test against regressions so this PR can be merged at the next occasion.

mcuee:
> @mcuee @MCUdude please test the `-xxtal` functionality and test against regressions so this PR can be merged at the next occasion.

Unfortunately I do not have the right HW to test this one as mentioned earlier. Hans should be able to test.

Ho-Ro:
> Unfortunately I do not have the right HW to test this one as mentioned earlier. Hans should be able to test.

I'm back at my home desk and can also provide some evidence for tests with my (original) STK500 with v2 FW.

MCUdude:
Just tested with two separate STK500 board running v1 and v2 firmware. Everything seems to work fine! I've not found any issues.

Ho-Ro:
I measured the SCK timing

### nanoSTK (stk500v1)

```sh
$ avrdude_Ho-Ro -pm328p -cstk500v1 -v -xxtal=16M -U flash:r:/dev/null:r -B1
...

               Programmer Type : STK500
               Description     : Atmel STK500 version 1.x firmware
               Hardware Version: 2
               Firmware Version: 1.29
               Vtarget         : 4.8 V
               Varef           : 0.0 V
               Oscillator      : 1.000000 MHz
               SCK period      : 1.0 us
               XTAL frequency  : 16.000000 MHz
avrdude_Ho-Ro: AVR device initialized and ready to accept instructions
avrdude_Ho-Ro: device signature = 0x1e950f (probably m328p)

avrdude_Ho-Ro: processing -U flash:r:/dev/null:r
avrdude_Ho-Ro: reading flash memory ...
Reading | ################################################## | 100% 4.47 s 
avrdude_Ho-Ro: writing output file /dev/null

avrdude_Ho-Ro done.  Thank you.
```
SCK timing nanoSTK (stk500v1)



### Original STK500

```
$ avrdude_Ho-Ro -pm328p -cstk500v2 -v -U flash:r:/dev/null:r -B8

...

               Programmer Type : STK500V2
               Description     : Atmel STK500 version 2.x firmware
               Programmer Model: STK500
               Hardware Version: 2
               Firmware Version Controller : 2.01
               Topcard         : Unknown
               Vtarget         : 5.0 V
               Varef           : 0.0 V
               Oscillator      : 3.686400 MHz
               SCK period      : 8.7 us
               XTAL frequency  : 7.372800 MHz
avrdude_Ho-Ro: AVR device initialized and ready to accept instructions
avrdude_Ho-Ro: device signature = 0x1e950f (probably m328p)

avrdude_Ho-Ro: processing -U flash:r:/dev/null:r
avrdude_Ho-Ro: reading flash memory ...
Reading | ################################################## | 100% 16.46 s 
avrdude_Ho-Ro: writing output file /dev/null

avrdude_Ho-Ro done.  Thank you.
```
SCK timing of STK500 (stk500v2)


And with half speed xtal setting it calculates the values correctly based on this setting (but uses normal speed as expected)

```
$ avrdude_Ho-Ro -pm328p -cstk500v2 -v -U flash:r:/dev/null:r -xxtal=3.686400M -B17

...

               Programmer Type : STK500V2
               Description     : Atmel STK500 version 2.x firmware
               Programmer Model: STK500
               Hardware Version: 2
               Firmware Version Controller : 2.01
               Topcard         : Unknown
               Vtarget         : 5.0 V
               Varef           : 0.0 V
               Oscillator      : 1.843200 MHz
               SCK period      : 17.4 us
               XTAL frequency  : 3.686400 MHz

```

stefanrueger:
Brilliant, thanks for confirming and tests @Ho-Ro and @MCUdude

PR #1553, opened 2023-11-02
To adopt PR #1264 approach of using Readline replacement for MinGW build (WIP, DNM)

mcuee:
PR https://github.com/avrdudes/avrdude/pull/1264 has been merged and it performs much better than using GNU Readline under Windows. It is good to adopt the same approach for MinGW.

This PR will fix Issue #1271.

This PR was initially developed by Xiaofan Chen based on PR #1264 but it does not work.
@Youw (Ihor Dutchak) helps to sort out the remaining issues.

PR #1569, opened 2023-11-14
Update src/CMakeLists.txt

mcuee:
This is to adopt the CMake improvements in PR #1553 even though the PR itself was abondoned.

The author of the change is @Youw.


mcuee:
@stefanrueger and @MCUdude 
This change should be safe to merge. As always, we should probably wait for a few days before merging.

PR #1577, opened 2023-11-23
Provide and use libavrude functions to access part config values

### LINKED ISSUE #1537: jtagmkI.c and jtagmkII.c rely on OCDEN being bit 7 of hfuse
... but on the ATmega406 it's bit 1
Screenshot from 2023-10-26 23-10-49

And here an example where it is used:
https://github.com/avrdudes/avrdude/blob/d6c61f5fab4d0d061fb85233acee6ecaa85dd9c0/src/jtagmkII.c#L98-L107

https://github.com/avrdudes/avrdude/blob/d6c61f5fab4d0d061fb85233acee6ecaa85dd9c0/src/jtagmkII.c#L1350-L1352

Not a biggie (false alerts) but should be fixed at some point.

stefanrueger:
This PR provides `avr_get_config_value()` and `avr_set_config_value()` functions

Example usage:
```    
  int ocden = 0;
  if(avr_get_config_value(pgm, p, "ocden", &ocden) == 0 && ocden) // ocden == 1 means disabled
    pmsg_warning("OCDEN fuse not programmed, single-byte EEPROM updates not possible\n");
```

Fixes #1537 

Best way to test this PR is to program a classic part with jtag mkI or jtag mkII programmers. There should be no warning when the OCDEN fuse is programmed, and there should be a warning if it isn't.

### DIFF
--- a/src/avrcache.c
+++ b/src/avrcache.c
@@ -29,7 +29,6 @@
 
 #include "avrdude.h"
 #include "libavrdude.h"
-#include "avrintel.h"
 
 /*
  * Provides an API for cached bytewise access
--- a/src/avrpart.c
+++ b/src/avrpart.c
@@ -493,6 +493,108 @@ AVRMEM_ALIAS *avr_find_memalias(const AVRPART *p, const AVRMEM *m_orig) {
   return NULL;
 }
 
+const Configitem_t *avr_locate_config(const Configitem_t *cfg, int nc, const char *name) {
+  const Configitem_t *match = NULL;
+  int matches = 0, n1;
+
+  if(!cfg || !name || !(n1 = *name))
+    return NULL;
+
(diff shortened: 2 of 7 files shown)


MCUdude:
Thanks for the PR! I just tested it using an AVR Dragon (in JTAG mode), and it works perfectly fine when connected to an ATmega1284p.

stefanrueger:
Embarrassingly, I found a mistake for accessing the config table with "ocden" fuse info, which will not have been noticed by testing. Corrected now.

stefanrueger:
While at it provided a few libavrdude functions (not currently used for AVRDUDE) that are might become useful for other projects (or future AVRDUDE functionality). Planning a new terminal command `regfile`, so looking to merging soon to avoid merge conflicts.

PR #1581, opened 2023-11-26
Ignore port settings when using -c dryrun

### LINKED ISSUE #1578: Upload port required when using `-cdryrun` on MacOS
Why do I have to specify an upload port (I've been using `-PUSB`) when using the dryrun programmer on MacOS?
Is it a bug, or does it just have to be this way?

```
$ ./avrdude -cdryrun -patmega1284p

avrdude error: no port has been specified on the command line or in the config file
        specify a port using the -P option and try again
```

stefanrueger:
Fixes #1578

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1215,7 +1215,8 @@ int main(int argc, char * argv [])
     }
   }
 
-  if(port[0] == 0 || str_eq(port, "unknown")) {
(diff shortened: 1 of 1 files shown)


mcuee:
This should be good to be merged.

PR #1583, opened 2023-11-28
Add more PL2303 compatible USB PIDs

MCUdude:
I recently bought a PL2303GC chip to replace a USB to serial adapter I accidentally fried.
To my surprise, this chip uses a different USB PID than the more common PL2303HX (0x23a3 instead of 0x2303).

I found [a list](https://github.com/torvalds/linux/blob/18d46e76d7c2eedd8577fae67e3f1d4db25018b0/drivers/usb/serial/pl2303.h#L3) of the most common PL2303 PIDs supported by the Linux kernel. I decided to only include the "native" ones, and not 3rd party VID/PIDs.


mcuee:
This should be safe to merge.

Issue #1582, opened 2023-11-27
Poor coverage of (concept) index in documentation


stefanrueger:
Now that the functionality and documentation of AVRDUDE has grown over the last couple of years, I couldn't help noticing that the (concept) index of the documentation does not have a great coverage:

![index](https://github.com/avrdudes/avrdude/assets/14282046/16145211-8df2-43b8-8640-bc713f6bd844)

It is a matter of carefully sprinkling a number of relevant `@cindex` entries into `src/doc/avrdude.texi`. That needs attention to detail and style, but would be an activity worth pursuing before the next release.

Ho-Ro:
The 1st step can be done automatically:

```python
#!/usr/bin/python

'''
Create @cindex entries for all @section, @subsection and @subsubsection entries.
Also create "@cindex Option -..." entries for all "@item -..." entries.
And finally replace the ugly "-x ..." index entries with "Option -x ...".
Write new texi file to stdout.
'''

filename = 'avrdude.texi'

with open( filename, 'r' ) as texi:
  for line in texi:
    cindex = None
    for tag, index_tag in (
        ( '@section', '@cindex' ),
        ( '@subsection', '@cindex' ),
        ( '@subsubsection', '@cindex' ),
        ( '@item -', '@cindex Option -' )
    ):
      # create index from structure elements
      if line.startswith( tag ):
        cindex = line.replace( tag, index_tag  )
        break
    # reformat Index "-x ..." -> "Option -x"
    if line.startswith( '@cindex @code{-x}' ):
      line = line.replace( '@cindex @code{-x}', '@cindex Option -x' )

    print( line, end='' )

    if cindex:
      print( cindex, end='' )
```

![image](https://github.com/avrdudes/avrdude/assets/12542359/c33c27be-fca0-41ee-bb2e-cb0419d5df85)
![image](https://github.com/avrdudes/avrdude/assets/12542359/0726554f-cfbd-4cb3-b0a2-96520144c90e)

Next steps should be an aligned wording, e.g. *Configuration File* vs. *Configuration Files*, renaming *Configuration Files* in the Windows section to *Windows Configuration Files*, unifying upper/lower case in the headers and removing section A.2.4 b/c this is n/a for Windows.

Ho-Ro:
Another small change populates the index even more with all programmer types, but maybe this is too much:

```diff
diff --git a/src/doc/programmers.cmake b/src/doc/programmers.cmake
index b8611c0..76c77ac 100644
[...]

PR #1592, opened 2023-12-10
Update avrdude.texi: more index entries; header uppercase; correct @xref

Ho-Ro:
As discussed in #1582


stefanrueger:
Neat! There is still scope for manual improvement, but those added automatically are a good addition. Thanks

mcuee:
> Neat! There is still scope for manual improvement, but those added automatically are a good addition. Thanks

Great. In that case, I think it is good to merge this as well before avrdude 7.3 release. We can always improve the documentation further in the future.

Issue #1591, opened 2023-12-08
Make programmer -c name checking more forgiving


MCUdude:
Avrdude supports a bunch of different programmers, and sometimes if can be a bit hard to remember what the correct name for it is. Some programmers have dashes in them, like `digilent-hs2`. Others have underscores, like `dragon_pp`, and other has none, like `stk600pp`.

There are no good pattern or rule to follow; the names are what they are, some with underscores, some without.
But what if Avrdude could be a little more forgiving, and "understand" that the user probably wants to use `stk600pp` when using `-cstk600_pp` or `-cstk600-pp`, and `atmelice_isp` when using `-catmeliceisp` or `-catmelice-isp`?

dl8dtl:
Well, it reminds me: back in those old days, sendmail also tried to be forgiving in slightly misspelled names.

As a sysadmin, someday I noticed that one of the pseudo accounts named "yp" (for "yellow pages", later renamed "network information services" by Sun) had a mail in its mailbox. It contained a private mail from one of our employees, obviously by no means intended for that pseudo account at all.

I was a bit surprised, until one other sysadmin got me the clue: sendmail tries to deliver misspelled addressees.

Elm (a famous MUA then), by default asked at the end: "Copies?". It was meant to give the user a chance to add a Cc addressee. That employee, however, assumed it were asking him whether he'd like to get a copy, and he answered "y".

Sendmail then found the pseudo user "yp" to be the closest match for an addressee named "y", and delivered …

Bottom line: better no automatism. However, giving the user a hint might be good, so if the user specifies `-cstk600_pp`, they can be hinted that `Did you mean -c stk600pp?`.

MCUdude:
> However, giving the user a hint might be good, so if the user specifies -cstk600_pp, they can be hinted Did you mean -c stk600pp?.

Yes, that's just as good.
Currently, Avrdude prints out all 140+ supported programmer options if the programmer name doesn't match perfectly. It would be great if Avrdude could provide a suggestion, and explicitly mention that -c? will print all compatible programmers.

stefanrueger:
[...]

PR #1593, opened 2023-12-10
Use Levenshtein distance to suggest a valid programmer name if -c is invalid

MCUdude:
See #1591 for details. The same for -p could also be done, but ideally in a separate PR.

Right now there is no weighing, but this can easily be added by tweaking the last four parameters in this line:

```c
int dist = levenshtein(programmer, id, 1, 1, 1, 1);
```

Examples:
```
$ ./avrdude -cusbast -patmega328p

avrdude error: cannot find programmer id usbast
similar matches:
usbasp = USBasp ISP and TPI programmer
use -c? to see all possible programmers for this part


$ ./avrdude -csdk500 -patmega328p

avrdude error: cannot find programmer id sdk500
similar matches:
stk500   = Atmel STK500 (probes STK500v2 first then STK500v1)
stk600   = Atmel STK600
stk500pp = Atmel STK500 v2 in parallel programming mode
stk500v1 = Atmel STK500 version 1.x firmware
stk500v2 = Atmel STK500 version 2.x firmware
use -c? to see all possible programmers for this part

$ ./avrdude -cstk600_pp -patmega328p

avrdude error: cannot find programmer id stk600_pp
similar matches:
stk600pp   = Atmel STK600 in parallel programming mode
stk500pp   = Atmel STK500 v2 in parallel programming mode
stk600     = Atmel STK600
stk600hvsp = Atmel STK600 in high-voltage serial programming mode
use -c? to see all possible programmers for this part
```

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -470,9 +470,9 @@ static int cmp_pgmid(const void *a, const void *b) {
 }
 
 static int suggest_programmers(const char *programmer, LISTID programmers) {
-  const int max_distance = 64; // Don't show suggestions if they are way far out
+  const int max_distance = 64;  // Don't show suggestions if they are way far out
 
-  size_t nid = 0;              // Number of possible programmer ids
+  int nid = 0;                  // Number of possible programmer ids
   for(LNODEID ln1 = lfirst(programmers); ln1; ln1 = lnext(ln1)) {
     PROGRAMMER *pgm = ldata(ln1);
     if(is_programmer(pgm))
@@ -483,7 +483,7 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
   pgm_distance *d = cfg_malloc(__func__, nid*sizeof*d);
 
   // Fill d[] struct
-  size_t idx = 0;
+  int idx = 0;
   AVRPART *p = locate_part(part_list, partdesc);
   for(LNODEID ln1 = lfirst(programmers); ln1; ln1 = lnext(ln1)) {
     PROGRAMMER *pgm = ldata(ln1);
@@ -500,7 +500,7 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
     }
   }
 
-  size_t n = 0, pgmid_maxlen = 0, comp = 0;
+  int n = 0, pgmid_maxlen = 0, comp = 0, len;
   if(nid) {                     // Sort list so programmers according to string distance
     qsort(d, nid, sizeof(*d), cmp_pgmid);
     size_t dst = d[nid > 2? 2: nid-1].dist; // Print at least 3 close suggestions if possible
@@ -509,8 +509,7 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
     for(; n < nid; n++) {
       if(d[n].dist > dst)
         break;
-      size_t len = strlen(d[n].pgmid);
-      if(len > pgmid_maxlen)
+      if((len = strlen(d[n].pgmid)) > pgmid_maxlen)
         pgmid_maxlen = len;
       if(d[n].common_modes)
         comp++;
@@ -518,7 +517,7 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
   }
   if(comp) {
     msg_info("similar programmer name%s:\n", str_plural(comp));
-    for(size_t i = 0; i < n; i++)
+    for(int i = 0; i < n; i++)
       if(d[i].common_modes)
         msg_info("  %-*s = %s\n", pgmid_maxlen, d[i].pgmid, d[i].desc);
   }
@@ -527,26 +526,54 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
 }
 
 static void programmer_not_found(const char *programmer, PROGRAMMER *pgm) {
-  msg_error("\n");
-  if(programmer && str_eq(programmer, "?")) {
-    msg_error("\nValid programmers are:\n");
+  msg_error("\v");
+  if(!programmer || !*programmer) {
+    pmsg_error("no programmer has been specified on the command line or in the\n");
+    imsg_error("config file(s); specify one using the -c option and try again\n");
+    return;
+  }
+
+  if(str_eq(programmer, "?")) {
+    msg_error("Valid programmers are:\n");
     list_programmers(stderr, "  ", programmers, ~0);
     msg_error("\n");
-  } else {
-    if(programmer && *programmer) {
-      if(!pgm || !pgm->id || !lsize(pgm->id)) {
-        pmsg_error("cannot find programmer id %s\n", programmer);
-        suggest_programmers(programmer, programmers);
-        msg_info("use -c? to see all possible programmers\n");
-      }
-      else
-        pmsg_error("programmer %s lacks %s setting\n", programmer,
-          !pgm->prog_modes? "prog_modes": !pgm->initpgm? "type": "some");
-    } else {
-      pmsg_error("no programmer has been specified on the command line or in the\n");
-      imsg_error("config file(s); specify one using the -c option and try again\n");
-    }
+    return;
   }
+
+  sort_programmers(programmers);
+
+  // If there were partial matches then they were not unique: count and list them
+  int pmatches = 0, maxlen = 0, len;
+  for(LNODEID ln1=lfirst(programmers); ln1; ln1=lnext(ln1)) {
+    PROGRAMMER *p = ldata(ln1);
+    if(is_programmer(p))
+      for(LNODEID ln2=lfirst(p->id); ln2; ln2=lnext(ln2)) {
+        const char *id = (const char *) ldata(ln2);
+        if(str_casestarts(id, programmer)) { // Partial initial match
+          pmatches++;
+          if((len = strlen(id)) > maxlen)
+            maxlen = len;
+        }
+      }
+  }
+  if(pmatches) {
+    msg_error("%s is not a unique start of a programmer name; consider:\n", programmer);
+    for(LNODEID ln1=lfirst(programmers); ln1; ln1=lnext(ln1)) {
+     PROGRAMMER *p = ldata(ln1);
+      if(is_programmer(p))
+        for(LNODEID ln2=lfirst(p->id); ln2; ln2=lnext(ln2)) {
+          const char *id = (const char *) ldata(ln2);
+          if(str_casestarts(id, programmer))
+            msg_error("  %-*s = %s\n", maxlen, id, p->desc);
+      }
+    }
+  } else if(!pgm || !pgm->id || !lsize(pgm->id)) {
+    pmsg_error("cannot find programmer id %s\n", programmer);
+    suggest_programmers(programmer, programmers);
+    msg_info("use -c? to see all possible programmers\n");
+  } else
+    pmsg_error("programmer %s lacks %s setting\n", programmer,
+      !pgm->prog_modes? "prog_modes": !pgm->initpgm? "type": "some");
 }
 
 static void part_not_found(const char *partdesc) {
@@ -1159,7 +1186,7 @@ int main(int argc, char * argv [])
   if(partdesc) {
     if(str_eq(partdesc, "?")) {
       if(pgmid && *pgmid && explicit_c) {
-        PROGRAMMER *pgm = locate_programmer_set(programmers, pgmid, &pgmid);
+        PROGRAMMER *pgm = locate_programmer_starts_set(programmers, pgmid, &pgmid);
         if(!pgm || !is_programmer(pgm)) {
           programmer_not_found(pgmid, pgm);
           exit(1);
@@ -1208,7 +1235,7 @@ int main(int argc, char * argv [])
     exit(1);
   }
 
-  pgm = locate_programmer_set(programmers, pgmid, &pgmid);
+  pgm = locate_programmer_starts_set(programmers, pgmid, &pgmid);
   if (pgm == NULL || !is_programmer(pgm)) {
     programmer_not_found(pgmid, pgm);
     exit(1);
--- a/src/main.c
+++ b/src/main.c
@@ -455,6 +455,7 @@ static int dev_opt(const char *str) {
 
 typedef struct {
   size_t dist;
+  int common_modes;
   const char *pgmid;
   const char *desc;
 } pgm_distance;
@@ -471,7 +472,7 @@ static int cmp_pgmid(const void *a, const void *b) {
 static int suggest_programmers(const char *programmer, LISTID programmers) {
   const int max_distance = 64; // Don't show suggestions if they are way far out
 
-  size_t nid = 0;               // Number of possible programmer ids
+  size_t nid = 0;              // Number of possible programmer ids
   for(LNODEID ln1 = lfirst(programmers); ln1; ln1 = lnext(ln1)) {
     PROGRAMMER *pgm = ldata(ln1);
     if(is_programmer(pgm))
@@ -483,6 +484,7 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
 
   // Fill d[] struct
   size_t idx = 0;
+  AVRPART *p = locate_part(part_list, partdesc);
   for(LNODEID ln1 = lfirst(programmers); ln1; ln1 = lnext(ln1)) {
     PROGRAMMER *pgm = ldata(ln1);
     if(!is_programmer(pgm))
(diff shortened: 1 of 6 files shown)


dl8dtl:
cool!

src/main.c
```diff
@@ -453,12 +453,73 @@ static int dev_opt(const char *str) {
...
+}
+
+static int suggest_programmers(const char *programmer, LISTID programmers) {
+  const int max_distance = 3;
+
+  int idx = 0;
+  int pgmid_maxlen = 0;
+  typedef struct {
+    int distance;
+    char *pgmid;
```
stefanrueger, line 467: I suspect if pgmid and desc are `const char *` these can be directly assigned below (no need for `cfg_strdup()` and no need for a later `free()`).

src/main.c
```diff
@@ -453,12 +453,73 @@ static int dev_opt(const char *str) {
...
+        d[idx].desc = cfg_strdup(__func__, pgm->desc);
+        idx++;
+        if(strlen(id) > (size_t)pgmid_maxlen)
+          pgmid_maxlen = (int)strlen(id);
+      }
+    }
+  }
+  // Sort list so programmers with the smallest distance gets printed first
+  qsort(d, idx, sizeof(*d), cmp_ptr);
+  if(idx) {
```
stefanrueger, line 500: Could show the closest match only and all those that have the same distance as closest match.

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
 
   return (char *) q;
 }
+
+/*
+ * This function implements the Damerau-Levenshtein algorithm to
```
stefanrueger, line 1036: where is this from? Acknowledge authors and check whether licence is compatible with AVRDUDE's

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+ * - del (as in "Deletion")
+ *
+ * Note that this algorithm calculates a distance _iff_ del == add.
+ */
+
+int levenshtein(const char *str1, const char *str2,
+  int swap, int subst, int add, int del) {
+
+  int i, j, len1 = strlen(str1), len2 = strlen(str2);
+  int *row0 = calloc(len2 + 1, sizeof(int*));
```
stefanrueger, line 1077: Must be `sizeof(*row0)` or `sizeof(int)`, not `sizeof(int *)`

Why not use `cfg_malloc(__func__, (len2+1)*sizeof*row0)` - this takes care of  unexpected "out of memory"

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+  int *row1 = calloc(len2 + 1, sizeof(int*));
```
stefanrueger, line 1078: same here and below

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+  int *row2 = calloc(len2 + 1, sizeof(int*));
+
+  for (j = 0; j <= len2; j++)
+    row1[j] = j * add;
```
stefanrueger, line 1082: adding an underscore or hyphen from str2 could have less penalty (say, add/4)

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+  for (i = 0; i < len1; i++) {
+    row2[0] = (i+1) * del;
+    for (j = 0; j < len2; j++) {
+      // Substitution
+      row2[j+1] = row1[j] + subst * (str1[i] != str2[j]);
```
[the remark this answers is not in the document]
stefanrueger: also could use much smaller penalty if both different character are only different by case

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+      // Swap
+      if (i > 0 && j > 0 && str1[i-1] == str2[j] &&
+        str1[i] == str2[j-1] && row2[j+1] > row0[j-1] + swap)
+        row2[j+1] = row0[j-1] + swap;
+      // Deletion
+      if (row2[j+1] > row1[j+1] + del)
+        row2[j+1] = row1[j+1] + del;
+      // Insertion
+      if (row2[j+1] > row2[j] + add)
```
stefanrueger, line 1096: if `str2[j] ` is hyphen or underscore then the cost of adding should be much less, say, `add/4` not `add`

src/strutil.c
```diff
@@ -1031,3 +1031,79 @@ char *str_nexttok(char *buf, const char *delim, char **next) {
...
+    for (j = 0; j < len2; j++) {
+      // Substitution
+      row2[j+1] = row1[j] + subst * (str1[i] != str2[j]);
+      // Swap
+      if (i > 0 && j > 0 && str1[i-1] == str2[j] &&
+        str1[i] == str2[j-1] && row2[j+1] > row0[j-1] + swap)
+        row2[j+1] = row0[j-1] + swap;
+      // Deletion
+      if (row2[j+1] > row1[j+1] + del)
+        row2[j+1] = row1[j+1] + del;
```
stefanrueger, line 1094: if `str2[j]` is hyphen or underscore then cost of deletion should be much less, say, `del/4`

src/strutil.c
MCUdude: Didn't think of that. Neat!

src/strutil.c
MCUdude: Forgot to add a URL to the source. It's from the git project repository. They use GPL 2 as well.
https://github.com/git/git/blob/master/levenshtein.c

src/main.c
MCUdude: Thanks! Fixed

src/main.c
```diff
@@ -453,12 +453,73 @@ static int dev_opt(const char *str) {
     !!strchr(str, '/');
 }
 
+static int cmp_ptr(const void *a, const void *b) {
+   return (*(int*)a - *(int*)b);
+}
+
+static int suggest_programmers(const char *programmer, LISTID programmers) {
+  const int max_distance = 3;
```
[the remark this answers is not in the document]
MCUdude: Great idea. I've rewritten `suggest_programmers` to do this instead.

src/main.c
MCUdude: Thanks, done!

src/strutil.c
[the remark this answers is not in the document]
MCUdude: > If both characters are different and both are hyphen or underscore could use much smaller penalty. If the characters are neighbours on the QUERTY keyboard could use smaller penalty.

> also could use much smaller penalty if both different characters are only different by case

It sounds like this would introduce significant changes to the algorithm, and I have no idea how to implement this. If you have suggestions to how this can be implemented, I would be grateful if you could share a code suggestion.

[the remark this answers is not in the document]
stefanrueger: OK, I've copied and modified the levenshtein() function to provide plausible typo weighting. Also suggested a way to apply this routine in the context.

stefanrueger:
Weighted Damerau-Levenshtein implementation needs testing. Here one example:

```
$ avrdude -c sdk499 -p m328

avrdude error: cannot find programmer id sdk499
similar programmer name:
  stk500 = Atmel STK500 (probes STK500v2 first then STK500v1)
use -c? to see all possible programmers
```

stefanrueger:
> only show valid programmers for a particular part?

The part is not necessarily known, but yes, if you find a way of doing this might be worth our while.

MCUdude:
I've now pushed a possible fix to prevent printing non-compatible programmers.

I've also made it so that the incredible long list of programmers only is only printed if the user uses `-c?`:

```
$ ./avrdude -cpickit4_i -patmega328

avrdude error: cannot find programmer id pickit4_i
similar programmer names:
  pickit2     = MicroChip's PICkit2 Programmer
  pickit4_isp = MPLAB(R) PICkit 4 in ISP mode
use -c? to see all possible programmers
$ ./avrdude -cdoesnotexist -patmega328

avrdude error: cannot find programmer id doesnotexist
use -c? to see all possible programmers
$ ./avrdude -c? -patmega328

Valid programmers for part ATmega328 are:
  2232hio           = 2232hio based on FT2232H with buffer and LEDs via ISP
  4232h             = FT4232H based generic programmer via ISP
 [...]
  xplainedmini      = Atmel AVR XplainedMini via ISP
  xplainedmini_isp  = Atmel AVR XplainedMini via ISP
  xplainedmini_dw   = Atmel AVR XplainedMini via debugWIRE
```

MCUdude:
@stefanrueger please review the last commits I pushed a few days ago.

As mentioned earlier, we should probably do the same Levenshtein checking for ``-p`` as well, but it might also make sense for ``-U`` memory names.

src/main.c
```diff
@@ -492,12 +494,14 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
         d[idx].pgmid = ldata(ln2);
         d[idx].desc = pgm->desc;
         d[idx].dist = str_weighted_damerau_levenshtein(d[idx].pgmid, programmer);
+        if(p)
+          d[idx].common_modes = p->prog_modes & pgm->prog_modes;
```
stefanrueger, line 498: If we don't have a valid part, it's better to replace the two lines with
```
        d[idx].common_modes = pgm->prog_modes & (p? p->prog_modes: -1);
 ```
Some programmers don't need the part be specified as they can figure out which part is connected.

File: src/main.c [hunk 12]
```diff
@@ -509,12 +513,15 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
       size_t len = strlen(d[n].pgmid);
       if(len > pgmid_maxlen)
         pgmid_maxlen = len;
+      if(d[n].common_modes)
+        comp++;
     }
   }
-  if(n) {
+  if(comp) {
```
stefanrueger, line 520: next line should be `str_plural(comp)` not `str_plural(n)`

src/main.c
```diff
@@ -522,32 +522,31 @@ static int suggest_programmers(const char *programmer, LISTID programmers) {
...
-    else
-      pmsg_error("programmer %s lacks %s setting\n", programmer,
-        !pgm->prog_modes? "prog_modes": !pgm->initpgm? "type": "some");
-  } else {
-    pmsg_error("no programmer has been specified on the command line or in the\n");
-    imsg_error("config file(s); specify one using the -c option and try again\n");
-  }
-
-  if(pgm_suggestions <= 0) {
+  if(str_eq(programmer, "?")) {
```
stefanrueger, line 532: programmer might be NULL. Guard  against this

stefanrueger:
@MCUdude Good ideas throughout. I am glad you raised the issue with the long programmer lists; this now looks like a useful way to deal with them.

As to symmetrising the `-c`, `-p` and `-U` treatment of names, I wonder whether we first should make `-c` accept partial names, eg, I suspect `-c ur` would already address `urclock` uniquely in the same way `ee` is a unique start for `eeprom`. 

The other thing is that I am keen to freeze features after this batch of PRs; this would introduce the current typo checking using the weighted Damerau-Levenshtein distance as an *experimental* feature in v7.3 that we could build on after the release...

MCUdude:
>As to symmetrising the -c, -p and -U treatment of names, I wonder whether we first should make -c accept partial names, eg, I suspect -c ur would already address urclock uniquely in the same way ee is a unique start for eeprom.

This would be very neat. Are you thinking before or after the 7.3 release?

> The other thing is that I am keen to freeze features after this batch of PRs; this would introduce the current typo checking using the weighted Damerau-Levenshtein distance as an experimental feature in v7.3 that we could build on after the release...

I agree. Let's finish this PR, and continue adding -p and -U fo the next (7.4?) release.

src/main.c
MCUdude: Like this?
`if(programer && str_eq(programmer, "?"))`

MCUdude:
Another neat thing is that -c? -p will only print compatible programmers for _that_ particular part, while -c? -p? will print all programmers.

src/main.c
```diff
@@ -453,23 +453,100 @@ static int dev_opt(const char *str) {
...
+  size_t n = 0, pgmid_maxlen = 0, comp = 0;
+  if(nid) {                     // Sort list so programmers according to string distance
+    qsort(d, nid, sizeof(*d), cmp_pgmid);
+    size_t dst = d[nid > 2? 2: nid-1].dist; // Print at least 3 close suggestions if possible
+    if(dst > max_distance)
+      dst = max_distance;
+    for(; n < nid; n++) {
+      if(d[n].dist > dst)
+        break;
+      size_t len = strlen(d[n].pgmid);
```
stefanrueger, line 512: Small point: maxlen should not be affected by programmers that we are no longer considering to print.

stefanrueger:
OK, I have now gone through the PR and it looks good. And added partial initial match of a programmer, if unique. 

Needs to be looked at and tested, though.

mcuee:
I should be able to give this PR some simple tests over the weekend.

MCUdude:
Seems to work great! The only thing is that when a programmer isn't unique, it suggest incompatible programmers as well. Is this what we want?

```
$ ./avrdude -cusbt -patmega1284p
avrdude error: cannot find USBtiny device (0x1781/0xc9f)
avrdude error: unable to open port usb for programmer usbtiny

avrdude done.  Thank you.

$ ./avrdude -csnap_ -patmega1284p
snap_ is not a unique start of a programmer name; consider:
  snap_jtag = MPLAB(R) SNAP in JTAG mode
  snap_isp  = MPLAB(R) SNAP in ISP mode
  snap_pdi  = MPLAB(R) SNAP in PDI mode
  snap_tpi  = MPLAB(R) SNAP in TPI mode
  snap_updi = MPLAB(R) SNAP in UPDI mode
```

stefanrueger:
> suggest incompatible programmers as well

Yes, the commit was cutting corners (and hoped nobody would notice :open_mouth:) - fixed:

```
$ avrdude -c dr -p t11
dr is not a unique start of a programmer name; consider:
  dragon_hvsp = Atmel AVR Dragon in HVSP mode
  dryrun      = Emulates programming without a programmer

$ avrdude -c dr -p t13
dr is not a unique start of a programmer name; consider:
  dragon_dw   = Atmel AVR Dragon in debugWire mode
  dragon_hvsp = Atmel AVR Dragon in HVSP mode
  dragon_isp  = Atmel AVR Dragon in ISP mode
  dryrun      = Emulates programming without a programmer

$ avrdude -c dr
dr is not a unique start of a programmer name; consider:
  dragon_dw   = Atmel AVR Dragon in debugWire mode
  dragon_hvsp = Atmel AVR Dragon in HVSP mode
  dragon_isp  = Atmel AVR Dragon in ISP mode
  dragon_jtag = Atmel AVR Dragon in JTAG mode
  dragon_pdi  = Atmel AVR Dragon in PDI mode
  dragon_pp   = Atmel AVR Dragon in PP mode
  dryrun      = Emulates programming without a programmer
```

mcuee:
This PR seems to work fine now.

The last one is a bit strange though as I do not see any similarities between `ser` and `xbee`.

```
P> .\avrdude_pr1593 -C .\avrdude_pr1593.conf -c dr -p m328pb -t
dr is not a unique start of a programmer name; consider:
  dragon_dw  = Atmel AVR Dragon in debugWire mode
  dragon_isp = Atmel AVR Dragon in ISP mode
  dragon_pp  = Atmel AVR Dragon in PP mode
  dryrun     = Emulates programming without a programmer

PS>.\avrdude_pr1593 -C .\avrdude_pr1593.conf -c usb -p m328pb -t
usb is not a unique start of a programmer name; consider:
  usbasp       = USBasp ISP and TPI programmer
  usbasp-clone = Any usbasp clone with correct VID/PID
  usbtiny      = USBtiny simple USB programmer
PS>.\avrdude_pr1593 -C .\avrdude_pr1593.conf -c usba -p m328pb -t
usba is not a unique start of a programmer name; consider:
  usbasp       = USBasp ISP and TPI programmer
  usbasp-clone = Any usbasp clone with correct VID/PID

PS>.\avrdude_pr1593 -C .\avrdude_pr1593.conf -c pickit -p m328pb -t
pickit is not a unique start of a programmer name; consider:
  pickit2     = Microchip PICkit 2 programmer in ISP mode
  pickit4_isp = MPLAB(R) PICkit 4 in ISP mode

PS>.\avrdude_pr1593 -C .\avrdude_pr1593.conf -c ser -p m328p -t
avrdude_pr1593 error: cannot find programmer id ser
similar programmer name:
  xbee = XBee for Series 2 Over-The-Air (XBeeBoot) bootloader using STK500 v1 protocol
use -c? to see all possible programmers
```

stefanrueger:
> I do not see any similarities between ser and xbee

Ahhh: replace s with x (small penalty as s and x are neighbours on the keyboard), add b (normal penalty), observe e is correct, and then replace r with e (again small penalty being keyboard neighbours)

With the same reasoning we have
```
$ avrdude -c yeckivk
avrdude error: cannot find programmer id yeckivk
similar programmer name:
  urclock = Urclock programmer for urboot bootloaders using urprotocol
```

The weighted Damerau-Levenshtein distance can tell if the text is typed with the hands shifted by one key. There is a limit, though, of what AVRDUDE accepts as similar (tops out at 5 close keys on keyboards or 2 normal errors such as insertion deletion or replacement of unrelated chars)

mcuee:
@stefanrueger 

Thanks for the explanation. I think this PR is good to go.

PR #1596, opened 2023-12-18
Use -xmode to switch PICkit4 and SNAP application modes

MCUdude:
Resolves issue #1027

If you have a PICkit4 or an MPLAB SNAP that happens to be in PIC mode, it has to be switched to AVR mode before Avrdude can communicate with it. However, with this PR, Avrdude can now do the switching. Use `-xmode=avr` to perform the switch. This is important because it has been reported over at the Avrfreaks forums that the latest MPLAB X doesn't do the switch when programming AVRs. It stays in "PIC mode" and speaks a protocol similar (or identical) to the PICkit5, which Avrdude currently doesn't support.

TODO:
* Implement `-xmode=pic` to [switch back to PIC mode](https://github.com/microchip-pic-avr-tools/pyedbglib/blob/main/pyedbglib/protocols/housekeepingprotocol.py#L104).
* Implement switching for `pickit4_isp` and `snap_isp`

```

avrdude error: MPLAB SNAP in PIC mode detected
        use -xmode=avr to enter AVR mode

avrdude error: unable to open port usb for programmer snap_updi

avrdude done.  Thank you.

$ ./avrdude -csnap_updi -patmega4808 -xmode=avr

avrdude error: MPLAB SNAP in PIC mode detected
        switching to AVR mode
        please re-run Avrdude

avrdude error: unable to open port usb for programmer snap_updi

avrdude done.  Thank you.

$ ./avrdude -csnap_updi -patmega4808 -xmode=avr
avrdude warning: programmer is already in AVR mode. Ignoring -xmode
avrdude: device signature = 0x1e9650 (probably m4808)

avrdude done.  Thank you.
```


MCUdude:
It looks like it isn't going to be all that easy to get mode switching to work for `pickit4_isp` and `snap_isp`.
This is because the variable that "tells" avrdude to switch mode is local to stk500v2.c, and can't easily be copied into `jtag3_open_common`, which lives in jtag3.c. @stefanrueger If you have an idea of how this can be done in a way that isn't terribly ugly, I'm all ears!

mcuee:
@MCUdude 

One way is to ship this as a separate utility along with avrdude without integrating into avrdude, if it is a bit difficult to integrate into existing avrdude source codes.

Let's also hear from @stefanrueger and @dl8dtl.

MCUdude:
@mcuee try this project:
[Set_PIC_mode.X.zip](https://github.com/avrdudes/avrdude/files/13717960/Set_PIC_mode.X.zip)

When the project is loaded you can click the green arrows to switch mode:

Skjermbilde 2023-12-19 kl  17 16 44

mcuee:
Somehow the switching failed for my PICKit 4.

```
 *****************************************************

 
The tool was disconnected while it was busy.
Switcher failed: getPid
```

mcuee:
Same for SNAP.

```
*****************************************************

 

Snap timeout during the firmware download process.
Connection Failed.

```

Looks like I need to try out in another computer.

MCUdude:
@mcuee I managed to implement `-xmode=pic`. It did use a different "data packet" than what [the command](https://github.com/microchip-pic-avr-tools/pyedbglib/blob/main/pyedbglib/protocols/housekeepingprotocol.py#L104) suggested by @xedbg earlier, but I managed to track down the correct package using Wireshark USB capture 🤓

Skjermbilde 2023-12-20 kl  22 00 09

MCUdude:
@stefanrueger I'd like to hear what you would consider the neatest way to trigger the PIC/AVR mode switching from stk500v2.c. Now that the base functionality of this PR is up and running, this would be the next step.

> It looks like it isn't going to be all that easy to get mode switching to work for pickit4_isp and snap_isp.
This is because the variable that "tells" avrdude to switch mode is local to stk500v2.c, and can't easily be copied into jtag3_open_common, which lives in jtag3.c. @stefanrueger If you have an idea of how this can be done in a way that isn't terribly ugly, I'm all ears!

stefanrueger:
It isn't clear to me why a user option is needed.

If the user requests an operation in these programmers for which they need to be in AVR mode, would it not make sense to query which mode they are in, memorise the mode in an appropriate `PDATA(pgm)->...` component, switch to (presumably) AVR mode, do the operation(s) and then in the end switch to the memorised mode the programmer was in, if needed.

I am sure I have overlooked (or don't know) a crucial problem on the way.

mcuee:
@MCUdude 

I will try this PR later this evening. Now I know how to switch to PIC mode using MPLAB X IDE 6.15. And your [libusb example code](https://github.com/avrdudes/avrdude/issues/1027#issuecomment-1861665783) works fine to switch from PIC mode to AVR mode.

mcuee:
@MCUdude 

You mentioned that PICKit 4 and SNAP can not program the AVR EB chips in AVR mode and MPLAB X IDE uses PIC mode for that purpose. Is this confirmed? 

If that is the case, we may have to block them and print a warning message for now.

MCUdude:
> You mentioned that PICKit 4 and SNAP can not program the AVR EB chips in AVR mode and MPLAB X IDE uses PIC mode for that purpose. Is this confirmed?

@mcuee read this post. It appears that PK4/SNAP can program AVR-EBs, but only in PIC mode using MPLAB X

https://www.avrfreaks.net/s/topic/a5C3l000000UkyzEAC/t193040

> If that is the case, we may have to block them and print a warning message for now.

Ideally, yes. But it _may_ just be a firmware update away from working in AVR mode as well.

MCUdude:
> It isn't clear to me why a user option is needed.

If the user requests an operation in these programmers for which they need to be in AVR mode, would it not make sense to query which mode they are in, memorise the mode in an appropriate PDATA(pgm)->... component, switch to (presumably) AVR mode, do the operation(s) and then in the end switch to the memorised mode the programmer was in, if needed.
I am sure I have overlooked (or don't know) a crucial problem on the way.

@stefanrueger There are good reasons why one wants to switch modes explicitly. First, this is a rather niche issue that only deals with two official programmers, and one of them has been discontinued.

After a switch, the programmer restarts and reenumerates. Avrdude would have to start over again after the switch,  and this is something I don't want to deal with for now; hence returning -1 after the switch.

And if the programmer is in PIC mode, there are good reasons why you want to keep it in AVR mode. It's quicker, since the programmer doesn't have to switch modes all the time, and you can use it as a USB to serial adapter because it shows up as a serial port. But if you borrowed someone's programmer that originally was in PIC mode, you would probably like to switch it back before returning it. And having to install MPLAB plus a PIC compiler and create an empty project just to switch back is just dumb. So it's IMO much better to have -xmode where the user can do whatever he or she wants.

And I forgot to mention. MPLAB really struggles with switching modes sometimes. Sometimes it works, while others you have to restart MPLAB to get it to do the switch. Just having a utility that can do this that's _not_ MPLAB is a good thing.

stefanrueger:
> After a switch, the programmer restarts and reenumerates. Avrdude would have to start over again after the switch

Could that be packed into the programmer's open() function? Don't we now have code that can figure out when a device re-enumerates? And the switching back to PIC could be put into the close function(), not?

Having an option makes it necessary to document it; users won't know the option exists and wouldn't know the procedures, so AVRDUDE would need to tell them that the programmer is in pic mode and they need to call avrdude again with some option and then again, to do the work and then again to put it into PIC mode. And somehow it would be much slicker if AVRDUDE did all that on its own accord...

MCUdude:
> Could that be packed into the programmer's open() function? 

Yes, feel free! I'm leaving for Christmas holiday tomorrow and will be away for a little more than a week. I'm planning to bring my computer + some hardware, but I have limited time to spend on this PR. However, I can see if I can spend some time in the evenings to do a little testing with the SNAP and PICkit4.

Automatic switching would be nice, but I think it's important to also be able to explicitly set a mode, so the user can utilize features like the USB-UART interface the programmer provides in AVR mode.

mcuee:
> And I forgot to mention. MPLAB really struggles with switching modes sometimes. Sometimes it works, while others you have to restart MPLAB to get it to do the switch. Just having a utility that can do this that's not MPLAB is a good thing.

I will agree with this one. MPLAB X IDE is really troublesome to use for switching the mode.

MCUdude:
Another thing I should look into is to see how the -xmode performs when the SNAP or PICkit4 is in bootloader mode. When in bootloader mode, they have a different PID, so it's easy to detect. However, I doubt -xmode will work, because I'm pretty sure they would require a different command. But this _can_ be fixed in a separate PR. If the programmer is in bootloader mode, it means that it's either in FW upgrade mode or something is wrong with it.

mcuee:
Mode switching (PIC <->AVR) is working well for Microchip SNAP as well.

```
PS> .\avrdude_pr1596 -C .\avrdude_pr1596.conf -c snap -p m32a
avrdude_pr1596: AVR device initialized and ready to accept instructions
avrdude_pr1596: device signature = 0x1e9502 (probably m32a)

avrdude_pr1596 done.  Thank you.

PS> .\avrdude_pr1596 -C .\avrdude_pr1596.conf -c snap -p m32a -xmode=pic
               switching to PIC mode
               PIC mode switch successful
avrdude_pr1596 error: unable to open port usb for programmer snap

avrdude_pr1596 done.  Thank you.

PS> .\avrdude_pr1596 -C .\avrdude_pr1596.conf -c snap -p m32a

avrdude_pr1596 error: MPLAB SNAP in PIC mode detected
               use -xmode=avr to enter AVR mode

avrdude_pr1596 error: unable to open port usb for programmer snap

avrdude_pr1596 done.  Thank you.

PS> .\avrdude_pr1596 -C .\avrdude_pr1596.conf -c snap -p m32a -xmode=avr

avrdude_pr1596 error: MPLAB SNAP in PIC mode detected
               switching to AVR mode
               please re-run Avrdude

avrdude_pr1596 error: unable to open port usb for programmer snap

avrdude_pr1596 done.  Thank you.

PS> .\avrdude_pr1596 -C .\avrdude_pr1596.conf -c snap -p m32a
avrdude_pr1596: AVR device initialized and ready to accept instructions
avrdude_pr1596: device signature = 0x1e9502 (probably m32a)

avrdude_pr1596 done.  Thank you.
```

stefanrueger:
OK, looks like a good enhancement. Thanks for tackling this problem, @MCUdude!

MCUdude:
Time is still very limited, but I think this PR is soon ready for review.
Things I'd like to get/see fixed:

* Modes cannot be switched using `-cpickit4_isp` and `-csnap_isp`. The problem is that you can't just pass private `PDATA(pgm)` data from stk500v2.c/h to jtag3.c/h, which is necessary to toggle a mode switch. @stefanrueger I don't want to make a full copy of `jtag3_open_common` in stk500v2.c with tiny modifications, but I don't know how it could be done differently without adding additional parameters to the function. And that wouldn't make it "common" anymore, wouldn't it?
* After a switch attempt, Avrdude always prints `avrdude error: unable to open port usb for programmer snap(/pickit4)`. It would be great if it could skip this like and just print the `Avrdude done, thank you` message instead.

MCUdude:
The last commit adds an additional parameter to `jtag3_open_common`, which makes it possible to perform a SNAP/PK4 mode switch form within stk500v2.c. It's not all that elegant, but I couldn't come up with a better solution, without copying the entire contents of `jtag3_open_common` into stk500v2.c.

src/jtag3.c
```diff
@@ -1597,6 +1600,27 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
       }
     }
 
+    else if (str_starts(extended_param, "mode=") &&
```
stefanrueger, line 1603: There should be an error message if -xmode=... is used but the programmer is neither snap nor pickit4.

src/jtag3.c
```diff
@@ -1597,6 +1600,27 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
+      (str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))) {
+      char mode[3];
```
stefanrueger, line 1605: Should that not be `mode[4]` and `mode[3] = 0` set (otherwise how can `mode` ever be `str_caseeq()` with the *4*-character `"avr"`?

src/jtag3.c
```diff
@@ -1597,6 +1600,27 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
+      int sscanf_success = sscanf(extended_param, "mode=%3s", mode);
+      if (sscanf_success < 1 || (!str_caseeq(mode, "avr") && !str_caseeq(mode, "pic"))) {
+        pmsg_error("invalid mode setting '%s'\n", extended_param);
```
stefanrueger, line 1608: ... but `-xmode=avr_rocks` would pass ...

All in all probably better and shorter to `str_caseeq(extended_param, "mode=avr")` for one mode (no sscanf needed) and `str_caseeq(extended_param, "mode=pic")` for the other mode and complain otherwise.

src/jtag3.c
```diff
@@ -1685,44 +1707,48 @@ int jtag3_open_common(PROGRAMMER *pgm, const char *port) {
...
+          // Retry with bootloader USB PID
+          pinfo.usbinfo.pid = bl_pid;
           pic_mode = serial_open(port, pinfo, &pgm->fd);
         }
         if(pic_mode >= 0) {
           msg_error("\n");
-          pmsg_error("PICkit4 in PIC mode detected!\n");
-          imsg_error("Use MPLAB X or Microchip Studio to switch to AVR mode\n\n");
-          return -1;
+          pmsg_error("%s in %s mode detected\n",
```
stefanrueger, line 1730: I don't understand how `serial_open(port, pinfo, &pgm->fd) >= 0` signifies that the programmer is in PIC mode. Could there not be other reasons why `serial_open()` returns non-negative?

src/jtag3.c
```diff
@@ -1685,44 +1707,48 @@ int jtag3_open_common(PROGRAMMER *pgm, const char *port) {
...
+            if(pinfo.usbinfo.pid == bl_pid)
+              serial_send(&pgm->fd, exit_bl_cmd, sizeof(exit_bl_cmd));
+            else {
+              serial_send(&pgm->fd, enter_avr_mode_cmd, sizeof(enter_avr_mode_cmd));
+              usleep(250*1000);
+              serial_send(&pgm->fd, reset_cmd, sizeof(reset_cmd));
+            }
+            imsg_error("please run Avrdude again to continue the session\n\n");
+          } else
+            imsg_error("use -xmode=avr to enter AVR mode\n\n");
```
stefanrueger, line 1743: The error message could be more helpful and construct the full command line `"to switch into AVR mode try avrdude -c%s. -p%s -P%s -xmode=avr\n"`

src/jtag3.c
```diff
@@ -1685,44 +1707,48 @@ int jtag3_open_common(PROGRAMMER *pgm, const char *port) {
...
+          return LIBAVRDUDE_SOFTFAIL;
```
stefanrueger, line 1744: Why return with SOFTFAIL? Is the caller able to do sth with that?
 
The impression I got is that after `-xmode=avr` or `-xmode=pic` AVRDUDE should exit. Why not *document* that so (`-xmode=avr` or `-xmode=pic` switches the PICkit4 or Snap programmers into the  corresponding mode and exits) and `exit(0)`?

Come to think of this, why not create a function `switch_mode()` or some such that switches modes and call this function when the user issues `-xmode=...` This way the switching code does not have to reside in `open()` as there is no intention to open and continue. That also might(!) not require a new parameter in the `common_open()` function.

src/jtag3.c
```diff
@@ -1749,11 +1777,21 @@ int jtag3_open_common(PROGRAMMER *pgm, const char *port) {
...
+  // Switch from AVR to PIC mode
+  if (mode_switch == PK4_SNAP_MODE_PIC) {
+    imsg_error("switching to PIC mode\n");
+    unsigned char *resp, buf[] = {SCOPE_GENERAL, CMD3_FW_UPGRADE, 0x00, 0x00, 0x70, 0x6d, 0x6a};
+    if (jtag3_command(pgm, buf, sizeof(buf), &resp, "enter PIC mode") < 0) {
+      imsg_error("entering PIC mode failed\n");
+      return -1;
+    }
+    imsg_error("PIC mode switch successful\n");
+    return LIBAVRDUDE_SOFTFAIL;
```
stefanrueger, line 1792: See above re returning

src/stk500v2.c
```diff
@@ -2068,6 +2068,27 @@ static int stk500v2_jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extp
       }
     }
 
+    else if (str_starts(extended_param, "mode=") &&
+      (str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))) {
+      char mode[3];
```
stefanrueger, line 2073: See above comments

src/jtag3.c
MCUdude: It is! See the next line.

$ ./avrdude -catmelice -patmega1284p -xmode=avr
avrdude error: invalid extended parameter 'mode=avr'
avrdude error: unable to parse extended parameter list

``` c
    else if (str_starts(extended_param, "mode=") &&
      (str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))) {
```

src/jtag3.c
MCUdude: Great idea!

src/jtag3.c
MCUdude: It will return non-negative if it's able to connect. And we know because `pinfo.usbinfo.pid` is set to the "PIC mode" USB PID.

So we get a non-zero value because it's able to connect to a device in PIC mode.

src/jtag3.c
MCUdude: Good idea!

src/jtag3.c
MCUdude: OK. Returning -3 instead.

src/jtag3.c
MCUdude: > Why not document that 

I'm using `exit(0)` now. Do you mean documenting it in the docs?

> Come to think of this, why not create a function switch_mode()

Well, it _could_ be done so that `jtag3_open_common` returned known values if in PIC or bootloader mode, and did the switching outside. That's probably the cleanest solution. I'll look into it.

MCUdude: I spent some time trying to mode the switching into a dedicated function so that `jtag3_open_common` can be left untouched. It turned out to be more complicated than I first thought, and I don't think it's worth my time. The current solution works really well and is easy to understand and follow (IMO).

src/jtag3.c
MCUdude: > Could there not be other reasons why serial_open() returns non-negative?

No, at least when using libusb or hidapi, it will return -1 if there is an error, and 0 if OK.

src/avrdude.1
```diff
@@ -1535,6 +1535,19 @@ alone.
 .It Ar help
 Show help menu and exit.
 .El
+.It Ar PICkit 4
+.It Ar MPLAB SNAP
+.Bl -tag -offset indent -width indent
+.It Ar mode=avr,pic
+The PICkit 4 and MPLAB SNAP programmer can only be utilized by Avrdude when in
```
stefanrueger, line 1542: Should document that avrdude terminates after `-xmode=avr` or `-xmode=pic`. COuld say
```
Switch programmer to AVR or PIC mode, then exit: the PICkit 4 and MPLAB SNAP programmer can only be utilised by
.Nm
when in AVR mode.
```

src/doc/avrdude.texi
```diff
@@ -1146,6 +1146,21 @@ The current set-voltage can be read by @samp{-xvtarg} alone.
...
 
+@cindex Option @code{-x} PICkit 4
+@cindex Option @code{-x} MPLAB SNAP
+@item PICkit 4
+@item MPLAB SNAP
+
+The PICkit 4 and MPLAB SNAP programmers accept the following extended parameters:
+@table @code
+@item @samp{mode=avr,pic}
+The PICkit 4 and MPLAB SNAP programmer can only be utilized by Avrdude when in
```
stefanrueger, line 1157: See above

src/jtag3.c
stefanrueger: OK, I see: `-xmode=...` is only accepted for snap and pickit4, indeed.

src/jtag3.c
```diff
@@ -1597,6 +1600,25 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
 
+    else if (str_starts(extended_param, "mode=") &&
+      (str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))) {
+      // Flag a switch to AVR mode
+      if (str_caseeq(extended_param, "mode=avr")) {
+        PDATA(pgm)->pk4_snap_mode = PK4_SNAP_MODE_AVR;
+        continue;
+      }
+      // Flag a switch to PIC mode
+      else if (str_caseeq(extended_param, "mode=pic")) {
```
stefanrueger, line 1611: Not great practice to else a continue. It's not wrong but I find it confusing.

src/jtag3.c
```diff
@@ -1597,6 +1600,25 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
+        PDATA(pgm)->pk4_snap_mode = PK4_SNAP_MODE_PIC;
+        continue;
+      }
+      else {
```
stefanrueger, line 1615: Don't else a continue. Below can be simplified to 
``` diff
diff --git a/src/jtag3.c b/src/jtag3.c
index 33bec590..e4af8cfc 100644
--- a/src/jtag3.c
+++ b/src/jtag3.c
@@ -1612,11 +1612,9 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
         PDATA(pgm)->pk4_snap_mode = PK4_SNAP_MODE_PIC;
         continue;
       }
-      else {
-        pmsg_error("invalid mode setting '%s'. Use -xmode=avr or -xmode=pic\n", extended_param);
-        rv = -1;
-        break;
-      }
+      pmsg_error("invalid mode setting '%s'. Use -xmode=avr or -xmode=pic\n", extended_param);
+      rv = -1;
+      break;
     }
 
     else if (str_eq(extended_param, "help")) {
```

Actually. the else that follows the break doesn't need to be there as well.

src/jtag3.c
```diff
@@ -1613,6 +1635,9 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
         msg_error("  -xvtarg                 Read on-board target supply voltage\n");
         msg_error("  -xvtarg=<arg>           Set on-board target supply voltage\n");
       }
+      if(str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap")) {
+        msg_error("  -xmode=avr|pic          Set programmer to AVR or PIC mode\n");
```
stefanrueger, line 1639: Might want to say Set programmer to AVR or PIC mode, *then exit*.

src/jtag3.c
stefanrueger: OK, understood now. Thanks for expanation

src/jtag3.c
```diff
@@ -1624,7 +1649,7 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
   return rv;
 }
 
-int jtag3_open_common(PROGRAMMER *pgm, const char *port) {
+int jtag3_open_common(PROGRAMMER *pgm, const char *port, const int mode_switch) {
```
stefanrueger, line 1652: Please remove the const in const int mode_switch. It serves little purpose: the compiler passes a *copy* of the int argument to the function, so the caller does not need to worry what the function does with that copy. Some style guides discourage the use of const for call-by-value variables, and AVRDUDE is not in the habit of consting int aruments, so I wouldn't start that here.

The other const in const char *port is important, though, as it warrants that the function does not try to modify what port points at (call by reference).

src/jtag3.h
```diff
@@ -25,7 +25,7 @@
 extern "C" {
 #endif
 
-int  jtag3_open_common(PROGRAMMER *pgm, const char *port);
+int  jtag3_open_common(PROGRAMMER *pgm, const char *port, const int mode_switch);
```
stefanrueger, line 28: const int mode_switch -> int mode_switch

src/stk500v2.c
```diff
@@ -2068,6 +2068,25 @@ static int stk500v2_jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extp
...
 
+    else if (str_starts(extended_param, "mode=") &&
+      (str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))) {
+      // Flag a switch to AVR mode
+      if (str_caseeq(extended_param, "mode=avr")) {
+        PDATA(pgm)->pk4_snap_mode = PK4_SNAP_MODE_AVR;
+        continue;
+      }
+      // Flag a switch to PIC mode
+      else if (str_caseeq(extended_param, "mode=pic")) {
```
stefanrueger, line 2079: remove else

src/stk500v2.c
```diff
@@ -2068,6 +2068,25 @@ static int stk500v2_jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extp
...
+        PDATA(pgm)->pk4_snap_mode = PK4_SNAP_MODE_PIC;
+        continue;
+      }
+      else {
```
stefanrueger, line 2083: remove else { and }

src/stk500v2.c
```diff
@@ -2080,6 +2099,8 @@ static int stk500v2_jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extp
         msg_error("  -xvtarg               Read on-board target supply voltage\n");
         msg_error("  -xvtarg=<arg>         Set on-board target supply voltage\n");
       }
+      if(str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))
+        msg_error("  -xmode=avr|pic        Set programmer to AVR or PIC mode\n");
```
stefanrueger, line 2103: add comma then exit

MCUdude: Comma? like `Set programmer to AVR, or PIC mode\n`?

It already exits after printing -xhelp:

```c
      if(str_starts(pgmid, "pickit4") || str_starts(pgmid, "snap"))
        msg_error("  -xmode=avr|pic        Set programmer to AVR or PIC mode\n");
      msg_error  ("  -xhelp                Show this help menu and exit\n");
      exit(0);
```

MCUdude: Nevermind. I started looking at the review from the bottom of this page. I'll assume this is what you wanted:

```c
msg_error("  -xmode=avr|pic        Set programmer to AVR or PIC mode, then exit\n");
```

src/jtag3.c
MCUdude: Thanks for the tip!

src/jtag3.c
MCUdude: Good idea!

MCUdude:
Thanks for another review!

src/jtag3.c
```diff
@@ -1585,18 +1585,18 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
           PDATA(pgm)->vtarg_set = true;
-          continue;
         }
         // Get target voltage
-        else if(str_eq(extended_param, "vtarg")) {
+        else if(str_eq(extended_param, "vtarg"))
           PDATA(pgm)->vtarg_get = true;
-          continue;
-        }
+        else
```
stefanrueger, line 1597: I am not convinced about this flow of control. Should further -x... processing be stopped through this `break` and `rv` returned (whatever it might be at this point?) 

Generally, I find this parsing function has grown over time (not the fault of this PR) into something that is hard to read (but this PR is the straw that breaks the camel's back).

One thing that would make it less hard to read is to remove the `else` after `continue` or `break`.

Another thing: When the code `break`s out of the loop it should set `rv = -1` before and print an error message. `break` should only be used for errors (as it leaves potentially other options unparsed)

Both measures would make this *a lot* more readable

src/jtag3.c
```diff
@@ -1585,18 +1585,18 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
           int sscanf_success = sscanf(extended_param, "vtarg=%lf", &vtarg_set_val);
           PDATA(pgm)->vtarg_data = (double)((int)(vtarg_set_val * 100 + .5)) / 100;
           if (sscanf_success < 1 || vtarg_set_val < 0) {
-            pmsg_error("invalid vtarg value '%s'\n", extended_param);
+            pmsg_error("invalid vtarg value %s\n", extended_param);
```
stefanrueger, line 1588: This change is good! thanks

src/jtag3.c
MCUdude: I've pushed a commit that solves most (all?) of the issues you pointed out. Please review.

src/jtag3.c
```diff
@@ -1520,11 +1520,25 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
 
-    else if ((str_eq(extended_param, "hvupdi")) && (lsize(pgm->hvupdi_support) > 1)) {
+    // HVUPDI
+    // All programmers that supports UPDI programming should have hvupdi_support=1 in avrdude.conf
+    // Type 0: 12V pulse on UPDI pin
+    // Type 1: No HV UPDI
+    // Type 2: 12V pulse on RESET pin
+    if(str_starts(extended_param, "hvupdi")) {
+      if(lsize(pgm->hvupdi_support) < 1) {
+        pmsg_error("programmer does not support high voltage UPDI programming\n", extended_param);
```
stefanrueger, line 1533: `extended_param` argument not needed (no `%s` in format)

src/jtag3.c
```diff
@@ -1535,12 +1549,12 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
     // Bit 2 EOF: Agressive power-down, sleep after 5 seconds if no USB enumeration when set to 0
     // Bit 1 LOWP: forces running at 1 MHz when bit set to 0
     // Bit 0 FUSE: Fuses are safe-masked when bit sent to 1 Fuses are unprotected when set to 0
-    else if (str_starts(extended_param, "suffer") ) {
+    if(str_starts(extended_param, "suffer")) {
```
stefanrueger, line 1552: `suffer`, `vtarg_switch` and `vtarg` extended options seem to be ignored when the programmer does not have the corresponding feature:

For example,

``` C
   if(str_starts(extended_param, "suffer")) {
     if(pgm->extra_features & HAS_SUFFER) {
       // Process option
     }
   }
```

If you don't want that, there are two possible ways out:
``` C
   if(str_starts(extended_param, "suffer") && pgm->extra_features & HAS_SUFFER) {
     // Process option
   }
```
In this case we get the default error message `pmsg_error("invalid extended parameter %s\n", extended_param);`

Or, alternatively, with a bespoke error message
``` C
   if(str_starts(extended_param, "suffer")) {
     if(pgm->extra_features & HAS_SUFFER) {
       // Process option (ending in continue/break)
     }
     pmsg_error("extended parameter %s used despite programmer not having this feature\n", extended_param);`
     rv = -1;
     break;
   }
```

src/jtag3.c
stefanrueger: Much better! Thanks for carefully compartmentalising the parsing options. Two small remaining comments (see code comments)

src/jtag3.c
stefanrueger: OK, I didn't think this through well enough. The code is OK and gives an error message. My first suggested "solution" is equivalent to existing code.

PR #1615, opened 2024-01-11
Update the GPL version 2 text

ndim:
Update the GPL version 2 text to the current text as published at

    https://www.gnu.org/licenses/old-licenses/gpl-2.0.txt

The changes are
  * the updated mailing address of the FSF
  * reference to LGPL with name change ("Library" -> "Lesser")
  * whitespace formatting
  * "write to" new FSF address to get a copy of the license instead of "seeing http://www.gnu.org/licenses/"


mcuee:
Looks like to a good change for avrdude 7.3 release.

PR #1618, opened 2024-01-16
Remove artificial 1Mhz limit on jtag3/atmelice/etc ISP bitclock

matthijskooijman:
Previously, the code would limit the clock to 1Mhz. If a higher clock was specified, it would be set to 1.024Mhz (0x400 khz).

This commit removes this artificial limit, while still preventing overflowing the two bytes available in the command (effectively limiting to 65535 kHz).

The original limit seems inappropriate, because the JTAGEICE3 and ATMEL ICE actually support higher clocks than the old limit. For SPI, the JTAGICE3 documents supporting up to 1.875Mhz and the ATMEL ICE up to 5Mhz. In practice the JTAGEICE3 (with the newer EDBG version where it becomes pretty much identical to the ATMEL ICE) also works up to 5Mhz.

When trying to set higher values, the behaviour seems to depend on the current value. The programmer seems to either keep the previous value (e.g. from 3Mhz to 10Mhz), or use 8kHz / 125μs (e.g. from 5Mhz to 10Mhz). In any case, the value read back afterwards (and printed with -v) reflects the actual value.

Also note that the code also applies a lower limit of 1kHz (to ensure the raw value sent to the device, in kHz, is never zero), but in practice sending a value of lower than 8kHz ends up setting 8kHz (125μs). This is also the minimum value in the device documentation.

This commit was tested with:

         Programmer Type       : JTAG3_ISP
         Description           : Atmel AVR JTAGICE3 in ISP mode
         ICE HW version        : 2
         ICE FW version        : 3.55 (rel. 130)
                 
~~I've marked this PR as a draft, since I want to test it with an Atmel ICE as well, but I won't be around one until later this week.~~ W00ps, clicked the wrong button. Oh well.

It might also be good to test this on other programmers that use the same code. From avrdude.conf.in, that would be:
 - Atmel AVR XplainedMini in ISP mode
 - Atmel PowerDebugger (ARM/AVR) in ISP mode
 - MPLAB(R) PICkit 4 in ISP mode
 - MPLAB(R) SNAP in ISP mode

I do not have access to any of these, unfortunately.

### DIFF
--- a/src/stk500v2.c
+++ b/src/stk500v2.c
@@ -3567,8 +3567,8 @@ static int stk500v2_jtag3_set_sck_period(const PROGRAMMER *pgm, double v) {
   unsigned char value[3];
   unsigned int sck;
 
-  if (v < 1E-6)
-    sck = 0x400;
+  if (v < 1 / (1000.0 * 0xffff))
+    sck = 0xffff;
   else if (v > 1E-3)
     sck = 1;
   else


MCUdude:
Neat! I never realized that the ISP clock frequency was limited. I'll hoop up my oscilloscope and test my Powerdebugger, MPLAB SNAP and PICkit4. I'll keep you updated!

MCUdude:
Maybe @xedbg know what the actual upper limit for the 16-bit bitclock value is?

MCUdude:
Strange. I can't get my PICkit4 and SNAP to clock higher than ~360kHz.
However, my Powerdebugger that previously was limited to 1 MHz (verified on my oscilloscope) can now clock at a whopping 5.26 MHz! Flashing a 128kiB part takes just a few seconds!

```
$ ./avrdude -cpowerdebugger_isp -pat90can128 -Uflash:w:/Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex:i -B6MHz
sck is 6000khz
avrdude: device signature = 0x1e9781 (probably c128)
avrdude: Note: flash memory has been specified, an erase cycle will be performed.
avrdude: erasing chip

avrdude: processing -U flash:w:/Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex:i
avrdude: reading input file /Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex for flash
         with 131072 bytes in 1 section within [0, 0x1ffff]
         using 512 pages and 0 pad bytes
avrdude: writing 131072 bytes flash ...
Writing | ################################################## | 100% 3.05 s 
avrdude: 131072 bytes of flash written
avrdude: verifying flash memory against /Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex
Reading | ################################################## | 100% 2.25 s 
avrdude: 131072 bytes of flash verified

avrdude done.  Thank you.


$ ./avrdude -cpickit4_isp -pat90can128 -Uflash:w:/Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex:i -B6MHz
sck is 6000khz
avrdude: device signature = 0x1e9781 (probably c128)
avrdude: Note: flash memory has been specified, an erase cycle will be performed.
avrdude: erasing chip

avrdude: processing -U flash:w:/Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex:i
avrdude: reading input file /Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex for flash
         with 131072 bytes in 1 section within [0, 0x1ffff]
         using 512 pages and 0 pad bytes
avrdude: writing 131072 bytes flash ...
Writing | ################################################## | 100% 46.71 s 
avrdude: 131072 bytes of flash written
avrdude: verifying flash memory against /Users/hans/Documents/avrdude-test-script/test_files/the_quick_brown_fox_131072B.hex
Reading | ################################################## | 100% 46.55 s 
avrdude: 131072 bytes of flash verified

avrdude done.  Thank you.
```

@xedbg or @janegilruud why is the PICkit4/SNAP ISP clock so slow compared to the Power Debugger? Is it Avrdude that doesn't send the right commands? I _do_ get a 125kHz clock if I use -B8 with the PICkit4, so something must be working.

MCUdude:
The Xplained Mini ISP is a little odd. Its ISP clock speed is ~55.2kHz regardless of the clock setting. The Xplained Mini ISPis very limited, so I wouldn't expect wonders here. A fixed ISP clock speed is probably fine for a tiny board with a fixed target chip with not-that-much flash memory.

xedbg:
> Maybe @xedbg know what the actual upper limit for the 16-bit bitclock value is?

The upper limit is the maximum that the tool is allowed (by the user) to clock at.  The actual capability of the tool depends on its hardware.  Clocking slower does not cause issues, but clocking too fast often does.

Most tools try to use a SPI module, but the step granularity of the clock depends on the module and how the peripheral clock is wired.  At lower clocks they usually have to revert to bit-banging, or drop the main clock to a lower frequency to allow the peripheral derived clock to be low, which has its challenges too.  So dynamic range is tricky.

The UC3/EDBG is generally better at bit-banging than a SAME70 (IMHO) as it has local IO bus access.  It also has a JTAG accelerator (but that is irrelevant here), and more time was spent optimising SPI timing since it was designed in the "SPI-era".  So its quite likely that it out-performs PICkit4 when using those older physical interfaces.

Xplained Mini is what it is - don't expect more! :)

MCUdude (approved):
Looks good to me! None of the programmers I've tested this on had any issues with the bitclock number being higher than what it was able to physically output.

matthijskooijman:
> Neat! I never realized that the ISP clock frequency was limited. 

I had noticed it somewhere last year and thought it was weird, but didn't actually investigate until now, since I expected it to be some complex interaction between software and firmware, but turns out it was just a arbitrary limit (which has been present since the code was first introduced, btw).

I've tested on Atmel ICE today as well, which works just like the JTAGICE3 (up to 5Mhz) as expected.

This would be the point where I would remove the "draft" from the PR, but I just noticed I actually forgot to mark it as draft (even though I said I would). Anyway, it looks like all tests have been positive (thanks to everyone for testing!), so this can be merged as far as I am concerned :-)

stefanrueger:
Happy to merge for v7.3

PR #1634, opened 2024-01-23
Fix upload of one-byte files to flash

### LINKED ISSUE #1633: Writing one-byte files to flash fails
```
$ avrdude -qqc dryrun -p m328p -U flash:w:42:m && echo OK
avrdude warning: verification mismatch
        device 0xff != input 0x2a at addr 0x0000 (error)
avrdude error: verification mismatch
```

A rather rare case, but still, if someone wants to just change/overwrite the first byte of flash it won't work. This problem is programmer and part independent. What should happen is that (at least) 2 bytes are read from flash, the first byte modified and (at least) 2 bytes written back to the device.

stefanrueger:
Fixes #1633 

With this PR all works fine:
```
$ avrdude -qqc dryrun -p m328p -U flash:w:42:m && echo OK
OK
```

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -308,7 +308,7 @@ int avr_mem_hiaddr(const AVRMEM * mem)
 
   /* return the highest non-0xff address regardless of how much
      memory was read */
-  for (i=mem->size-1; i>0; i--) {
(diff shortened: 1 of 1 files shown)


mcuee:
This PR shoudl be good to be merged.

PR #1663, opened 2024-02-06
Automatically determine Autoconf package version from cmake package version

ndim:
This PR will determine the `PACKAGE_VERSION` which the Automake/Autoconf avrdude package uses from the package version which the top level `CMakeLists.txt` file uses.

ETA 2024-02-06 23:59:59 UTC.

PR #1668, opened 2024-02-07
The test(1) string comparison uses =, not ==

ndim:
The test(1) command string comparison operator is a single equals sign, not a double equals sign.

Pointed out by @mcuee at https://github.com/avrdudes/avrdude/pull/1663#issuecomment-1931024192


mcuee:
Looks good to me.

PR #1680, opened 2024-02-07
Fix -Wcalloc-transposed-args warning

ndim:
The calloc(3) function has two arguments: the number of members, and the size of one member.

sizeof(...) makes more sense as member size, and 1 as number of members. At least that is what the gcc warning wants.

This was the only compiler warning I could get on my Fedora 40 test build (gcc 14) with CFLAGS="-Wall -Wextra -Werror".

### DIFF
--- a/src/avrftdi.c
+++ b/src/avrftdi.c
@@ -1271,7 +1271,7 @@ avrftdi_setup(PROGRAMMER * pgm)
 	avrftdi_t* pdata;
 
 	
-	if(!(pgm->cookie = calloc(sizeof(avrftdi_t), 1))) {
+	if(!(pgm->cookie = calloc(1, sizeof(avrftdi_t)))) {
 		log_err("Error allocating memory.\n");
 		exit(1);
 	}


ndim:
At this time, I am content that avrdude compiles without warnings on my systems.

stefanrueger:
@ndim Good catch. I plan to review avrdude's use of [cm]alloc/strdup etc and replace the offending if block with sth like
```
    pgm->cookie = cfg_malloc(__func__, sizeof(avrftdi_t));
```
sometime in the future for consistency of error messages etc. Actually, will probably `'#define mmt_alloc(n) cfg_malloc(__func__, n)` in an auspicious header and then use the magic-memory-tree allocation shortcut. But, yes, will merge your PR at the next mergefest to remove the warning in the meantime.

> macOS build warning has been there for long time and probably difficult to fix.

@mcuee That's an upstream problem with macOS's flex, which generates the code. Better to complain with them :wink:

> warnings generated by MSYS2 clang64

@mcuee best done by someone who works with clang64 and who can verify that other (windows) compiler won't introduce a warning as consequence. I suggest to live with this for the time being.

mcuee:
@ndim and @stefanrueger 

No issues. I am just trying my luck. :-)

PR #1687, opened 2024-02-14
Remove 32bit MSYS2 mingw32 and clang32 build

mcuee:
MSYS2 has dropped 32bit binary packages for libusb, libusb-compat-git and libftdi.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -257,10 +257,8 @@ jobs:
     strategy:
       matrix:
         include:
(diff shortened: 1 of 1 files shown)


mcuee:
Tested to be okay here before I created this PR:
https://github.com/mcuee/avrdude/actions/runs/7896972282

mcuee:
This PR also passed the check.
https://github.com/avrdudes/avrdude/pull/1687/checks

@stefanrueger 
This PR should be safe to fix so that CI will not fail again for new PRs.

Issue #1683, opened 2024-02-11
avrintel.h missing from installation


drmpjz:
Build environment: Linux openSUSE 15.5

Sorry, not a cmake expert, so I don't have a patch :-(

Problem:
libavrdude.h ist installed in the include directory of  CMAKE_INSTALL_PREFIX:PATH (e.g. /usr/include or /usr/local/include)  

In libavrdude.h you have since 7.3

#include "avrintel.h"

but this is not installed to the same path, so any compilations of programs using libavrdude.h will fail.

avrdude itself just works, because it is build in the full source tree.

I've found that after the build in build_linux there is a file named install_manifest.txt, which contains a list of the files and path of all the files that final install step will distribute on the system. I expect that avrintel.h just needs to be added there.  

Unfortunately I did not find an (to me) obvious place in the build systems to do so.

mcuee:
It seems to me we need to change the following file.
https://github.com/avrdudes/avrdude/blob/main/src/CMakeLists.txt

Unfortunately I also do not know much about CMake and I do not know how to fix it.

drmpjz:
Maybe here?
set_target_properties(libavrdude PROPERTIES
    PREFIX ""
    PUBLIC_HEADER "libavrdude.h"
    VERSION 1.0.0
    SOVERSION 1
    )

PUBLIC_HEADER seems to be the keyword. Now this only needs to be a list.

P.S. I have tested that if I copy avrintel.h manually to /usr/include both compilation of the program works and the resulting binary also works as expected.

drmpjz:
PUBLIC_HEADER "libavrdude.h;avrintel.h"

seems to do the trick (still no cmake expert, only stackoverflow knowledge...)

mcuee:
@Youw 

Sorry to disturb you. Just wondering if the work-around by @drmpjz is correct or not. Thanks.

ndim:
As we only need a pointer to that type in `libavrdude.h`, we could use an incomplete type in `libavrdude.h`, without needing to include `avrintel.h` from `libavrdude.h`.

That would be my proposed resolution to this issue.

drmpjz:
Thanks a lot for your serious consideration of this issue!
As for the "are there users of libavrdude out there part", I know of at least one and that is actually the reason why I am building avrdude :-)
[...]

PR #1688, opened 2024-02-15
libavrdude.h fixes and installation of libavrdude-avrintel.h

ndim:
This cleans up `libavrdude.h` a bit and installs `libavrdude-avrintel.h`:

  * `libavrdude.h` will create an error if `LIBAVRDUDE_INCLUDE_INTERNAL_HEADERS` is defined
  * add comment on top of `libavrdude-avrintel.h` about not directly `#include`ing it
  * install `libavrdude-avrintel.h` alongside `libavrdude.h` (both cmake and autotools)
  * use proper C99 and later definitions for a C function without parameters: `rettype funcname(void);`
  * have the `libavrdude.h`/`libavrdude-avrintel.h` combination `#include` all required libc headers declaring the types used in them (add `stddef.h` for `size_t`)

Test programs for testing language compatibility will come later when I figure out how to build them in cmake. A program like the following should build with all compilers and all supported language standards:

```C
#include <libavrdude.h>

int main(void) {
  return 0;
}
```


mcuee:
From github action, it seems to be good under Linux (CMake and auto-tools).

I have also tested this under macOS.

I think this PR is good to go and it finally fixes #1683.
* #1683

stefanrueger:
I have reviewed the changes in this PR, and they look good to me. Happy to merge at next mergefest (we tend to let PR lie around for a few days and normally wait to merge them until a few ones are ready)

PR #1689, opened 2024-02-16
Updated section about NVM models support

dbuchwald:
Two small updates to SerialUPDI documentation. Removed section about fuses (no longer required) and updated list of supported NVM versions.


mcuee:
Great. Looks good to me.

stefanrueger:
Thanks for updating the documentation @dbuchwald! Will merge at the next mergefest

PR #1681, opened 2024-02-07
Improve autotool build

ndim:
This Draft PR addresses the Autotools build issues named in https://github.com/avrdudes/avrdude/issues/1664 and other things.

I will keep this a Draft PR until development and surrounding discussions have finished and it can merged.

PR #1698, opened 2024-02-21
Add build information to libavrdude and avrdude and display it

ndim:
The goal of this Draft PR is to develop a way to add useful build information to libavrdude and avrdude, give libavrdude users (like avrdude) access to that buildinfo, and have avrdude show the buildinfo.

The ideas for this were originally discussed in https://github.com/avrdudes/avrdude/pull/1681#issuecomment-1952668125 and the following comments.

This is a basic proof of concept for showing a long verbose message with build information. The verbosity, the exact information, its representation, the API and ABI are subject to discussion.

At this time, the code is based on PR #1681.

This is quick and dirty and not tested comprehensively:
    
  * Switch avrdude's `main.c` from getopt() to getopt_long() (available on GNU, BSD, and
    the existing msvc/getopt.[ch] already implement getopt_long() on
    Windows)
    
  * Add `--version` argument which shows a long version message.
  
    As `-V` is already in use, the version message is available only with the long form `--version`.

  * Define build information for libavrdude and the avrdude tool
      * in `buildinfo.c` and `libavrdude.h` for libavrdude
      * in `main.c` for avrdude

The result looks about like

```
avrdude (...) 7.3-20240221 (28e6f254)
Copyright (C) ...
License GPL...
This is free software...
avrdude 7.3-20240221 (28e6f254)
  1. buildsystem: autotools
libavrdude 7.3-20240221 (28e6f254)
  1. buildsystem: autotools
  2. libelf
  3. libusb
  4. libusb_1_0
  5. libhidapi
  6. libftdi1
  7. libreadline
  8. libserialport
  9. parport
 10. linuxgpio
 11. linuxspi
```


src/libavrdude.h
```diff
@@ -1522,8 +1526,8 @@ void terminal_setup_update_progress();
 extern "C" {
 #endif
 
-void win_sys_config_set(char sys_config[PATH_MAX]);
-void win_usr_config_set(char usr_config[PATH_MAX]);
+void win_sys_config_set(char *sys_config);
```
stefanrueger, line 1533: Well spotted!

src/main.c
```diff
@@ -218,6 +220,39 @@ const char *pgmid;              // Programmer -c string
...
+  printf("License GPL...\n");
+  printf("This is free software...\n");
+
+  printf("avrdude %s\n", avrdude_buildinfo[0]);
+  print_buildinfos(avrdude_buildinfo);
+
+  const char *const *libavrdude_buildinfo = avr_get_buildinfo();
+  printf("libavrdude %s\n", libavrdude_buildinfo[0]);
+  print_buildinfos(libavrdude_buildinfo);
+}
```
MCUdude, line 253: Is there a good reason to use printf instead of msg_info here?

ndim: It is a quick hack proof of concept, and I have not familiarized myself with the message API yet, so... No :-)

MCUdude: No worries, take your time. For reference, here are all of them:
https://github.com/avrdudes/avrdude/blob/5c1649dceed6410e967e6fda57ac5e675fa5017c/src/avrdude.h#L64-L96

* There are `msg_*` macros that just print whatever you insert.
* There are `pmsg_*` that prints `avrdude: ` and then whatever you insert
* There are `imsg_*` that indents so that text printed on the screen will align with `p_msg`

And suffixes:
* There are `_info` which prints what you insert regardless of mode
* There are `notice` that will print with one or more `-v`'s
* There are `notice2` that will print with two or more `-v`'s
* And more 🙂

ndim:
As build version information is system independent, build version (git branches, etc.) information it relatively simple to add.

Adding information about the optional libraries and features built into this version is a bit more complex. To do that properly, we need to build a bit of build infrastructure, and then make use of that:

- [ ] cmake buildsystem function generate `.c` files containing build information for avrdude and libavrdude
- [ ] automake buildsystem macro generate `.c` files containing build information for avrdude and libavrdude, possibly emulating the `VERSIONINFO_SETUP` to fetch that information from the cmake config files
- [ ] build object file with build information and link that into the avrdude executable and libavrdude library
- [ ] Add `avrdude --version` or something similar which shows all that verbose build information (note that `avrdude` does not support long form arguments until now)

Aiming for a quicker, "good enough" class of solution first, then the proper one.

stefanrueger:
BTW, I think the merge conflict comes about b/c the `getopt()` options were slightly changed: `Y` was deprecated and has no longer an optional argument, so the last `:` was removed....

ndim:
The biggest problem here is that my Linux `getopt(3)` man page says `getopt()` is part of `POSIX.1-2008`, while `getopt_long()` is `GNU`. And GNU functions do not necessarily work on BSD, OSX, and Windows.

ndim:
Oh, and I want a different C API for the version information. The "list of strings" as used by libgphoto2 has drawbacks. I would prefer a "list of key-value-pairs" type of interface, with keys like "gpiod" and values like "<1.4" or ">=1.4<1.8" or ">= 1.8".

Such an API should be well thought out, as it should last at least 10 years. Not sure this should be rushed within the next 10 days.

ndim:
> BTW, I think the merge conflict comes about b/c the `getopt()` options were slightly changed: `Y` was deprecated and has no longer an optional argument, so the last `:` was removed....

And `w` was removed as well.

ndim:
That this PR now building on the https://github.com/avrdudes/avrdude/pull/1936 PR which now introduces the long options like `--help` in addition to the classic short `-?` options.

stefanrueger:
@ndim

Thanks for engaging with the AVRDUDE project! Is your draft PR still something you'd like to progress for the  avrdude 8.2 release?

mcuee:
I guess this PR will not be ready for 8.2 release now that @ndim seems to be busy.

stefanrueger:
Closing this PR as it's been an unchanged draft for more than 1.5 years.

PR #1700, opened 2024-02-22
Fix reporting of too short r/w times for files with holes

stefanrueger:
Given the recent obsession with speed benchmarks I reviewed the code that reports timing and found that files with holes in them will have times reported that are too low. Uploading `tools/test_files/rjmp_loops_for_bootloaders_32768B.hex` will be reported correctly with this PR (this file has large "holes").


mcuee:
Looks like a good change. WIll test this later to see if the differences.

Issue #1706, opened 2024-02-25
error: use of undeclared identifier 'AVRDUDE_FULL_VERSION'


dl8dtl:
I'm not sure, but I think this is somehow related to this PR …
Everything works fine when I rund `build.sh` on a clone of the official AVRDUDE repo.
However, when running it on my private fork (where the official repo is only `upstream` remote), even with the same source tree as the official repo, I eventually get:
```
[ 94%] Built target libavrdude
[ 96%] Building C object src/CMakeFiles/avrdude.dir/main.c.o
/home/joerg/src/avrdude/src/main.c:55:23: error: use of undeclared identifier 'AVRDUDE_FULL_VERSION'
char * version      = AVRDUDE_FULL_VERSION;
```

_Originally posted by @dl8dtl in https://github.com/avrdudes/avrdude/issues/1681#issuecomment-1962747588_

dl8dtl:
Just turning it into an issue for easier tracking.
My current fork has stashed changes I'd like to preserve (but not commit yet), so cloning the repo anew isn't a good solution for me.
Also, I think we'd better know the reason for why this happens. Alas, I couldn't understand within 5 minutes last night what's going on here. ;-)

dl8dtl:
For reference, here's the verbose build output for
`./build.sh -f -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON`
[logfile.txt](https://github.com/avrdudes/avrdude/files/14395615/logfile.txt)

mcuee:
I can reproduce the issue under Windows MSYS2 mingw64 build environment.
* https://github.com/avrdudes/avrdude/pull/1681#issuecomment-1962826787

@dl8dtl 

Interestingly I just reproduced your issue under MSYS2 mingw64 with clone from official avrdude, and the solution is the same -- redo the clone. I am not so sure what is the issue though. Occassionally there is the issue of getting wrong version number. And the solution has always the same for me -- to delete the local repo and clone again.

Note: I have delete the build directory before the build but it does not help.

<details>

```
MINGW64 /c/work/avr/avrdude_test/avrdude_main
$ rm -rf build_mingw64_nt-10.0-22631/

$ ./build.sh
-- Building for: Ninja
-- The C compiler identification is GNU 13.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/msys64/mingw64/bin/cc.exe - skipped
-- Detecting C compile features
[...]

PR #1707, opened 2024-02-25
Fix cmake build errors related to ac_cfg.h

ndim:
This fixes a few errors related to how cmake builds deal with `ac_cfg.h`. This has been discussed in https://github.com/avrdudes/avrdude/pull/1681#issuecomment-1962747588 and following comments, and then in issue https://github.com/avrdudes/avrdude/issues/1706.

  * While testing, I ran `cmake .` a few times which generates a cmake buildsystem inside the source tree. Some of those files still had to be added to `.gitignore`.

  * Using the identifier `version` inside `src/main.c` to refer to `AVRDUDE_FULL_VERSION` is about as confusing as using `VERSION` instead of `AVRDUDE_FULL_VERSION` was. So I removed `char *version` and just let the functions reference `AVRDUDE_FULL_VERSION` directly. I trust compilers these days to have all those references refer to the same bit of constant memory, and not allocate several copies of `AVRDUDE_FULL_VERSION`.
 
  * Fix cmake not finding the generated `ac_cfg.h` when the (often unused and unnecessary) `src/ac_cfg.h` does *NOT* exist. See commit message.

  * And most importantly, fix the C preprocessor mistakenly looking in `src/` before `build_$ostype/src/` when looking for `ac_cfg.h`: Change from `#include "ac_cfg.h"` to `#include <ac_cfg.h>` in all source files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,9 @@ cscope.out
 out/
 build/
 build_*/
+CMakeFiles/
+/CMakeCache.txt
+cmake_install.cmake
 
 # Visual Studio
 .vs/
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -113,10 +113,19 @@ endif()
 # Configure files
 # =====================================
 
-configure_file(cmake_config.h.in ac_cfg.h)
-configure_file(avrdude.spec.in avrdude.spec)
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/cmake_config.h.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/ac_cfg.h"
+)
+configure_file(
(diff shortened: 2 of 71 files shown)


ndim:
Note that this does **not** add checks to `src/CMakeLists.txt` looking for a `src/ac_cfg.h` when building out of tree, and then aborting the build if one is found. Letting the C preprocessor look in the out of tree build directory means such a check is not needed any more.

ndim:
The issue of the rogue `src/ac_cfg.h` would probably have been an issue in autotools builds as well, but those have other issues with out-of-tree builds aborting at this time, so I have not run into those yet.

mcuee:
Looks like a good change to me.

Tested under Windows with different build (autotools build under MSYS2 mingw64, CMAKE mingw64 build and CMAKE VS2019 build, no build issues found, the version info of the binaries generated is also correct.

Issue #1726, opened 2024-03-28
Wrong message level in `jtag3_prmsg()`


dl8dtl:
`jtag3_prmsg()` is used to produce debug and trace messages, it is only called when verbose level is DEBUG or higher:

```
  } else if (verbose >= 3) {
    msg_info("\n");
    jtag3_prmsg(pgm, *resp, status);
```
Yet, it produces messages at MSG_INFO level (corresponding to default logging, no -v needed):
```
static void jtag3_prmsg(const PROGRAMMER *pgm, unsigned char *data, size_t len) {
  if (verbose >= 4) {
…
    msg_trace("Raw message:\n");
…
      msg_trace("%02x ", data[i]);
(OK so far)
…
  }
…
      msg_info("[info] ");
      msg_info("[general] ");
      msg_info("[AVRISP] ");
      msg_info("[AVR] ");
      msg_info("[scope 0x%02x] ", data[0]);
      msg_info("OK\n");
      msg_info("FAILED");
        msg_info(", reason: %s\n", reason);
        msg_info(", unspecified reason\n");
      msg_info("Data returned:\n");
      msg_info("Info returned:\n");
          msg_info("%c", data[i]);
          msg_info("\\%03o", data[i]);
      msg_info("\n");
        msg_info("PC reply too short\n");
        msg_info("PC 0x%0lx\n", pc);
      msg_info("unknown message 0x%02x\n", data[1]);
```
Seems to me these ought to become `msg_debug()` instead.

dl8dtl:
ps: Same for `jtag3_prevent()`.

PR #1727, opened 2024-03-28
Replace msg_info() by msg_debug() in jtag3_prmsg() and jtag3_prevent()

dl8dtl:
Both these functions are called at verbosity level DEBUG and above, so they must not produce INFO level messages.

Supposed to fix issue #1726


stefanrueger:
Looks good to me, can be merged.

PR #1728, opened 2024-03-29
Provide lmsg_xyz() functions to start message at left margin

stefanrueger:
This PR adds a `MSG2_LEFT_MARGIN` flag as alternative to the `\v` messaging convention.

Some messaging needs a functionality to end an open line, so that the next message starts at the left margin, ie, the ability to print a `\n` iff the last printed character was not already a `\n`. This used to be effected by `\v` as first char in the message format.

This PR introduces a dedicated `lmsg_`*xyz*`()` set of messaging functions that utilises a `MSG2_LEFT_MARGIN` flag to enforce a left margin start for the message. The PR keeps vertical tab `\v` at  the start of the format string for the time being for backward compatibility, but neither AVRDUDE itself nor `libavrdude` uses the `\v` convention any more.

The PR is motivated by the observation in the current PR #1714 that, for other applications, `\v` processing can be more difficult than processing a flag such as `MSG2_LEFT_MARGIN`.

In order to test against regression, I suggest `dryrun` runs with `-vvv` and compare outputs, eg,
```
$ avrdude_main -vvv -cdryrun -pm328p -U some-sketch.hex >& /tmp/a
$ avrdude_pr   -vvv -cdryrun -pm328p -U some-sketch.hex |& diff - /tmp/a
```

### DIFF
--- a/src/avrdude.h
+++ b/src/avrdude.h
@@ -60,6 +60,7 @@ int avrdude_message2(FILE *fp, int lno, const char *file, const char *func, int 
 #define MSG2_INDENT1      16 // Start by printing indentation of progname+1 blanks
 #define MSG2_INDENT2      32 // Start by printing indentation of progname+2 blanks
 #define MSG2_FLUSH        64 // Flush before and after printing
+#define MSG2_LEFT_MARGIN 128 // Print \n unless last character printed was \n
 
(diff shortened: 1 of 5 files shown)


dl8dtl (approved):
Looks good to me!

PR #1729, opened 2024-03-29
Change return type for avr_(ms|us)timestamp() to uint64_t

### LINKED ISSUE #1722: avr*timestamp() oddities
```
// Return us since program start, rolls over after ca 1h 12min
unsigned long avr_ustimestamp() {
```
…
```
// Return ms since program start, rolls over after ca 49d 17h
unsigned long avr_mstimestamp() {
  return avr_ustimestamp()/1000;
```
How could `avr_ustimestamp` possibly roll over after only 49 days when `avr_ustimestamp`, which is used as the base for its computation, already rolls over after 1 hour? I see that `avr_ustimestamp` maintains an internal start time as `unsigned long long`, but as the result is shortened to `unsigned long` (and `avr_mstimestamp` cannot access its internal state), the extra long intermediate state doesn't help much.
Besides, I think the figures listed there depend on the CPU architecture:
```
% cat size.c
#include <stdio.h>

int main(void) {
  printf("sizeof(unsigned long): %zu\n", sizeof(unsigned long));
  printf("sizeof(unsigned long long): %zu\n", sizeof(unsigned long long));
  return 0;
}
% cc -m32 -O -o size size.c
j@uriah 125% ./size
sizeof(unsigned long): 4
sizeof(unsigned long long): 8
% cc -m64 -O -o size size.c
% ./size
sizeof(unsigned long): 8
sizeof(unsigned long long): 8
```
I'd suggest to move the "epoch" state to outside of these functions (maybe even with the option to trigger an external reset of them), and then have the `avr_ustimestamp` and `avr_mstimestamp` routines access that state individually.

stefanrueger:
Fixes #1722

@ndim Does commit https://github.com/avrdudes/avrdude/commit/2283b970ecda3da4f2824a56fcb26905e3dd32f7 increment the libavrdude version correctly?

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -572,15 +572,15 @@ error:
 }
 
 
-// Return us since program start, rolls over after ca 1h 12min
-unsigned long avr_ustimestamp() {
+// Return us since first call
+uint64_t avr_ustimestamp() {
   struct timeval tv;
 
   memset(&tv, 0, sizeof tv);
   if(gettimeofday(&tv, NULL) == 0) {
-    static unsigned long long epoch;
+    static uint64_t epoch;
     static int init;
-    unsigned long long now;
+    uint64_t now;
 
     now = tv.tv_sec*1000000ULL + tv.tv_usec;
     if(!init) {
@@ -593,8 +593,8 @@ unsigned long avr_ustimestamp() {
   return 0;
 }
 
-// Return ms since program start, rolls over after ca 49d 17h
-unsigned long avr_mstimestamp() {
+// Return ms since first call to avr_ustimestamp() above
+uint64_t avr_mstimestamp() {
   return avr_ustimestamp()/1000;
 }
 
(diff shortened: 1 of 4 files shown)


ndim:
> @ndim Does commit [2283b97](https://github.com/avrdudes/avrdude/commit/2283b970ecda3da4f2824a56fcb26905e3dd32f7) increment the libavrdude version correctly?

@stefanrueger Yes, but....

Yes. Changing the return types is an incompatible change, so changing the soname is the right thing to do, and commit [2283b97](https://github.com/avrdudes/avrdude/commit/2283b970ecda3da4f2824a56fcb26905e3dd32f7) does that correctly.

But... if you focus on compatibility with avrdude releases (as opposed to compatibility with each and every commit which happens between avrdude releases), you do not need to do anything. We already have such a change to the soname after the last release in commit a83f3cf08.

It could be debated whether you prefer to change the soname in the middle or at the end of a release  cycle, but you need some process to ensure it does happen before a new release. Right now, avrdude has changed the soname in the middle of the release cycle, and therefore soname changes are not strictly necessary.

So... commit 2283b97 certainly does not break anything, and it avoid unrecognized ABI breakage. It increases the soname number by one more than strictly necessary from release to release, but the supply of integer numbers is infinite. I would think one soname change per release cycle is enough, and people who build some non-release commit can be entrusted with figuring out and dealing with (i.e. recompiling) such changes on their own.

stefanrueger:
@ndim OK, thanks for the detailed explanation. I agree it is sufficient to increase the soname for every release. So, will revert the increment.

PR #1741, opened 2024-04-08
Fix message levels in jtag3_print_data()

dl8dtl:
These are debugging functions, and not supposed to print at MSG_INFO level.


mcuee:
This should be safe to be merged.

PR #1754, opened 2024-04-17
Remove gcc warning when printf format is ""

stefanrueger:
This warning only occurs when checking the msg_...() functions for having the correct parameters. The way I do that is by replacing the avrdude.h file with one that #defines the msg_... functions to printf for which the compiler knows which paramaters of which type are expected given that the format is a string constant. That check, however, produces false alarms which are not helpful.

This commit switches these unhelpful warnings from gcc off.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -80,7 +80,7 @@ if(MSVC)
     list(APPEND EXTRA_WINDOWS_INCLUDES "msvc")
 else()
     set(LIB_MATH m)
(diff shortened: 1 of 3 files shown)


stefanrueger:
@Floessie: Thanks for the tip; really helpful. And now that gcc checks the printf syntax for the messaging we *defo* need to suppress the warning about empty formats: they make sense for the AVRDUDE lmsg_...() messaging functions that issue a `\n` unless we are already at the left margin.

PR #1767, opened 2024-04-19
Make programmer stk500 (v1) libavrdude ready

stefanrueger:
- Make static variables const tables
 - Utilise magic memory tree interface

Should be safe to merge


MCUdude:
Looks good to me!

PR #1780, opened 2024-04-22
fix: worker errors get dispatched correctly

koen1711:


mcuee:
@koen1711 

Sorry but what are you trying to do here.

Issue #1782, opened 2024-04-22
cmake enabling LINUXGPIO on ubuntu 20.04 fails


stefanrueger:
After installing `libgpiod-dev` on ubuntu 20.04 (long-term-support until April 2030), cmake enabling LINUXGPIO fails to build: `gpiod_line_set_direction_input()` is missing in `libgpiod`. @mhei @mcuee Any ideas how to solve this?

```
$ sudo apt install libgpiod-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Suggested packages:
  libgpiod-doc
The following NEW packages will be installed
  libgpiod-dev
0 to upgrade, 1 to newly install, 0 to remove and 0 not to upgrade.
Need to get 49.2 kB of archives.
After this operation, 286 kB of additional disk space will be used.
Get:1 http://gb.archive.ubuntu.com/ubuntu focal/universe amd64 libgpiod-dev amd64 1.4.1-4 [49.2 kB]
Fetched 49.2 kB in 0s (1,202 kB/s)      
Selecting previously unselected package libgpiod-dev:amd64.
(Reading database ... 401048 files and directories currently installed.)
Preparing to unpack .../libgpiod-dev_1.4.1-4_amd64.deb ...
Unpacking libgpiod-dev:amd64 (1.4.1-4) ...
Setting up libgpiod-dev:amd64 (1.4.1-4) ...

$ sudo rm -rf build_linux
$  cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -D HAVE_LINUXGPIO=1 -D HAVE_LINUXSPI=1 -D HAVE_PARPORT=1 -B build_linux
-- The C compiler identification is GNU 9.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: /usr/bin/git (found version "2.25.1") 
-- Found FLEX: /usr/bin/flex (found version "2.6.4") 
-- Found BISON: /usr/bin/bison (found version "3.5.1")
-- Looking for libelf.h
-- Looking for libelf.h - found
-- Looking for libelf/libelf.h
-- Looking for libelf/libelf.h - not found
-- Looking for usb.h
-- Looking for usb.h - found
-- Looking for lusb0_usb.h
-- Looking for lusb0_usb.h - not found
[...]

PR #1788, opened 2024-04-23
Update build.yml to include libgpiod-dev for Linux builds

mcuee:
It is good to include  libgpiod in the Linux github action builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,6 +53,7 @@ jobs:
           libftdi1-dev
           libreadline-dev
           libserialport-dev
+          libgpiod-dev
           texinfo
           texlive
           texi2html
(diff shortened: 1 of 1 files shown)


mhei (approved):
LGTM

stefanrueger:
What about Issue #1782? Currently, my workaround is to *not* install libgpiod-dev. Wouldn't this PR make #1782 worse? I realise that my linux desktop doesn't have GPIOs so I should not `cmake -D HAVE_LINUXGPIO=1` *but* as maintainer I like to see whether the code at least compiles even though I have no device to test with.

mcuee:
> What about Issue #1782? Currently, my workaround is to _not_ install libgpiod-dev. Wouldn't this PR make #1782 worse? I realise that my linux desktop doesn't have GPIOs so I should not `cmake -D HAVE_LINUXGPIO=1` _but_ as maintainer I like to see whether the code at least compiles even though I have no device to test with.

github action is using Ubuntu 22.04 so it is not affected.

PR #1763, opened 2024-04-19
Make programmer usbasp libavrdude ready

stefanrueger:
- Move static variables to PDATA space in usbasp.c
 - Utilise magic memory tree interface for usbasp.c
 - Implement usbasp_tpi_write_byte()

It's a draft that is supposed to also fix #1755.

PR #1789, opened 2024-04-23
Change fuse bit test from wdton to ckout for reduced-core ATtiny parts

stefanrueger:
Setting WDT on in the fuse can render reduced-core ATtiny parts unprogrammable using TPI. Affected parts are t4, t5, t9, t10, t20, t40, t102 and t104. So, the tests better involve putting the clock signal on PB2 (fuse bit ckout).

Still needs to be confirmed and tested (<s>@mhei?</s> [edit: @brouwer?] @MCUdude?)

### DIFF
--- a/tools/test-avrdude
+++ b/tools/test-avrdude
@@ -348,8 +348,8 @@ for (( p=0; p<$arraylength; p++ )); do
         specify="fuse access: clear, set and read eesave fuse bit"
         command=(${avrdude[@]} -T '"config eesave=0; config eesave=1; config eesave"')
       else
-        specify="fuse access: clear, set and read wdton fuse bit"
-        command=(${avrdude[@]} -T '"config wdton=0; config wdton=1; config wdton"')
+        specify="fuse access: clear, set and read ckout fuse bit"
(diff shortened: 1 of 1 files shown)


MCUdude:
All good! Tested with #1763 and #1789 applied

```
$ ./test-avrdude -p "-c usbasp -p attiny9"
Testing avrdude version 7.3-20240424 (2c747dc0)
Prepare "-c usbasp -p attiny9" and press 'enter' or 'space' to continue. Press any other key to skip
✅   0.270 s: fuse access: clear, set and read wdton fuse bit
✅   0.452 s: chip erase
✅   1.145 s: flash -U write/verify holes_rjmp_loops_1024B.hex
✅   0.312 s: flash -T write/verify holes_rjmp_loops_1024B.hex
✅   0.517 s: chip erase and spot check flash is actually erased
```

stefanrueger:
> Setting WDT on in the fuse can render reduced-core ATtiny parts unprogrammable using TPI. Affected parts are t4, t5, t9, [...]

I now think this is a false alarm as the real culprit for the observed behaviour was a bug in the usbasp tpi fuse write code. As above test shows setting wdton active can *and does* work. As such this PR switching to ckout isn't really necessary.

PR #1790, opened 2024-04-25
Remove calls to exit() from libavrdude library functions

### LINKED ISSUE #774: Remove all calls to exit(3) from libavrdude library functions
(Previously, this had been marked as a "task" on Savannah, with nobody actually backing it.)
Quite some time ago, all the backend functionality of AVRDUDE had been moved out into libavrdude. The idea behind that was that the current `src/main.c` just describes *one* possible frontend, so it could be e.g. accompanied by a kind of GUI frontend as well.
However, some of the historic implementations gratuitously call `exit()` in case of a failure. This is fatal for a GUI application, which rather would prefer to give the user a kind of graphical response of the failure. Thus, call calls to `exit()` need to be converted into returning (and handling) an error code instead.

stefanrueger:
The PR addresses a longstanding issue #774. Rather than using `exit(0)` the latest git main already uses `return LIBAVRDUDE_EXIT`, but these return now need to be percolated to the top layer (`main.c` in the AVRDUDE application), which then should `exit(0)`. Other applications built with `libavrdude` might want to stop the session without further errors. There is still a smattering of `exit(1)` in the `libavrdude` code, but that is for really tricky situations where recovery is neigh impossible, for example when the process has run out of memory. To all intents and purposes this PR fixes #774.

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -471,7 +471,7 @@ int avr_read_mem(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *mem, con
   if (mem_is_signature(mem)) {
     if (pgm->read_sig_bytes) {
       int rc = pgm->read_sig_bytes(pgm, p, mem);
-      if (rc < 0)
+      if (rc < 0 && rc != LIBAVRDUDE_EXIT)
         led_set(pgm, LED_ERR);
       led_clr(pgm, LED_PGM);
       return rc;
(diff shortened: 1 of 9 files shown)


stefanrueger:
With this PR
```
$ avrdude -qq -P ch340 -c urclock -xshowversion  && echo "OK"
u7.7 w-u-jpr--
OK
```

Without this PR
```
$ avrdude -qq -P ch340 -c urclock -xshowversion  || echo "Not OK"
u7.7 w-u-jpr--
avrdude error: unable to read signature data for part ATmega328P, rc=-4
avrdude error: unable to read signature data, rc=-4
        use -F to override this check
Not OK
```

Above weird output is from a very recent git main, where `exit(0)` had been replaced with `return LIBAVRDUDE_EXIT` as an interim measure; this PR now finishes the work needed to remove `exit()` calls.

PR #1798, opened 2024-04-28
Add macOS autotools to github action

mcuee:


mcuee:
This PR should be safe to be merged. And it will help to check whether autotools build is okay under macOS or not, since @MCUdude still uses autotools build more often under macOS.

**Please use "Squash and merge" option**. Thanks.

mcuee:
Oops, I just noticed there is a typo. So I just fixed it in git main directly. No need to change NEWS file.

PR #1815, opened 2024-05-27
Guard serial_number access in usbhid_open() to avoid segfault

stefanrueger:
Unfortunately, I can no longer remember which particular instance lead to the segfault, when I played with a Pro Micro flushed as a programmer, but I could narrow it down to accessing a pointer that a library call returned, so it's reasonable to guard the access.

Should be safe to merge.


mcuee:
Looks good for me.

Issue #1711, opened 2024-02-28
Which programmer supports "-E reset"?


Tropaion:
Hello,
I don't know where to ask/find it.
I have an application, where, when the programming an ATxmega32E5 is done, the reset signal should stay activated until said otherwise.
I have the programmers AVRISP MK2 and Atmel ICE and tried both.
With both it says "avrdude warning: -E option not supported by this programmer type", but I can't find anywhere, which programmers support this.
So, can someone tell which programmer I should use?

stefanrueger:
Good question! Originally `-E` was implemented for parallel programmers (remember the parallel port for printers from the last millennium?). Only a couple of other programmers followed suit. Looking at the source code we have
```
$ grep "parseexitspecs *= *[^=]" -r src
src/pgm.c:  pgm->parseexitspecs = NULL;
src/par.c:  pgm->parseexitspecs = par_parseexitspecs;
src/linuxspi.c:    pgm->parseexitspecs = linuxspi_parseexitspecs;
src/flip2.c:  pgm->parseexitspecs   = flip2_parseexitspecs;
```

Maybe the hardware pros amongst the team can help out with alternative ways of doing what you want to achieve... @mcuee @MCUdude

Tropaion:
> remember the parallel port for printers from the last millennium?

I'm too young for that, but I think I know what you mean 😄 .
I currently have an working but not very nice solution. I keep an python thread running in the background which keeps avrdude open in terminal mode, but sometimes this creates some annoying issues.
This was also the reason I came upon the "-E" command while searching for alternatives.

mcuee:
@MCUdude

I think you are more familar with Microchip tools like Atmel ICE. Maybe you can help to see if we can offer such an options for some of the programmers.

mcuee:
@stefanrueger and @MCUdude 

I think we can probably implement such an option for programmers which can toggle the reset pin at different state. 

linuxgpio should be one candidate. @stefanrueger mentioned that linuxspi already has `-E` support.

We have some programmers which have similar functions, not using `-E`. I think serialupdi is one example. 
* https://github.com/libusb/libusb/issues/1166

MCUdude:
I'll look into what it would take to implement this for jtag3 programmers.

@dioannidis I don't know much about the USBasp firmware, but it is possible to leave the reset pin low after programming?

Tropaion:
@MCUdude That would be really nice, especially for Microchip tools.
[...]

Issue #1697, opened 2024-02-21
AVR109 auto reset functionality


MCUdude:
First pointed out [here](https://github.com/avrdudes/avrdude/discussions/1022#discussioncomment-8539144):

> My bootloader doesn't need an IO pin to enter bootloader mode. It automatically enters bootloader mode after the MCU resets itself and return to application after 500ms to 4s (the compiled hex has a 1 second delay), so the first upload is free, since there is no code in the application section. The second upload, you need to press the reset button. AVR109 doesn't issue an DTR or RTS signal on serial, so auto upload is not possible with avr109.

We should consider adding DTR/RTS toggling to the `avr109` programming option as we do with `urclock`, `arduino`, and `wiring`, as this would make it easier to enter bootloader mode using a simple auto-reset circuit.

MCUdude:
I've pushed a development branch here:
https://github.com/MCUdude/avrdude/tree/avr109-autoreset

However, I've not been able to find a bootloader that with auto reset, and doesn't work without it.
I've tried Xboot, and if I set `USE_ENTER_UART = yes` in the `arduino328p.conf.mk` file, it jumps straight into the bootloader even though no auto reset. And if I use `no`, I'm not able to enter bootloader mode, not even with auto reset present.

stefanrueger:
Your development branch looks neat and tidy
 - Does the usleep(250 * 1000) really need to be that long? 50 ms is plenty to discharge the reset cap (should work even for C = 10 µF and R = 5k). `-c arduino` needs that long sleep b/c the optiboot bootloader needs a 200 ms pause between two successive calls. `-c xbee` might need a long pause perhaps b/c it involves sending packets over the air, but the standard AVR109 board?
 - Would it not be better to keep default behaviour at *not* plucking the reset line?  I just imagine someone's board might blow up if DTR is suddenly toggled
 - Documentation would be cool

stefanrueger:
@MCUdude Do you want to pursue AVR109 auto reset further?

MCUdude:
I don't quite remember why I didn't continue working on this. But I remember using the Xboot bootloader (that uses the AVR109 protocol) and for some reason it would "just upload" even though there where no reset pulse. Does there a different AVR109 bootloader I can test with that would require manual or auto reset?

MCUdude:
I've done a little research, and I'd like to explain a little.

[...]

PR #1842, opened 2024-07-18
Provide terminal disasm command

stefanrueger:
The new terminal command `disasm` disassembles the selected memory section. It knows about the register file of the part, so you get these symbols for free. You can also provide an ASCII tag file (and generate one via `tools/elf2tag` from an `.elf` file).

A big shoutout to @johndoe31415 whose 17 year old `avrdisas` provides the base for this PR. It is well written, relatively easy to port, and it was relatively easy to add 8 opcodes that had cropped up in the meantime and to add some 20 or so unallocated opcodes that are known to behave in a certain way (the famous `0xffff` that behaves like `sbrs r31, 7`). Option `-d` shows the unallocated opcodes as `u/`, so, eg, `u/sbrs r31, 7`. Option `-D` (the default) switches back to only disassembling the opcodes that the architecture of the part knows.

Here an example: use the files in [urclock_cs8_ad.zip](https://github.com/user-attachments/files/16289987/urclock_cs8_ad.zip)
```
$ avrdude -qq -p m328p -c dryrun -U urclock_cs8_ad.hex -t
avrdude> disasm flash 0x7e00 32
.equ PORTB_DDRB, 0x4
.equ PORTB_PORTB, 0x5
.equ SPI_SPCR, 0x2c
.equ CPU_MCUSR, 0x34
7e00:   11 24           clr     r1
7e02:   c4 b7           in      r28, CPU_MCUSR
7e04:   2c 2e           mov     r2, r28
7e06:   14 be           out     CPU_MCUSR, r1
7e08:   80 e0           ldi     r24, 0x00       ; 0
7e0a:   bd d0           rcall   watchdogConfig
7e0c:   c1 fd           sbrc    r28, 1          ; 0x02 = 2
7e0e:   3e c0           rjmp    ldi_wdto
7e10:   c3 ff           sbrs    r28, 3          ; 0x08 = 8
7e12:   3b c0           rjmp    rjmp_application
7e14:   85 e1           ldi     r24, 0x15       ; 21
7e16:   85 b9           out     PORTB_PORTB, r24
7e18:   8d e2           ldi     r24, 0x2d       ; 45
7e1a:   84 b9           out     PORTB_DDRB, r24
7e1c:   80 e5           ldi     r24, 0x50       ; 80
7e1e:   8c bd           out     SPI_SPCR, r24

avrdude> disasm flash 0x7e00 32 -t=urclock_cs8_ad.tag 
.equ PORTB_DDRB, 0x4
[...]

PR #1844, opened 2024-07-22
Extended parameters (`-x` and `-E`) improvements

MCUdude:
Adresses #1697.

I've added the `-x autoreset` to the AVR109 programmer so that compatible bootloaders can utilize this feature. I've also added the `-x noautoreset` to the `arduino` and `wiring` programmers.

@stefanrueger I didn't touch the Urclock implementation, since the way it deals with extra parameters is so different than the other. It would be great if you could look into this!

This PR is still a draft, so feedback is welcome.

### DIFF
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -37,6 +37,45 @@
 #include "stk500.h"
 #include "arduino.h"
 
+static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
+  const char *extended_param;
+  int attempts;
+  int rv = 0;
+
+  for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
+    extended_param = ldata(ln);
+
+    if (sscanf(extended_param, "attempts=%2d", &attempts) == 1) {
+      PDATA(pgm)->retry_attempts = attempts;
+      pmsg_info("setting number of retry attempts to %d\n", attempts);
+      continue;
+    }
+
+    if(str_starts(extended_param, "no_autoreset")) {
+      if(!str_eq(extended_param, "no_autoreset")) {
+        pmsg_error("invalid -xno_autoreset value %s. Use -xno_autoreset\n", extended_param);
+        rv = -1;
+        break;
+      }
+      PDATA(pgm)->autoreset = false;
+      continue;
+    }
+
+    if (str_eq(extended_param, "help")) {
+      msg_error("%s -c %s extended options:\n", progname, pgmid);
+      msg_error("  -xattempts=<arg> Specify no. connection retry attempts\n");
+      msg_error("  -xno_autoreset   Don't toggle RTS/DTR lines on port open to prevent a hardware reset\n");
+      msg_error("  -xhelp           Show this help menu and exit\n");
+      return LIBAVRDUDE_EXIT;
+    }
+
+    pmsg_error("invalid extended parameter '%s'\n", extended_param);
+    rv = -1;
+  }
+
+  return rv;
+}
+
 /* read signature bytes - arduino version */
 static int arduino_read_sig_bytes(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m) {
   unsigned char buf[32];
@@ -85,28 +124,28 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
     return -1;
   }
 
-  // This code assumes a negative-logic USB to TTL serial adapter
-  // Set RTS/DTR high to discharge the series-capacitor, if present
-  serial_set_dtr_rts(&pgm->fd, 0);
-  /*
-   * Long wait needed for optiboot: otherwise the second of two bootloader
-   * calls in quick succession fails:
-   *
-   * avrdude -c arduino -qqp m328p -U x.hex; avrdude -c arduino -qqp m328p -U x.hex
-   */
-  usleep(250 * 1000);
-  // Pull the RTS/DTR line low to reset AVR
-  serial_set_dtr_rts(&pgm->fd, 1);
-  // Max 100 us: charging a cap longer creates a high reset spike above Vcc
-  usleep(100);
-  // Set the RTS/DTR line back to high, so direct connection to reset works
-  serial_set_dtr_rts(&pgm->fd, 0);
+  if(PDATA(pgm)->autoreset) {
+    // This code assumes a negative-logic USB to TTL serial adapter
+    // Set RTS/DTR high to discharge the series-capacitor, if present
+    serial_set_dtr_rts(&pgm->fd, 0);
+    /*
+    * Long wait needed for optiboot: otherwise the second of two bootloader
+    * calls in quick succession fails:
+    *
+    * avrdude -c arduino -qqp m328p -U x.hex; avrdude -c arduino -qqp m328p -U x.hex
+    */
+    usleep(250 * 1000);
+    // Pull the RTS/DTR line low to reset AVR
+    serial_set_dtr_rts(&pgm->fd, 1);
+    // Max 100 us: charging a cap longer creates a high reset spike above Vcc
+    usleep(100);
+    // Set the RTS/DTR line back to high, so direct connection to reset works
+    serial_set_dtr_rts(&pgm->fd, 0);
 
-  usleep(100 * 1000);
+    usleep(100 * 1000);
+  }
 
-  /*
-   * drain any extraneous input
-   */
+  // Drain any extraneous input
   stk500_drain(pgm, 0);
 
   if (stk500_getsync(pgm) < 0)
@@ -115,7 +154,7 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
   return 0;
 }
 
-static void arduino_close(PROGRAMMER * pgm)
+static void arduino_close(PROGRAMMER *pgm)
 {
   serial_close(&pgm->fd);
   pgm->fd.ifd = -1;
@@ -131,9 +170,12 @@ void arduino_initpgm(PROGRAMMER *pgm) {
   stk500_initpgm(pgm);
 
   strcpy(pgm->type, "Arduino");
+  PDATA(&pgm)->autoreset = true; // Auto-reset enabled by default
+
   pgm->read_sig_bytes = arduino_read_sig_bytes;
   pgm->open = arduino_open;
   pgm->close = arduino_close;
+  pgm->parseextparams = arduino_parseextparms;
 
   cx->avr_disableffopt = 1;     // Disable trailing 0xff removal
 }
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -45,7 +45,7 @@ static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
   for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
     extended_param = ldata(ln);
 
-    if (sscanf(extended_param, "attempts=%2d", &attempts) == 1) {
+    if (sscanf(extended_param, "attempts=%i", &attempts) == 1) {
       PDATA(pgm)->retry_attempts = attempts;
       pmsg_info("setting number of retry attempts to %d\n", attempts);
       continue;
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -38,13 +38,12 @@
 #include "arduino.h"
 
 static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
-  const char *extended_param;
   int attempts;
   int rv = 0;
   bool help = 0;
 
   for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
-    extended_param = ldata(ln);
+    const char *extended_param = ldata(ln);
 
     if (sscanf(extended_param, "attempts=%i", &attempts) == 1) {
       PDATA(pgm)->retry_attempts = attempts;
--- a/src/arduino.c
+++ b/src/arduino.c
@@ -40,7 +40,7 @@
 static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
   const char *extended_param;
   int attempts;
-  int rv = 0;
+  int rv = 0, help = 0;
 
   for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
     extended_param = ldata(ln);
@@ -57,17 +57,20 @@ static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
     }
 
     if (str_eq(extended_param, "help")) {
-      msg_error("%s -c %s extended options:\n", progname, pgmid);
-      msg_error("  -xattempts=<n> Specify the number <n> of connection retry attempts\n");
-      msg_error("  -xnoautoreset  Don't toggle RTS/DTR lines on port open to prevent a hardware reset\n");
-      msg_error("  -xhelp         Show this help menu and exit\n");
-      return LIBAVRDUDE_EXIT;
+      help = 1;
+      rv = LIBAVRDUDE_EXIT;
     }
 
-    pmsg_error("invalid extended parameter %s\n", extended_param);
-    rv = -1;
+    if (!help) {
+      pmsg_error("invalid extended parameter %s\n", extended_param);
+      rv = -1;
+    }
+    msg_error("%s -c %s extended options:\n", progname, pgmid);
+    msg_error("  -xattempts=<n> Specify the number <n> of connection retry attempts\n");
+    msg_error("  -xnoautoreset  Don't toggle RTS/DTR lines on port open to prevent a hardware reset\n");
+    msg_error("  -xhelp         Show this help menu and exit\n");
+    return rv;
   }
-
   return rv;
 }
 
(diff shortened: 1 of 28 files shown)


stefanrueger:
@MCUdude Very cool. Will review/add/comment in the next few days

src/arduino.c
```diff
@@ -37,6 +37,45 @@
...
 
+static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
+  const char *extended_param;
+  int attempts;
+  int rv = 0;
+
+  for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
+    extended_param = ldata(ln);
+
+    if (sscanf(extended_param, "attempts=%2d", &attempts) == 1) {
```
stefanrueger, line 48: `%2d` is a weird format, as it *only* reads 2 digits of the given number, but the user *doesn't know* and *it's hard to document*, so the user will be puzzled why `-xattempts=123` set the number of attempts to 12. Why not use `%i`? That allows hex numbers (`0x12`) or octal numbers (`077`), too. Normally one should also check that the number makes sense (must be positive here). Try `-xattempts=-1` and you'll see what I mean.

src/arduino.c
```diff
@@ -37,6 +37,45 @@
...
+      pmsg_info("setting number of retry attempts to %d\n", attempts);
+      continue;
+    }
+
+    if(str_starts(extended_param, "no_autoreset")) {
+      if(!str_eq(extended_param, "no_autoreset")) {
+        pmsg_error("invalid -xno_autoreset value %s. Use -xno_autoreset\n", extended_param);
+        rv = -1;
+        break;
+      }
```
stefanrueger, line 59: Why not replace these 6 lines with one
```
  if(str_eq(extended_param, "no_autoreset")) {
```
Note the `str_eq()` instead of the `str_starts()`. Anything that starts with `no_autoreset` but isn't `no_autoreset` is caught later on anyway (invalid parameter).

src/arduino.c
```diff
@@ -37,6 +37,45 @@
...
+      PDATA(pgm)->autoreset = false;
+      continue;
+    }
+
+    if (str_eq(extended_param, "help")) {
+      msg_error("%s -c %s extended options:\n", progname, pgmid);
+      msg_error("  -xattempts=<arg> Specify no. connection retry attempts\n");
```
stefanrueger, line 66: I'd write `-xattempts=<n> Specify the number <n> of connection retry attempts` for clarity. Also allows the option text to move left by two spaces

src/arduino.c
```diff
@@ -37,6 +37,45 @@
...
+      msg_error("  -xno_autoreset   Don't toggle RTS/DTR lines on port open to prevent a hardware reset\n");
+      msg_error("  -xhelp           Show this help menu and exit\n");
+      return LIBAVRDUDE_EXIT;
+    }
+
+    pmsg_error("invalid extended parameter '%s'\n", extended_param);
```
stefanrueger, line 72: In the last messaging clean-up the quotation marks tended to be removed, `"invalid extended parameter %s\n"` is the more recent AVRDUDE style.

src/arduino.c
```diff
@@ -115,7 +154,7 @@ static int arduino_open(PROGRAMMER *pgm, const char *port) {
   return 0;
 }
 
-static void arduino_close(PROGRAMMER * pgm)
+static void arduino_close(PROGRAMMER *pgm)
```
stefanrueger, line 157: If you're changing spaces, why not move the curly bracket up, too? :smile:

src/arduino.c
```diff
@@ -131,9 +170,12 @@ void arduino_initpgm(PROGRAMMER *pgm) {
   stk500_initpgm(pgm);
 
   strcpy(pgm->type, "Arduino");
+  PDATA(&pgm)->autoreset = true; // Auto-reset enabled by default
```
stefanrueger, line 173: This looks wrong (because it probably is): *Must* be `PDATA(pgm)` not `PDATA(&pgm)`.I think this sets an unknown memory address to 1 (but not `autoreset`).

src/butterfly.c
```diff
@@ -681,6 +694,37 @@ static int butterfly_read_sig_bytes(const PROGRAMMER *pgm, const AVRPART *p, con
...
 }
 
+static int butterfly_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
+  const char *extended_param;
+  int rv = 0;
+
+  for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
+    extended_param = ldata(ln);
+
+    if(str_starts(extended_param, "autoreset")) {
```
stefanrueger, line 704: see above

src/stk500.c
```diff
@@ -116,7 +116,7 @@ int stk500_getsync(const PROGRAMMER *pgm) {
 
   for (attempt = 0; attempt < max_sync_attempts; attempt++) {
     // Restart Arduino bootloader for every sync attempt
-    if (str_eq(pgm->type, "Arduino") && attempt > 0) {
+    if (PDATA(pgm)->autoreset && attempt > 0) {
```
stefanrueger, line 119: I think at this place the check `str_eq(pgm->type, "Arduino")` must remain. It's OK to have the *additional* constraint `PDATA(pgm)->autoreset` that the user controls. Only when the `-c` programmer is of the Arduino type (ie, if it's `-c arduino`) should the driver trigger reset here. I believe, this has to do with the weird way optiboot has at some point decided when to jump to the application and when to jump to the bootloader. Normal STK500 programmers don't need this and it would prossibly be harmful to the re-sync attempts.

src/wiring.c
```diff
@@ -106,10 +107,21 @@ static int wiring_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
       pmsg_notice2("%s(): delay set to %d ms\n", __func__, val);
       WIRINGPDATA(pgm)->delay = val;
       continue;
-    } else if (str_eq(extended_param, "help")) {
+    }
+    else if(str_starts(extended_param, "no_autoreset")) {
```
stefanrueger, line 111: See above

src/wiring.c
```diff
@@ -106,10 +107,21 @@ static int wiring_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
...
+      if(!str_eq(extended_param, "no_autoreset")) {
+        pmsg_error("invalid -xno_autoreset value %s. Use -xno_autoreset\n", extended_param);
+        rv = -1;
+        break;
+      }
+      WIRINGPDATA(pgm)->autoreset = false;
+      continue;
+    }
+    else if (str_eq(extended_param, "help")) {
```
stefanrueger, line 120: I know `<arg>` below is in existing code. But it's always nicer to be as specific as possible. Arg can be anything, but `<n>` alludes to this being a number. So  `-xsnooze=<n> Wait snooze <n> ms` is defo more specific. Same with `delay` below.

src/wiring.c
```diff
@@ -170,8 +182,7 @@ static int wiring_open(PROGRAMMER *pgm, const char *port) {
   return 0;
 }
 
-static void wiring_close(PROGRAMMER * pgm)
-{
+static void wiring_close(PROGRAMMER *pgm) {
```
stefanrueger, line 185: Cool!

stefanrueger:
@MCUdude You'll have seen a couple of review comments. What I haven't done is compiling or testing the code (but you have the scope to see the number of auto-reset spikes, so that's best verified by yourself that the code does what you want it to do).

I'll push code for urclock's `-xnoautoreset` functionality. There is a ton of other `no...` extended options *without* the `_` between `no` and feature that is being switched off. So, for consistency in `urclock.c` I'll drop the `_`.

stefanrueger:
And... this PR requires documentation in two places: `avrdude.1` and `avrdude.texi`

src/arduino.c
MCUdude: Sure! `%2d` is just copy-paste from stk500.c, but you have a point!

src/arduino.c
MCUdude: It works this way, and I got a segfault if not. I'll look into it again

src/stk500.c
MCUdude: Yes, but we only provide auto-reset functionality for the `-c arduino programmer`. So this should not run when using just another stk500v1 based programmer

MCUdude:
Thanks for your feedback @stefanrueger! I have some comments though.

BTW one thing we should look into that would make the extended parameters more helpful to a struggling user would be to print the `-xhelp`output when the user provided an incorrect extended parameter. At the moment it's just an error, with no additional information. I think printing the help menu when there is an error could be helpful.

src/arduino.c
stefanrueger: I think I know what happens: `arduino_initpgm()` is called *before* `pgm->setup()` which allocates the pdata structure. So, `pgm->cookie` is NULL here (as it hasn't been initialises yet) and you get a seg-fault when you access it. `PDATA(&pgm)` is a memory location in the PROGRAMMER sructure which is at the offset of autoreset from where the cookie pointer lives. This is defo not what you want! *Either* set `PDATA(pgm)->autoreset` in `pgm->setup()` or later but before the parameters are parsed. *Or* invert the logic: Use a variable `noautoreset` that gets automatically initialised to zero. So you don't need to initialise that.

src/arduino.c
stefanrueger: Thanks. This means the `stk500.c` code also needs changing (as that too is wrong)

src/stk500.c
stefanrueger: That is a good point. But just imagine someone *adds* the autoreset functionality for *another* stk500.c based programmer (just like arduino did that). Then this code would run for that programmer, too. But I don't think this will be needed. I think it's *only* optiboot that needs that code at this point. `str_eq(pgm->type, "Arduino")` is kinda a proxy for optiboot (the best AVRDUDE can do outside what urclock has been doing by examining the bootloader itself and computing hash sums).

TLDR; I would use `if(str_eq(pgm->type, "Arduino") && PDATA(pgm)->autoreset && attempt > 0)` to document it's intended for `-c arduino` even though this might be redundant right now.

stefanrueger:
> more helpful to a struggling user would be to print the `-xhelp` output 

Vvv good point! It's a lot of work to add this to what feels like 1500 programmers, but a great idea. (I think urclock already does this, so it's only needed for 1499 programmers :)

src/arduino.c
stefanrueger: > stk500.c code also needs changing

You already did! Apologies for stating the obvious.

src/stk500.c
MCUdude: Good point! Thanks for the explaination

stefanrueger: Actually, thinking about this more: as autoreset is introduced in the stk500.c PDATA and (one way or another) defaults to true this code *will* run for *all* stk500 programmers changing behaviour. It defo needs to be conditioned on Arduino.

MCUdude:
> Vvv good point! It's a lot of work to add this to what feels like 1500 programmers, but a great idea. (I think urclock already does this, so it's only needed for 1499 programmers :)Vvv good point! It's a lot of work to add this to what feels like 1500 programmers, but a great idea. (I think urclock already does this, so it's only needed for 1499 programmers :)

How about something like this (separate print function):

```c
static void extended_param_helptext() {
  msg_error("%s -c %s extended options:\n", progname, pgmid);
  msg_error("  -xattempts=<n> Specify the number <n> of connection retry attempts\n");
  msg_error("  -xnoautoreset  Don't toggle RTS/DTR lines on port open to prevent a hardware reset\n");
  msg_error("  -xhelp         Show this help menu and exit\n");
}

static int arduino_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
  const char *extended_param;
  int attempts;
  int rv = 0;

  for (LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
    extended_param = ldata(ln);

    if (sscanf(extended_param, "attempts=%i", &attempts) == 1) {
      PDATA(pgm)->retry_attempts = attempts;
      pmsg_info("setting number of retry attempts to %d\n", attempts);
      continue;
    }

    if(str_eq(extended_param, "noautoreset")) {
      PDATA(pgm)->autoreset = false;
      continue;
    }

    if (str_eq(extended_param, "help")) {
      extended_param_helptext();
      return LIBAVRDUDE_EXIT;
    }

    pmsg_error("invalid extended parameter %s\n", extended_param);
    extended_param_helptext();
    rv = -1;
  }

  return rv;
}
```

This will result in the following output:
```
$ ./avrdude -carduino -patmega328p -P /dev/cu.usbmodem14201 -xinvalid
avrdude error: invalid extended parameter invalid
avrdude -c arduino extended options:
  -xattempts=<n> Specify the number <n> of connection retry attempts
  -xnoautoreset  Don't toggle RTS/DTR lines on port open to prevent a hardware reset
  -xhelp         Show this help menu and exit
avrdude error: unable to parse extended parameter list
```

stefanrueger:
> How about something like this

Output looks great; and yes, both ways would work.

MCUdude:
@stefanrueger I've now pushed a fix for all programmers that supports extended parameters. I've done some minor code tweaks, but the main feature is that the help text gets printed if an invalid parameter is specified. Please review!

MCUdude:
BTW I'll update the docs one #1842 is merged, to prevent merge conflicts.

stefanrueger:
@MCUdude I think your technique of printing the help message on invalid parameters is very elegant. Great code! I particularly like  that the help message code resides *inside* the `for()` loop. This allows those valid `-x` parameters with invalid assignment values to  `break;` out of the look forgoing the full help message (which would distract from the nature of the invalid value).

Seeing that this is a really neat usability improvement for AVRDUDE, would you want to extend your technique to the parsing of exit parameters (`-E`)? It's three places
 - `par_parseexitspecs()`
 - `flip2_parseexitspecs()`
 - `linuxspi_parseexitspecs()`

> Please review!

Great code; a couple of comments
 - Please double check `jtagmkII_parseextparms()`: `help` never set; help preamble `msg_error("%s -c %s extended options:\n", progname, pgmid);` output in wrong place. I predict `avrdude -c jtag2 -xhelp` won't be what you want.
 - `WIRINGPDATA(&pgm)->autoreset = true;` is *still* a problem
 - I'd prefer `str_eq(pgm->type, "Arduino")` over `str_eq(pgmid, "arduino")` in `stk500_setup()` to be consistent when Arduino-like programmers are meant

> fix for all programmers that supports extended parameters

All? The commit overlooked
 - `serprog_parseextparms()`
 - `stk500v2_parseextparms()`
 - `usbasp_parseextparms()`

While reviewing the `xyz_parseextendedparms()` functions I also noticed that some of the error messages for invalid values on valid parameters did not make the context sufficiently clear (so a user who has carefully built a 2 mile long CLI line might struggle to figure out to which part of the command line the message actually refers to). I took the liberty to push a commit that makes those error messages clearer mostly by saying `Invalid value in -x <user setting>`. Also massaged the help messages a bit, eg, inserting a space between `-x` and the parameter. 

Please review this commit!

I also happened upon a bug in old code where `str_eq(str_eq(extended_param, "jtagchain=")` clearly needed to be `str_starts(extended_param, "jtagchain=")`. Fixed.

MCUdude:
Thanks for the feedback @stefanrueger! I've done some work to `-E` as well, and I've added help text here as well (`-E help`). But it would be great if you had any ideas for better help text for these exitspec parameters. The ones I came up with are a bit too verbose I think.

src/serprog.c
```diff
@@ -482,14 +481,18 @@ static int serprog_parseextparams(const PROGRAMMER *pgm, const LISTID extparms)
     }
 
     if(str_eq(extended_param, "help")) {
-      msg_error("%s -c %s extended options:\n", progname, pgmid);
-      msg_error("  -x cs=<n> Sets the chip select line to CS<n> for supported programmers\n");
-      msg_error("  -x help   Show this help menu and exit\n");
+      help = true;
```
stefanrueger, line 484: The next line needs to be `rv = LIBAVRDUDE_EXIT;` not `return ...`

I recommend to checking all 22 treated programmers by running
 ```
avrdude -c ... -x invalid
avrdude -c ... -x help
```
to see whether there are surprises.

File: src/dryrun.c [hunk 13]
```diff
@@ -1036,42 +1036,53 @@ static void dryrun_teardown(PROGRAMMER *pgm) {
...
+
   for(LNODEID ln = lfirst(extparms); ln; ln = lnext(ln)) {
     const char *xpara = ldata(ln);
 
     if(str_starts(xpara, "init")) {
       dry.init = 1;
       continue;
-    } else if(str_starts(xpara, "random")) {
+    }
+    if(str_starts(xpara, "random")) {
```
stefanrueger, line 1049: BTW, really well spotted that a `continue` should not be `else`d. Neat!

src/jtag3.c
```diff
@@ -1610,9 +1610,10 @@ static int jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extparms) {
       rv = -1;
     }
     msg_error("%s -c %s extended options:\n", progname, pgmid);
-    if(str_eq(pgm->type, "JTAGICE3"))
+    if(str_eq(pgm->type, "JTAGICE3")) {
```
stefanrueger, line 1613: Great catch!

stefanrueger (approved):

MCUdude:
I've just tested all the programmers that support extended parameters. Things are looking good! Docs need to be updated, but I think it's best to wait until #1842 has been merged.

MCUdude:
Speaking of docs, now is a good time to finally resolve #1711 by adding a new chapter or section to the avrdude.texi docs here?

image

PR #1850, opened 2024-07-31
Remove obsolete SVN $Id$ tags from source files

ndim:
Remove SVN `$Id$` tags from source files which have been made obsolete by moving to git and serve no purpose any more.

This mostly removes the matches printed by the command

    git grep -E '\$[A-Za-z]+\$'

The remaining exceptions are some binary files (PDF and font files), and jquery js files with both `$Id$` and `$Date$` tags.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,3 @@
-$Id$
-
 Approximate change log for AVRDUDE by version.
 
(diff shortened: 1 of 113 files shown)


stefanrueger:
Neat! I think I can merge that with today's mergefest.

PR #1853, opened 2024-08-01
Add missing documentation from #1844

MCUdude:
Just a small documentation update.

@stefanrueger do you want me to create a dedicated section in the Texinfo document (Under [2. Command line options](https://avrdudes.github.io/avrdude/7.3/avrdude_3.html#Command-Line-Options)) where I list the -E exitspec arguments along with supported programmers, similar to what we have done with the -x extended parameters?

### DIFF
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -502,18 +502,18 @@ case before programming the cell.
 .Xc
 Pass
 .Ar exitspec
-to the chosen programmer. The interpretation of the exitspec arguments
+to the chosen programmer. The interpretation of the exitspec parameter
 depends on the programmer itself. See below for a list of programmers
-accepting extended parameters or issue
+accepting exitspec parameter options or issue
 .Nm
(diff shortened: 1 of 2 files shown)


stefanrueger:
> dedicated section in the Texinfo document [for] the -E exitspec

That would be brill!

MCUdude:
Should I call them "exitspec arguments" or "exitspec parameters"? We already call -x "extended parameters".

stefanrueger:
exitspec parameter (note: no plural)

In contrast to `-x` extended parameters that can appear multiple times on the command line, the exitspec option `-E` can only be usefully specified once: if is specified multiple times the last one prevails. The `-E` parameter is a comma separated list of ...

MCUdude:
Thanks for the feedback @stefanrueger! I've pushed some changes to avrdude.texi as well. Feel free to proofread the text I've committed, and don't be afraid to push changes directly to my fork if there are any obvious errors or mistakes.

PR #1854, opened 2024-08-01
Provide -P usb:vid:pid for USB hid i/f and allow long -P usb:serno

### LINKED ISSUE #1838: JTAGICE3 EDBG communication mode relies on hardcoded USBVID
Hello everyone

I am currently creating a jtagice3_*** compatible firmware that works with the AVR-DU series. It can be recognized as USB-HID + USB-CDC using only the standard drivers for Windows and macOS without installing any special drivers. (For the time being, the goal is to support UPDI and TPI, and debugWire cannot be used.)

To meet this condition and be recognized as a programmer by AVRDUDE, it must be in EDBG (CMSIS-DAP) communication mode, at least like pkobn_updi. However, I noticed that it is hard-coded to change to EDBG only when usbvid is *0x03EB* ('ATMEL'). Changing `usbvid` in `.avrduderc` does not work, and it is always fixed to `max_xfer = USBDEV_MAX_XFER_3 (912)`. The conditional branch is in `usb_hidapi.c:173`.

I could redesign the firmware to create a dedicated endpoint that can handle 912 byte packets, but I personally don't like the hard-coded part. This wouldn't have been an issue before. All the chips with jtagice3 were "ATMEL" branded.

The AVR-DU series is a "Microchip" (MCHP) product, so I'm reluctant to use the "ATMEL" VID. I'd like to use at least *0x04DB* (MCHP) or another VID if possible.

Any ideas on how to solve this problem? It seems like I need to add a configuration option or internal flag so that the condition at `usb_hidapi.c:173` is always true.

stefanrueger:
Fixes #1838

 - New port mode `-P usb:<vid>:<pid>` for USB hid interfaces (`jtag3.c`)
 - Allow 63 digits instead of 12 digit serial numbers for `-P usb:serialno` (libusb and hidapi)
 - Admit Microchip vid in addition to Atmel vid (both hard-coded)
 - Add `fd->usb.eep = 0;` setting for CMSIS-DAP programmers as dected by product string

Note that the PR also sets `fd->usb.max_xfer = 64;` but this will be overwritten further down if Atmel or Microchip vid. Awaiting comment from @askn37 

Needs checking with `jtag3.c` programmers. Any takers?

### DIFF
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -500,72 +500,20 @@ case before programming the cell.
 .It Xo Fl E Ar exitspec Ns
 .Op \&, Ns Ar exitspec
 .Xc
-By default,
-.Nm
-leaves the parallel port in the same state at exit as it has been
-found at startup.  This option modifies the state of the
-.Ql /RESET
-and
-.Ql Vcc
-lines the parallel port is left at, according to the
+Pass
 .Ar exitspec
-arguments provided, as follows:
-.Bl -tag -width noreset
-.It Ar reset
-The
-.Ql /RESET
-signal will be left activated at program exit, that is it will be held
-.Em low ,
-in order to keep the MCU in reset state afterwards.  Note in particular
-that the programming algorithm for the AT90S1200 device mandates that
-the
-.Ql /RESET
-signal is active
-.Em before
-powering up the MCU, so in case an external power supply is used for this
-MCU type, a previous invocation of
+to the chosen programmer. The interpretation of the exitspec parameter
+depends on the programmer itself. See below for a list of programmers
+accepting exitspec parameter options or issue
 .Nm
-with this option specified is one of the possible ways to guarantee this
-condition.
-.Em reset
-is supported by the linuxspi and flip2 programmer options, as well as all
-parallel port based programmers.
-.It Ar noreset
-The
-.Ql /RESET
-line will be deactivated at program exit, thus allowing the MCU target
-program to run while the programming hardware remains connected.
-.Em noreset
-is supported by the linuxspi and flip2 programmer options, as well as all
-parallel port based programmers.
-.It Ar vcc
-This option will leave those parallel port pins active
-.Pq \&i. \&e. Em high
-that can be used to supply
-.Ql Vcc
-power to the MCU.
-.It Ar novcc
-This option will pull the
-.Ql Vcc
-pins of the parallel port down at program exit.
-.It Ar d_high
-This option will leave the 8 data pins on the parallel port active.
-.Pq \&i. \&e. Em high
-.It Ar d_low
-This option will leave the 8 data pins on the parallel port inactive.
-.Pq \&i. \&e. Em low
-.El
-.Pp
-Multiple
-.Ar exitspec
-arguments can be separated with commas.
+-E help ... to see the options of the chosen programmer.
 .It Fl F
 Normally,
 .Nm
 tries to verify that the device signature read from the part is
 reasonable before continuing.  Since it can happen from time to time
 that a device has a broken (erased or overwritten) device signature
-but is otherwise operating normally, this options is provided to
+but is otherwise operating normally, this option is provided to
 override the check.
 Also, for programmers like the Atmel STK500 and STK600 which can
 adjust parameters local to the programming tool (independent of an
@@ -609,7 +557,7 @@ No-write: disables writing data to the MCU whilst processing -U
 .Nm avrdude
 ). The terminal mode continues to write to the device.
 .It Fl O
-Perform a RC oscillator run-time calibration according to Atmel
+Perform an RC oscillator run-time calibration according to Atmel
 application note AVR053.
 This is only supported on the STK500v2, AVRISP mkII, and JTAG ICE mkII
 hardware.
@@ -828,7 +776,6 @@ of that by erasing pages before programming them unless -e (chip erase) or
 -D (do not erase before writing) was requested. It should be noted that in
 absence of the -e chip erase option any ATxmega or UPDI flash pages not
 affected by the programming will retain their previous content.
-
 .Pp
 Classic devices may have the following memories in addition to eeprom, flash, signature and lock:
 .Bl -tag -width "  calibration" -compact
@@ -1854,6 +1801,13 @@ original STK500. Used by avrdude for the correct calculation of fosc and sck.
 .It Ar help
 Show help menu and exit.
 .El
+.It Ar AVR109
+.Bl -tag -offset indent -width indent
+.It Ar autoreset
+Toggle RTS/DTR lines on port open to issue a hardware reset.
+.It Ar help
+Show help menu and exit.
+.El
 .It Ar AVR910
 .Bl -tag -offset indent -width indent
 .It Ar devcode=VALUE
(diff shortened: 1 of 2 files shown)


askn37:
> Note that the PR also sets fd->usb.max_xfer = 64; but this will be overwritten further down if Atmel or Microchip vid. Awaiting comment from @askn37

First, we cannot determine whether the current usbhid_open is really the correct implementation. Let's assume that it is.

usbhid_open() first sets the following:

```
if (fd->usb.max_xfer == 0)
    fd->usb.max_xfer = USBDEV_MAX_XFER_3;
```

USBDEV_MAX_XFER_3 is a constant value of 912.

This has been in the code for a long time. If left this way, HIDAPI-related parts will fail here and there. Therefore, usbhid_open tries to look for "something else" and fix usb.max_xfer. Maybe that's the intention.

For Arm-designed CMSIS-DAP compatible specifications, it is at least 64. For *Standard-Speed*, it may be 16, but it is not a problem because "ATMEL vendor extension command (EDBG)" only works at 64. Look at jtag3_edbg_send() and jtag3_edbg_prepare(). If hid_get_product_string() contains `CMSIS-DAP`, follow it. This is the first code addition. The same test is also in usb_libusb.c, so you can trust it. Curiosity and Xplained pass this test. SNAP probably passes too. Any VID:PID is allowed.

Once you know it's CMSIS-DAP compatible you can get the correct packet size with the DAP-Info command.

So in the old code, you check if VID==ATMEL. If true, query the DAP-Info command and fix us.max_xfer to always 64 or 512. If there's no response, you port close, so it seems to assume that "ATMEL products always respond to DAP-Info". For ATMELICE3 it's 512, for Curiosity and Xplained it's 64. I don't know which it will be for PowerDebugger, but the test should pass. This time I'll add VID==MCHP here.

A bit strange, right? The initial usb.max_xfer is not used as long as the VID is ATMEL. If the returned value of DAP-Info is greater than USBDEV_MAX_XFER_3, it is rounded to that value.

This suggests that the initial value may be wrong. If it was originally 64, it seems good to test DAP_Info first, regardless of VID. If it fails, it may be USBDEV_MAX_XFER_3.

What doesn't seem to help at the moment is that ATMELICE3 doesn't work when linking up at *Full-Speed*. The link has to be up at least at *High-Speed*, but there seems to be no way for HIDAPI to know that.

mcuee:
@askn37 

There is for sure an issue here with the USB HID packet size. But somehow hdiapi is okay but libusb is not okay.

Reference:
* https://github.com/avrdudes/avrdude/issues/1221

mcuee:
> What doesn't seem to help at the moment is that ATMELICE3 doesn't work when linking up at Full-Speed. The link has to be up at least at High-Speed, but there seems to be no way for HIDAPI to know that.

Indeed HIDAPI may not know. libusb will know though. But we can not use libusb and hdiapi together under macOS (possible under Linux and Windows).

@Youw 
Just wondering if you have a suggestion here? Thanks.

stefanrueger:
OK, after carefully reading all comments (thank you), I think this PR is fine without the need for further changes. I'll merge soon unless there tests show that this is not advisable. I have tried with the [micoUPDI](https://github.com/MCUdude/microUPDI) programmer.

Without the PR:
```
$ sudo avrdude -p atmega4808 -c xplainedmini_updi -t -P usb:MICROUPDIPROGRAMMERX
avrdude error: invalid serial number MICROUPDIPROGRAMMERX
avrdude error: invalid serial number MICROUPDIPROGRAMMERX
avrdude error: no device found matching VID 0x03eb and PID list: 0x2145 with SN MICROUPDIPROGRAMMERX
avrdude error: unable to open port usb:MICROUPDIPROGRAMMERX for programmer xplainedmini_updi

avrdude done.  Thank you.

$  sudo avrdude -p atmega4808 -c xplainedmini_updi -t -P usb:03eb:2145
avrdude error: no matching device found
avrdude error: no device found matching VID 0x03eb and PID list: 0x2145 with SN 03eb2145
avrdude error: unable to open port usb:03eb2145 for programmer xplainedmini_updi

avrdude done.  Thank you.
```

With the PR

```
$ sudo avrdude -p atmega4808 -c xplainedmini_updi -t -P usb:MICROUPDIPROGRAMMERX
avrdude: device signature = 1E 96 50 (ATmega4808)
avrdude> 

$ sudo avrdude -p atmega4808 -c xplainedmini_updi -t -P usb:03eb:2145
avrdude: device signature = 1E 96 50 (ATmega4808)
avrdude> 
```

Youw:
> Just wondering if you have a suggestion here? Thanks.

I do not follow all of the context here, but if I got it right, and the problem is with HIDAPI+libusb backend and some device doesn't work correctly at Full-Speed (and probably has reports larger than 64 bytes if I'm not mistaken) - then there is no workaround at this moment. libusb backend of HIDAPI would have to be greately improved for that to work right.

askn37:
@stefanrueger 

> OK, after carefully reading all comments (thank you), I think this PR is fine without the need for further changes. I'll merge soon unless there tests show that this is not advisable. I have tried with the [micoUPDI](https://github.com/MCUdude/microUPDI) programmer.

Thank you for supporting the proposal.

This PR will be useful for future projects. At least, I am currently preparing a firmware for a UPDI/TPI programmer that imitates PICKIt4 and easily works with "Microchip AVR64DU32 Curiosity Nano", but I could not release it as open source because copying the VID and including it in the source code may violate MAPLAB's rights.

In the process of creating the firmware, I did a code review of jtag3.c and picked up the functions used in PICKit4, and this proposal was the most concerning. At least on macOS and Linux, this PR is a plus. I will continue to verify it because it may not be enough on Windows yet.

stefanrueger:
@askn37 The AVRDUDE project is very happy to support new open-source development of programmer FW.

Thanks, too, for reporting bugs and sharing your proposals for improving AVRDUDE. Always good to learn from expert analysis!

mcuee:
> > Just wondering if you have a suggestion here? Thanks.
> 
> I do not follow all of the context here, but if I got it right, and the problem is with HIDAPI+libusb backend and some device doesn't work correctly at Full-Speed (and probably has reports larger than 64 bytes if I'm not mistaken) - then there is no workaround at this moment. libusb backend of HIDAPI would have to be greately improved for that to work right.

Thanks. In this case, I think we can live with the following mentioned by @askn37.
> What doesn't seem to help at the moment is that ATMELICE3 doesn't work when linking up at Full-Speed. The link has to be up at least at High-Speed, but there seems to be no way for HIDAPI to know that.

We can document this as a limitation. 
**_Atmel ICE and Power Debugger have to be connected as High Speed USB device. They will not work when connected to Full Speed USB host._**

stefanrueger:
> Atmel ICE and Power Debugger have to be connected as High Speed USB device

There is code in the `usb_hidapi.c` open function that facilitates that. I don't know what's involved to carry over that code to the `usb_libusb.c` open function. I know next to nothing about USB. @MCUdude can you try?

MCUdude:
> There is code in the usb_hidapi.c open function that facilitates that. I don't know what's involved to carry over that code to the usb_libusb.c open function. I know next to nothing about USB. @MCUdude can you try?

Why is High-speed USB only a problem? High-speed USB has been around for quite a while.

stefanrueger:
> Why is High-speed USB only a problem?

I think we all talk about cross-purposes in the wrong place. This refers to #1221, I think... I won't get involved in that.

PR #1855, opened 2024-08-01
Add more WCH USB to serial chips

MCUdude:
I just got a CH342 USB to dual serial port adapter in the mail. It seems like a neat chip, so I added it to avrdude.conf.  When looking for its official product page on the WCH manufacturer website, I stumbled across a whole range of chips from WCH that we currently "don't support", or recognise if you want. And since the [Linux driver was easy to obtain](https://github.com/WCHSoftGroup/ch343ser_linux/blob/main/driver/ch343.c), I found the corresponding USB PIDs for the different chips.

One annoying thing with these dual/quad serial port chips is that they (obviously) share the same serial number, so the only way to distinguish them is to use their full /dev path, and not their `ch34x` name:

```
$ ./avrdude -carduino -patmega328p -P?
avrdude OS error: cannot open port ?: No such file or directory
avrdude error: unable to open port ? for programmer arduino
Possible candidate serial ports are:
  -P /dev/cu.Bluetooth-Incoming-Port
  -P /dev/cu.usbmodem58760071941 (via ch342 serial adapter)
  -P /dev/cu.usbmodem58760071943 (via ch342 serial adapter)
Note that above ports might not be connected to a target board or an AVR programmer.
Also note there may be other direct serial ports not listed above.

avrdude done.  Thank you.
```

MacOS system profile (Unlike the CH340, the CH342 has a serial number):

image


stefanrueger:
@MCUdude Cool as! Thanks for the contribution

mcuee:
Looks good to me.

PR #1856, opened 2024-08-01
cmake: LIB_NCURSES also accept pdcurses

ndim:
The Fedora mingw environment comes with pdcurses, which states

> Note that ncurses is not available for MinGW / Windows.
> Applications which need curses functionality can use this
> package, provided they don't use any of the extensions
> specific to ncurses.

This at least allows the cross compile build to succeed, I have not tested the executable under Windows or Wine.


stefanrueger:
Thanks for the input. I am sure @mcuee is in a position to test Windows

mcuee:
@ndim 

As I do not use Fedora or cross-compile myself, please upload the Windows binary here and I can carry out the test.

PR #1857, opened 2024-08-01
doc: Stop building and installing .ps and .dvi

ndim:
Building info, pdf, and HTML is enough.

ps and dvi are not needed any more.


mcuee:
The change seems to be good based on the github action runs.

stefanrueger:
Not merging now as earmarked as draft

ndim:
Could not change it from draft to non-draft from the GitHub Android client, had to wait to get back to a proper computer.

stefanrueger:
@ndim Thanks, will merge next round.

PR #1865, opened 2024-08-03
Provide current avrdude.pdf documentation in repo root

### LINKED ISSUE #1847: Auto-build avrdude.pdf from texinfo sources
_Originally posted by @stefanrueger in https://github.com/avrdudes/avrdude/issues/1842#issuecomment-2258214165 but I am creating a new issue as this is very much unrelated to that issue. The following words are from @stefanrueger, not mine._

It occurred to me that the `avrdude.pdf` documentation is installed somewhere on the user's destination system, but (as far as I am aware!) it *isn't available* in the avrdude repository for reference.

So, when discussing issues it is not possible for us maintainers to refer to the documentation of the current git-main avrdude other than saying something like "Find out where avrdude.pdf is installed on *your* system and look at Section 3.2"... This is *bizarre*.

What would be a good way to keep a current `avrdude.pdf` document (ideally created automatically when commits/merges happen to git main) in the repository? Ideally this would reside in a top-level directory such as `documentation`. Thoughts? @ndim @mcuee @dl8dtl @MCUdude

stefanrueger:
Closes #1847 but doesn't fix it.

I think it's easy enough to update avrdude.pdf from time to time, eg, after each mergefest, and update manually seeing all the difficulties that automation brings with it as discussed in #1847


mcuee:
Looks good to me.

ndim:
I would expect (about 98% confident) that only one person ever updating `avrdude.pdf` and only in the `main` branch should prevent merge conflicts on `avrdude.pdf`.

mcuee:
I am fine with this PR.

During the release time make sure to update the file as well. Probably we can add this point to the release checklist.

PR #1867, opened 2024-08-04
Provide dev option `-c \*/u` that shows udev rules

### LINKED ISSUE #1860: Does avrdude need to run as root in Linux for programming via `-c xplainedmini_updi`?
The [micoUPDI](https://github.com/MCUdude/microUPDI) programmer poses as `-c xplainedmini_updi` programmer. On my Linux I get permission errors when using it, see below. I assume this is the same problem for all programmers derived from `jtag3.c` : ATMEL ICE, snap, xplained boards etc. I also assume the problem originates in the USB HID mode that `jtag3.c` uses.

```
$ avrdude  -p atmega4808 -c xplainedmini_updi -t

avrdude error: cannot read serial number: error sending control message: Operation not permitted
avrdude error: cannot read product name: error sending control message: Operation not permitted
avrdude warning: unable to set configuration 1: could not set config 1: Operation not permitted
avrdude error: unable to claim interface 0: could not claim interface 0: Operation not permitted
avrdude error: unable to claim interface 1: could not claim interface 1: Operation not permitted
avrdude error: unable to claim interface 2: could not claim interface 2: Operation not permitted
avrdude warning: no usable interface found
avrdude error: no device found matching VID 0x03eb and PID list: 0x2145
avrdude error: unable to open port usb for programmer xplainedmini_updi

avrdude done.  Thank you.
```

When running avrdude as root things work like a charm:
```
$ sudo avrdude  -p atmega4808 -c xplainedmini_updi -t
avrdude: device signature = 1E 96 50 (ATmega4808)
avrdude> 
```

I suspect the problem is with the Linux drivers, but no matter which udev rule I enter in my `/etc/udev/rules.d/13-xplainedmini` I can only use that connection when I make myself root.  Here the current contents of that file:
```
SUBSYSTEM=="usb", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2145", MODE="0660", GROUP="plugdev", TAG+="uaccess"
# SUBSYSTEM=="hidraw", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2145", MODE="0660", GROUP="plugdev", TAG+="uaccess"
# KERNEL=="hidraw*", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2145", MODE="0660", GROUP="plugdev", TAG+="uaccess"
```

Now I realise this may or may not be an AVRDUDE problem (likely not) but it is not great practice to run avrdude as root and having to change all root ownership of file that avrdude created.

Does anyone have a solution for this?

stefanrueger:
Closes #1860 

Wildcards for the programmer name are allowed:
```
 avrdude -c \jtag*/u
# 1. Put Linux udev rules into /etc/udev/rules.d/55-avrdude.rules (or similar)
# 2. sudo udevadm control --reload-rules && sudo udevadm trigger (or similar)
# 3. Unplug the device and plug it in again

# jtag2dw, jtag2fast, jtag2, jtag2isp, jtag2pdi, jtag2slow, jtagmkII, jtagmkII_avr32, jtag2avr32
SUBSYSTEM=="usb", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2103", MODE="0660", TAG+="uaccess"

# jtag3, jtag3dw, jtag3isp, jtag3pdi, jtag3updi
SUBSYSTEM=="usb", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2110", MODE="0660", TAG+="uaccess"
KERNEL=="hidraw*", SUBSYSTEM=="hidraw", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2110", MODE="0660", TAG+="uaccess"
SUBSYSTEM=="usb", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2140", MODE="0660", TAG+="uaccess"
KERNEL=="hidraw*", SUBSYSTEM=="hidraw", ATTRS{idVendor}=="03eb", ATTRS{idProduct}=="2140", MODE="0660", TAG+="uaccess"

# jtagkey
SUBSYSTEM=="usb", ATTRS{idVendor}=="0403", ATTRS{idProduct}=="cff8", MODE="0660", TAG+="uaccess"
```

### DIFF
--- a/src/developer_opts.c
+++ b/src/developer_opts.c
@@ -1586,9 +1586,9 @@ void dev_output_pgm_defs(char *pgmidcp) {
   }
 
   if(udev && ui) {
-    dev_info("# 1. Put Linux udev rules into /etc/udev/rules.d/55-avrdude.rules (or similar)\n");
-    dev_info("# 2. sudo udevadm control --reload-rules && sudo udevadm trigger (or similar)\n");
-    dev_info("# 3. Unplug the device and plug it in again\n");
+    dev_info("# 1. Put Linux udev rules into, eg, /etc/udev/rules.d/55-avrdude.rules\n");
+    dev_info("# 2. Unplug the device and plug it in again\n");
+    dev_info("# 3. Enjoy user access to the USB programmer\n");
     qsort(udr, ui, sizeof *udr, udev_cmp);
     char *prev_head = mmt_strdup("<none>");
     for(Dev_udev *u = udr; u-udr < ui; u++) {
--- a/src/developer_opts.c
+++ b/src/developer_opts.c
@@ -1381,8 +1381,9 @@ static void dev_pgm_strct(const PROGRAMMER *pgm, bool tsv, const PROGRAMMER *bas
   for(int i=0; i<N_PINS; i++) {
     const char *str = pins_to_str(pgm->pin+i);
     const char *bstr = base? pins_to_str(base->pin+i): NULL;
-    if(!base || !str_eq(bstr, str))
-      _pgmout_fmt(avr_pin_lcname(i), "%s", str);
+    const char *pinname = avr_pin_lcname(i);
+    if((!base || !str_eq(bstr, str)) && !str_eq(pinname, "<unknown>"))
+      _pgmout_fmt(pinname, "%s", str);
   }
 
   pgmstr = dev_hvupdi_support_liststr(pgm);
@@ -1413,9 +1414,54 @@ static void dev_pgm_strct(const PROGRAMMER *pgm, bool tsv, const PROGRAMMER *bas
 }
 
 
-// -c <wildcard>/[dASsrtiBUPTIJWHQ]
+typedef struct {
+  int vid, pid, ishid;
+  const char *ids;
+} Dev_udev;
+
+
+static Dev_udev *add_udev(Dev_udev *ud, int *uip, int vid, int pid, int ishid, const char *ids) {
+  for(int i = 0; i < *uip; i++)  // Already entered?
+    if(ud[i].vid == vid && ud[i].pid == pid && ud[i].ishid == ishid && ud[i].ids == ids)
+      return ud;
+  if(*uip % 128 == 0)
+    ud = (Dev_udev *) mmt_realloc(ud, sizeof*ud*(*uip+128));
+
+  ud[*uip].vid = vid;
+  ud[*uip].pid = pid;
+  ud[*uip].ishid = ishid;
+  ud[*uip].ids = ids;
+  (*uip)++;
+
+  return ud;
+}
+
+static int udev_cmp_wout_ids(const Dev_udev *p1, const Dev_udev *p2) {
+  int diff;
+  if((diff = p1->vid - p2->vid))
+    return diff;
+  if((diff = p1->pid - p2->pid))
+    return diff;
+  return p1->ishid - p2->ishid;
+}
+
+static int udev_cmp(const void *v1, const void *v2) {
+  const Dev_udev *p1 = v1, *p2 = v2;
+  int diff;
+
+  if((diff = udev_cmp_wout_ids(p1, p2)))
+    return diff;
+  return strcmp(p1->ids, p2->ids);
+}
+
+#include "flip1.h"
+#include "flip2.h"
+#include "jtag3.h"
+#include "stk500v2.h"
+
+// -c <wildcard>/[duASsrtiBUPTIJWHQ]
 void dev_output_pgm_defs(char *pgmidcp) {
-  bool descs, astrc, strct, cmpst, raw, tsv, injct;
+  bool descs, astrc, strct, cmpst, raw, tsv, injct, udev;
   char *flags;
   int nprinted;
   PROGRAMMER *nullpgm = pgm_new();
@@ -1426,7 +1472,7 @@ void dev_output_pgm_defs(char *pgmidcp) {
   if(!flags && str_eq(pgmidcp, "*")) // Treat -c * as if it was -c */s
     flags = "s";
 
-  if(!*flags || !strchr("dASsrtiBUPTIJWHQ", *flags)) {
+  if(!*flags || !strchr("duASsrtiBUPTIJWHQ", *flags)) {
     dev_info("Error: flags for developer option -c <wildcard>/<flags> not recognised\n");
     dev_info(
       "Wildcard examples (these need protecting in the shell through quoting):\n"
@@ -1436,6 +1482,7 @@ void dev_output_pgm_defs(char *pgmidcp) {
       "  jtag?pdi matches jtag2pdi and jtag3pdi\n"
       "Flags (one or more of the characters below):\n"
       "         d  description of core programmer features\n"
+      "         u  show udev entry for programmer\n"
       "         A  show entries of avrdude.conf programmers with all values\n"
       "         S  show entries of avrdude.conf programmers with necessary values\n"
       "         s  show short entries of avrdude.conf programmers using parent\n"
@@ -1465,8 +1512,12 @@ void dev_output_pgm_defs(char *pgmidcp) {
   raw   = !!strchr(flags, 'r');
   tsv   = !!strchr(flags, 't');
   injct = !!strchr(flags, 'i');
+  udev  = !!strchr(flags, 'u');
 
   nprinted = dev_nprinted;
+
+  int ui = 0;
+  Dev_udev *udr = NULL;
 
   LNODEID ln1, ln2;
   for(ln1=lfirst(programmers); ln1; ln1=lnext(ln1)) {
@@ -1480,6 +1531,7 @@ void dev_output_pgm_defs(char *pgmidcp) {
     }
     if(!matched)
       continue;
+
     if(!prog_modes_in_flags(pgm->prog_modes, flags))
       continue;
 
@@ -1508,7 +1560,60 @@ void dev_output_pgm_defs(char *pgmidcp) {
         );
       }
 
+    if(udev && pgm->usbpid && (pgm->conntype == CONNTYPE_USB || is_serialadapter(pgm))) {
+      void (* pi)(PROGRAMMER *) = pgm->initpgm;
+      const char *ids = cache_string(str_ccpgmids(pgm->id));
+      int usbvid = pgm->usbvid, ishid =
+        pi == jtag3_initpgm || pi == jtag3_pdi_initpgm || pi == jtag3_updi_initpgm ||
+        pi == jtag3_dw_initpgm || pi == stk500v2_jtag3_initpgm || pi == jtag3_tpi_initpgm;
+
+      if(!lfirst(pgm->usbpid)) {
+        if(pi == flip1_initpgm || pi == flip2_initpgm) { // Bootloaders, add possible part pids
+          for(LNODEID lp = lfirst(part_list); lp; lp = lnext(lp)) {
+            AVRPART *pt = ldata(lp);
+            if(pt->usbpid)
+              udr = add_udev(udr, &ui, usbvid, pt->usbpid, 0, ids);
+          }
+        }
+      }
+
+      for(LNODEID pidn=lfirst(pgm->usbpid); pidn; pidn=lnext(pidn))
+        udr = add_udev(udr, &ui, usbvid, *(int *) ldata(pidn), ishid, ids);
+    }
+
     if(raw)
       dev_pgm_raw(pgm);
+  }
+
+  if(udev && ui) {
+    dev_info("# 1. Put Linux udev rules into /etc/udev/rules.d/80-avrdude.rules (or similar)\n");
+    dev_info("# 2. sudo udevadm control --reload-rules && sudo udevadm trigger (or similar)\n");
+    dev_info("# 3. Unplug the device and plug it in again\n");
+    qsort(udr, ui, sizeof *udr, udev_cmp);
+    char *prev_head = mmt_strdup("<none>");
+    for(Dev_udev *u = udr; u-udr < ui; u++) {
+      char head[1024] = {0}, *h = head;
+      strncpy(h, u->ids, sizeof head - 1), h += strlen(h);
+      for(Dev_udev *v = u+1; v-udr < ui; v++) {
+        if(udev_cmp_wout_ids(u, v))
+          break;
+        if((int) (strlen(v->ids) + 3 + h-head) <= (int) sizeof head) {
+          strcpy(h, ", "), h += 2;
+          strcpy(h, v->ids), h += strlen(v->ids);
+        }
+        u = v;
+      }
+      if(!str_eq(prev_head, head)) {
+        dev_info("\n# %s\n", head);
+        mmt_free(prev_head);
+        prev_head = mmt_strdup(head);
+      }
+      dev_info("SUBSYSTEM==\"usb\", ATTRS{idVendor}==\"%04x\", ATTRS{idProduct}==\"%04x\", "
+        "GROUP=\"plugdev\", MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
+      if(u->ishid)
+        dev_info("KERNEL==\"hidraw*\", SUBSYSTEM==\"hidraw\", ATTRS{idVendor}==\"%04x\", "
+        "ATTRS{idProduct}==\"%04x\", GROUP=\"plugdev\", MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
+    }
+    mmt_free(prev_head);
   }
 }
--- a/src/developer_opts.c
+++ b/src/developer_opts.c
@@ -1586,7 +1586,7 @@ void dev_output_pgm_defs(char *pgmidcp) {
   }
 
   if(udev && ui) {
-    dev_info("# 1. Put Linux udev rules into /etc/udev/rules.d/80-avrdude.rules (or similar)\n");
+    dev_info("# 1. Put Linux udev rules into /etc/udev/rules.d/55-avrdude.rules (or similar)\n");
     dev_info("# 2. sudo udevadm control --reload-rules && sudo udevadm trigger (or similar)\n");
     dev_info("# 3. Unplug the device and plug it in again\n");
     qsort(udr, ui, sizeof *udr, udev_cmp);
@@ -1609,10 +1609,10 @@ void dev_output_pgm_defs(char *pgmidcp) {
         prev_head = mmt_strdup(head);
       }
       dev_info("SUBSYSTEM==\"usb\", ATTRS{idVendor}==\"%04x\", ATTRS{idProduct}==\"%04x\", "
-        "GROUP=\"plugdev\", MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
+        "MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
       if(u->ishid)
         dev_info("KERNEL==\"hidraw*\", SUBSYSTEM==\"hidraw\", ATTRS{idVendor}==\"%04x\", "
-        "ATTRS{idProduct}==\"%04x\", GROUP=\"plugdev\", MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
+          "ATTRS{idProduct}==\"%04x\", MODE=\"0660\", TAG+=\"uaccess\"\n", u->vid, u->pid);
     }
     mmt_free(prev_head);
   }
--- a/src/developer_opts.c
+++ b/src/developer_opts.c
@@ -1586,9 +1586,20 @@ void dev_output_pgm_defs(char *pgmidcp) {
   }
 
   if(udev && ui) {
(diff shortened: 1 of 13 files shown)


stefanrueger:
@ndim Please have a look at this PR that shows udev rules and see whether this might be useful. Try `avrdude -c*/u` to get *all* udev rules for USB devices known to AVRDUDE. @mcuee or @MCUdude you might want to comment on the addition of usbvid and usbpid for USB programmers in avrdude.conf.

ndim:
A few general remarks without actually running the code (I am in the middle of a merge conflict). I will take a closer look later.

There are other supported operating systems which could make use of the same information in a different format in a different place. Option naming etc. should not rule out adding support for other output formats later. `avrdude` might just print the list in whatever format the operating system `avrdude` is being built for uses. As long as there are never two different formats, that would work.

  * IIRC FreeBSD has a text config file in a different format which can do that job (I have forgotten any details). Other BSDs might be similar to FreeBSD.

  * Windows and OSX are certainly doing their very own thing about which I have absolutely no idea.

Writing such rules does not necessarily need to happen inside the `avrdude` executable. A helper executable might do just as fine, and that would reduce the clutter in the avrdude options. On the other hand, avrdude already contains a lot of the necessary code, so adding just that functionality is less work initially, and probably to maintain as well. And it appears as if the functionality is integrating quite nicely into the `avrdude` programmer listing code.

ndim:
I have written down a few things about using udevadm somewhere IIRC for https://github.com/socratools ...

  * IIRC the reload happens automatically, but should be made explicit for programmatic use
  * IIRC just re-triggering ALL devices can cause some annoyances or even problems. The alternative is to unplug/replug the device in question, or tell `udevadm` to only re-trigger the events for a positive list of specific devices using a list usb vid/pid tuples. That list is easy to make for the three devices socratools supports, but not for avrdude's dozens and dozens and dozens of programmers.

When I have found that information, a few hints could be added to the comments about using `udevadm trigger`.

src/developer_opts.c
```diff
@@ -1508,7 +1560,60 @@ void dev_output_pgm_defs(char *pgmidcp) {
...
+    }
+
     if(raw)
       dev_pgm_raw(pgm);
   }
+
+  if(udev && ui) {
+    dev_info("# 1. Put Linux udev rules into /etc/udev/rules.d/80-avrdude.rules (or similar)\n");
+    dev_info("# 2. sudo udevadm control --reload-rules && sudo udevadm trigger (or similar)\n");
+    dev_info("# 3. Unplug the device and plug it in again\n");
```
ndim: I have written down a few things about using udevadm after changing the set of udev fules. Somewhere IIRC for https://github.com/socratools ...

  * IIRC the reload happens automatically, but should be made explicit for programmatic use to avoid race conditions.

  * IIRC just re-triggering ALL devices can cause some annoyances or even problems due at least to the volume of events.

    The alternative is to unplug/replug the device in question, or tell `udevadm` to only re-trigger the events for a positive list of specific devices using a list usb vid/pid tuples.

    That list is easy to make for the three devices socratools supports, but not for avrdude's dozens and dozens and dozens of programmers.

When I have found that information, a few hints could be added to the comments about using `udevadm trigger`. Oh, and you **either** re-trigger the udev events, or you unplug and replug the device. Doing both duplicates effort.

stefanrueger:
Thanks, @ndim for great tips! I have removed the advice about re-triggering udev rules. I am relaxed about the option, as it's part of the set of *developer* options, ie, not really needed by users, but aimed as people like yourself who might use it, eg, for packaging AVRDUDE for Fedora. Developer options (`-p part/x` and `-c prog/x`) can change and don't need to be documented (though I have documented *this* one). And I'd expect developers to be able to morph this particular set of Linux udev rules into what's needed on their system...

mcuee:
Usually this should not belong to avrdude application but in a seperate contrib directory. After all, avrdude is a cross platfrom application and udev is a Linux only thingy.

Example from OpenOCD:
https://github.com/openocd-org/openocd/tree/master/contrib

mcuee:
@dl8dtl 

Please help to share your comments. Thanks.

mcuee:
> IIRC FreeBSD has a text config file in a different format which can do that job (I have forgotten any details). Other BSDs might be similar to FreeBSD.

Yes there is a way to set up USB permission under FreeBSD. OpenBSD/NetBSD have their methods as well.

> Windows and OSX are certainly doing their very own thing about which I have absolutely no idea.

For Windows you do not really need to do anything as long as the driver is installed (Driver installation requires the admin privilige).

For macOS, basically for non-HID USB devices, there should be no kernel driver attached for avrdude to work (using libusb). For HID device, then avrdude needs to use HIDAPI since libusb will not work as there is no easy way to detach the kernel HID driver.

Linux is special that it can detach the kernel driver so that either HIDAPI or libusb will work with HID device.

But libusb project recommends using HIDAPI and not libusb to talk to HID device.

dl8dtl:
> Usually this should not belong to avrdude application but in a seperate contrib directory.

Agreed. It's merely a documentation thing to have it. Providing directly applicable snippets is certainly a good idea though, the usage instructions could be placed in comments on top.

mcuee:
> > > IIRC FreeBSD has a text config file in a different format which can do that job (I have forgotten any details). Other BSDs might be similar to FreeBSD.
> > 
> > 
> > Yes there is a way to set up USB permission under FreeBSD. OpenBSD/NetBSD have their methods as well.
> 
> So eventually we should add support for generating the respective configuration data formats at the same place where we print udev rules.

I do not like this PR myself. This should be a documentation and not part of the CLI print output.

> > > Windows and OSX are certainly doing their very own thing about which I have absolutely no idea.
> > 
> > For Windows you do not really need to do anything as long as the driver is installed (Driver installation requires the admin privilige).
> 
> And that driver has access to all USB devices, and avrdude having access to the driver then has access to all USB devices?

No, the driver only have access to the specific device (or devices) which the driver is installed for. There is no such thing as usbfs under Windows. usbfs is kind of the universal driver for Linux (once the kernel driver is detached), no such thing for Windows.

> > For macOS, basically for non-HID USB devices, there should be no kernel driver attached for avrdude to work (using libusb). For HID device, then avrdude needs to use HIDAPI since libusb will not work as there is no easy way to detach the kernel HID driver.
> 
> And avrdude, just as any regular user process, has write access to all devices by default?

Again, only for device without the kernel driver, you can consider IOKit as the driver for these devices. Most of the devices will have a kernel device driver attached and libusb can not access them under macOS. libusb can not access HID device under macOS because of this reason. HIDAPI will work with generic HID device under macOS (not only USB HID but also Bluetooth).

> > Linux is special that it can detach the kernel driver so that either HIDAPI or libusb will work with HID device.
> > But libusb project recommends using HIDAPI and not libusb to talk to HID device.
> 
> And HIDAPI does not require any privileges? Every userspace process can just write to every HID device?

HIDAPI and libusb both need the privileges under Linux, udev rules are the way to go. There were history thingy which we can forget now.

Linux is the only OS which you can easily detach the existing kernel driver (with the proper prividlges) programmatically.

ndim:
> > Usually this should not belong to avrdude application but in a seperate contrib directory.
> 
> Agreed. It's merely a documentation thing to have it. Providing directly applicable snippets is certainly a good idea though, the usage instructions could be placed in comments on top.

Are you talking about the code which iterates through the list of all programmers from `avrdude.conf` and then generating a udev rules file, or about the generated rules file itself belonging to a `/contrib/` directory?

> Suggested installation location: `${PREFIX}/share/doc/os-support/`

You probably mean `${PREFIX}/share/doc/avrdude/os-support/`?

Anyway, you should not install binary executables into `${PREFIX}/share`, even if the udev rules file which that executable produces using libavrdude to iterate through `avrdude.conf` is considered documentation.

For a finished udev rules file as well, I disagree with installing it into `${PREFIX}/share`.

You need that udev rules file either while installing avrdude from sources by hand. If you are using a ready made avrdude package, installation of that package should already have installed the accompanying udev rules file into the proper location. In neither case would such a udev rules file be required as installed documentation. Maybe as part of the documentation which explains installing avrdude.

It could be useful to have the C sources for a separate such program in a `/src/contrib/` directory and have that program built when building avrdude. As this PR stands, you invoke the built `avrdude` executable to produce the udev rules, but that would just as well work with a separate executable built alongside `avrdude`.

That requires duplicating a bit of code from avrdude for parsing CLI, initializing libavrdude and load avrdude.conf and iterate through the list of programmers. The duplicated parts should be simple enough, though.

Then whoever installs that built program can run that program and install the resulting rules file where they need. The common cases would be:

  * Somebody compiling from source and installing by hand: Run `src/contrib/generate-device-list linux.udev > 80-avrdude-programmers.rules` and then `sudo cp 80-avrdude-programmers.rules /etc/udev/rules.d/`.

  * Somebody building an avrdude distro package would have that package build generate the rules file and install it into `/usr/lib/udev/rules.d/`.

But there is another case: Somebody who writes a new programmer definition into their local `avrdude.rc` or `avrdude.conf`. That could be someone who does not even have the avrdude sources, let alone the compiler etc needed to build it. It would be helpful for that person had a direct and reproducible way to generate the corresponding udev rules. And that requires that either the installed `avrdude` executable can list the supported programmers in udev rules format, or that a companion executable is installed alongside `avrdude`.

Note: When I write about "udev rules", I also think other output formats for other systems.

stefanrueger:
What's the problem you are trying to solve @dl8dtl @mcuee? This is a *developer* option that gives potentially useful advice.

Here a usage scenario: An educator wants to use an RPi (great) under Linux (as you do) as host PC to program AVRs (fantastic) with AVRDUDE (brilliant); they bought an xplained mini board  (not bad), plug it in and AVRDUDE doesn't work. What's the problem with AVRDUDE saying Hang on a sec, this might be udev rules; check out avrdude -c xplainedmini/u?

That would be a million times better than what I went through when I couldn't get that to run as user for my life. Even asking the AVRDUDE community drew a blank (although I suspect that if I had understood @ndim's reply better and explored the Fedora rules more this might have worked, too). I *do* have reasonable Linux experience, patience, persistence and an ability to read documentation. Others must have a similar problems, as I've seen `sudo avrdude ...` command lines in past issues.

I'd *really* like AVRDUDE to help out. It's no good having a piece of text/executable lurking around in a dusty corner of the OS. How would the user know it's even there? It's AVRDUDE who recognises the problem, knows about a potential solution and ... keeps shtum?! Because ...?

And, as @ndim said, this PR or any future PR could help Windows users, MacOs, FreeBSD etc. Maybe it's not a problem there (because they do better than my Linux distro in making plug and play devices playable).

This one is particularly fiendish in Linux: For non-HID USB there are quite a few mildly complex instructions on the web (typically 10 steps one of which is how to find out the USB VID/PID, which of course AVRDUDE already knows). But for HID that's been absurdly difficult. Here a few things that make it hard: Although udev is documented  in general, it is still complex. Is it `ATTR` or is it `ATTRS`? What's the correct USB HID device SUBSYSTEM? You actually need to already understand a lot to be able to ask these questions. Linux distros are at war how to deal with the underlying problem: one fraction thinks there should be user groups that are allowed to plug and play; the other fractions either think that's absurd, a bug or not *pure* (or any combination of that) and offer an [undocumented `systemd` feature `TAG+="uaccess"`](https://github.com/systemd/systemd/issues/4288). All I could dig up was some speculation in Archlinux documentation how that works. And apparently that only works if the udev rules [file name starts with a number under 73](https://github.com/systemd/systemd/issues/4288#issuecomment-348166161) as (my take) existing rule 73 implements the `"uaccess"` tag. It seems like no one asks the question *How would a user know?*.  And don't get me started on the `KERNEL="hidraw*"` business. You really need a black belt in udev tools to figure out how to write an udev rule for *that*.

Anyway. What *is* the problem you are trying to solve @dl8dtl @mcuee?

ndim:
I have to withdraw my above opinion that a buildtime-only executable would be a good place to put the udev rules generating code. The udev rules generation code should be part of the software installed as part of installing the libavrdude library as part of the avrdude software package, whether the executable printing udev rules is called `avrdude` or something else.

I was not thinking about this properly, as I had some implicit assumptions from libgphoto2 (which I have been involved in for >20 years). In libgphoto2, the list of supported devices is maintained in the form of C structs, with each camera driver containing the list of cameras it supports. So the list of supported cameras can only change by rebuilding libgphoto2, and that means that udev rule generation can also happen at libgphoto2 build time.

However, libavrdude maintains the device list in `avrdude.conf` and actively encourages users to either edit `avrdude.conf` (which then overrides the `avrdude.conf` shipped with libavrdude) or to write a `avrdude.rc` file (which adds to the `avrdude.conf` shipped with libavrdude). That is clearly a much more flexible and user friendly way to define the list of devices, especially when dealing with a market like programmers where models can be very numerous and still work similarly or the same as others.

So users adding devices to be handled by libavrdude means that those users need proper access permissions to those devices. And on Linux these days, that means udev rules.

mcuee:
@stefanrueger 

> You really need a black belt in udev tools to figure out how to write an udev rule for that.

I understand. Been there. I was involved in libusb and libusb-win32 project since 2005 and I have gone through the changes in Linux with regard to USB device permission.

I wrote the following Wiki page 15 years ago mainly for my own reference at that time.
https://github.com/mcuee/picusb/blob/master/wiki/libusb_and_udev.wiki

Legacy Linux info for reference. udev rule --> HAL --> console kit / policy kit --> udev rules
http://www.gphoto.org/doc/manual/permissions-usb.html
 
> Anyway. What is the problem you are trying to solve @dl8dtl @mcuee?

I am more in favor of providing a reference udev rules like what OpenOCD did. 
https://github.com/openocd-org/openocd/blob/master/contrib/60-openocd.rules

The user can create the rules based on the reference example. They usually have only a few programmers and do not need a complex rules. My worry is also that the reference rule may not work for some Linux distros. For those special Linux distros they can ask in their communities for help.

But I understand what you are trying to do here -- to automatically generate the udev rules. 

In the end, you can go with this PR.

mcuee:
> I was not thinking about this properly, as I had some implicit assumptions from libgphoto2 (which I have been involved in for >20 years). In libgphoto2, the list of supported devices is maintained in the form of C structs, with each camera driver containing the list of cameras it supports. So the list of supported cameras can only change by rebuilding libgphoto2, and that means that udev rule generation can also happen at libgphoto2 build time.

It is interesting that the following file still mentioned the outdated HAL thingy.
https://github.com/gphoto/libgphoto2/blob/master/README.packaging

The following document from gPhoto used to be my reference, but it is too outdated now.
http://www.gphoto.org/doc/manual/permissions-usb.html

mcuee:
For FreeBSD, devd is the way to go.

Reference:
https://github.com/freebsd/freebsd-ports/blob/main/graphics/sane-backends/files/pkg-message.in

mcuee:
NetBSD reference:
https://wiki.netbsd.org/scanner/

dl8dtl:
> > Suggested installation location: `${PREFIX}/share/doc/os-support/`
> 
> You probably mean `${PREFIX}/share/doc/avrdude/os-support/`?

Sure.

> Anyway, you should not install binary executables into `${PREFIX}/share`, even if the udev rules file which that executable produces using libavrdude to iterate through `avrdude.conf` is considered documentation.
> 

I do not see any "binary executable" there.

> For a finished udev rules file as well, I disagree with installing it into `${PREFIX}/share`.

It seems we have vastly differing opinions on that.

To me, udev (or devd, on *BSD) rules are a documentation thing. We provide them as a courtesy to our users. It's up to them what they make out of that. As AVRDUDE maintainer, I would never dare to actually install these files into any "live" place.

Of course, package maintainers might opt to either use them as a template, or even automatically install them into their "live" location, if they think they trust us enough. That's where these files are leaving the documentation domain, and get into the system domain. But it's IMHO not our turn to decide about that.

ndim:
> > > Suggested installation location: `${PREFIX}/share/doc/os-support/`
> > 
> > 
> > You probably mean `${PREFIX}/share/doc/avrdude/os-support/`?
> 
> Sure.
> 
> > Anyway, you should not install binary executables into `${PREFIX}/share`, even if the udev rules file which that executable produces using libavrdude to iterate through `avrdude.conf` is considered documentation.
> 
> I do not see any "binary executable" there.

I thought you were talking about the executable which converts the information from `avrdude.conf` into a udev rules file, as this PR does not generate a udev rules file yet.

As to a generated udev rules file...

> > For a finished udev rules file as well, I disagree with installing it into `${PREFIX}/share`.
> 
> It seems we have vastly differing opinions on that.
> 
> To me, udev (or devd, on *BSD) rules are a documentation thing. We provide them as a courtesy to our users. It's up to them what they make out of that. As AVRDUDE maintainer, I would never dare to actually install these files into any "live" place.
> 
> Of course, package maintainers might opt to either use them as a template, or even automatically install them into their "live" location, if they think they trust us enough. That's where these files are leaving the documentation domain, and get into the system domain. But it's IMHO not our turn to decide about that.

...I agree with you. Had my developer and packager hats confused here, sorry.

Upstream avrdude (us here) should provide the code which converts the information from `avrdude.conf` into a udev rules file, and possibly create such a udev rules file during build and install it as documentation to give people installing avrdude from source some pointers about using those udev rules.

Downstream distro packagers (me over at Fedora and others at other distros) then can install that udev rules file or a derivative however they deem adequate.

mcuee:
> To me, udev (or devd, on *BSD) rules are a documentation thing. We provide them as a courtesy to our users. It's up to them what they make out of that. As AVRDUDE maintainer, I would never dare to actually install these files into any "live" place.
> 
> Of course, package maintainers might opt to either use them as a template, or even automatically install them into their "live" location, if they think they trust us enough. That's where these files are leaving the documentation domain, and get into the system domain. But it's IMHO not our turn to decide about that.

Totally agree with you.

Coming back to this PR, as it does not really install the udev rule files to any "live" places, are you okay with the PR? It just shows the example udev rules.

mcuee:
> Upstream avrdude (us here) should provide the code which converts the information from `avrdude.conf` into a udev rules file, and possibly create such a udev rules file during build and install it as documentation to give people installing avrdude from source some pointers about using those udev rules.

This will be indeed good. However this PR does not really generate the file or install it as a documentation example. If you want, maybe you can create a PR to address this after this PR is merged.

> Downstream distro packagers (me over at Fedora and others at other distros) then can install that udev rules file or a derivative however they deem adequate.

Yes, this should be the way.

stefanrueger:
And here the user experience in Linux of a *novice* user having bought the `usbasp` programmer:
```
$ avrdude -c usbasp -p t13a

Warning: cannot open USB device: Permission denied
Error: cannot find USB device with vid=0x16c0 pid=0x5dc vendor='www.fischl.de' product='USBasp'
Error: unable to open port usb for programmer usbasp

USB access errors detected; it is not good practice to run avrdude as root;
consider checking whether udev rules are set correctly: $ avrdude -c usbasp/u

Avrdude done.  Thank you.
```
Turns out AVRDUDE is in the best position to give live advice. Similar things could be implemented for FreeBSD, Windows, but I for one don't know what advice to give and how to check that the advice is correct. So that's for  someone else, but the infrastructure is there. OK, what does the Linux advice say?

```
$ avrdude -c usbasp/u

1. Examine the suggested udev rule below; to install run:

avrdude -c "usbasp/u" | tail -n +11 | sudo tee /etc/udev/rules.d/55-avrdude-usbasp.rules
sudo chmod 0644 /etc/udev/rules.d/55-avrdude-usbasp.rules

2. Unplug any AVRDUDE USB programmers and plug them in again
3. Enjoy user access to the USB programmer(s)

Note: To install all udev rules known to AVRDUDE follow: avrdude -c "*/u" | more

# Generated from avrdude -c "usbasp/u"

# usbasp
SUBSYSTEM=="usb", ATTRS{idVendor}=="16c0", ATTRS{idProduct}=="05dc", MODE="0660", TAG+="uaccess"
```

Step 1:

```
avrdude -c "usbasp/u" | tail -n +11 | sudo tee /etc/udev/rules.d/55-avrdude-usbasp.rules
# Generated from avrdude -c "usbasp/u"
# usbasp
SUBSYSTEM=="usb", ATTRS{idVendor}=="16c0", ATTRS{idProduct}=="05dc", MODE="0660", TAG+="uaccess"
```

Step 2: Unplug the programmer and plug it in again.

Then the all-important Step 3:

```
$ (udev 6U) avrdude -c usbasp -p t13a -t

avrdude> 
```

Voila.

And `avrdude -c "*/u" | more` is for @ndim

dl8dtl:
> Upstream avrdude (us here) should provide the code which converts the information from `avrdude.conf` into a udev rules file

I'm not even sure it is necessary to create this out of `avrdude.conf`. The information in the udev etc. rules is changing so slowly, we could easily maintain the respective rule files by hand.

> are you okay with the PR?

Personally, I don't think it's the job of the avrdude binary to provide that information. I'd favor a PR instead that creates some rule file templates as part of the (installed) documentation.

stefanrueger:
> don't think it's the job of the avrdude binary to provide that information

Sure not its job, but it's the best place for providing that info: the binary just experienced a permission error; it knows the usbvid/pid is attached, etc. Mind you, that usbvid/pid might only exist in the user's personal .conf file b/c they bought a programmer that had a different vendor id than the known ones in avrdude.conf.

>  I'd favor a PR instead that creates some rule file templates as part of the (installed) documentation.

You are welcome to do *that* for FreeBSD. I have started a section A.1.4.1 FreeBSD USB Permissions that you are welcome to extend. :wink:

stefanrueger:
Planning to merge soon...

PR #1863, opened 2024-08-02
PICkit 5 support

MX682X:
Alright, should be complete now, or at least complete enough for experimental release.
Supports: Writing and reading of all memories, chip erase, Setting Baudrate (which is the UPDI Baudrate in this case), Supplying Power through "-xvoltage=<arg>"[mV]

As the PICkit5 can read and write arbitrary amounts, I've added read_array()/write_array() functions, allowing a significant reduction in overhead, compared to single byte reads/writes of e.g. SRAM

When testing the terminal mode, I noticed that some functions were checking "rc" against non zero after byte reads, which didn't make sense for me, as the read functions are allowed to return number of read bytes (meaning returning 1 for 1 byte read was treated as error), so I decided to change it to check against negative rc

Oh, and one more thing:
I haven't programmed much for OS based systems, so I can't tell if it's bad or not: There are some functions, where I declare a small local array and pass it as pointer to another function. So far, no problem my test, but I can imagine it might cause Problems. Any suggestions on that topic?

PR #1879, opened 2024-08-11
Make USB product string available

MCUdude:
I figured it would be nice to have the actual USB string available so we could print it using `pgm->display`. Especially now that PR #1863 supports multiple different programmers using `-c pickit5`.

@stefanrueger I know we generally shouldn't touch the existing code formatting, but the usb_libusb.c formatting was _completely broken_, to a point where no text editor tabs/spaces setting could make the code acceptable. Indents were everywhere, and things were just a mess. I took the liberty to re-format the file as a separate commit.

### DIFF
--- a/src/jtag3.c
+++ b/src/jtag3.c
@@ -1760,9 +1760,13 @@ int jtag3_open_common(PROGRAMMER *pgm, const char *port, int mode_switch) {
     pmsg_notice2("found CMSIS-DAP compliant device, using EDBG protocol\n");
   }
 
-  // Make USB serial number available to programmer
-  if (serdev && serdev->usbsn)
-    pgm->usbsn = serdev->usbsn;
+  // Make USB serial number and USB product name available to programmer
+  if (serdev) {
+    if (serdev->usbsn)
+      pgm->usbsn = serdev->usbsn;
+    if (serdev->usbproduct)
+      pgm->usbproduct = serdev->usbproduct;
+  }
 
   // Drain any extraneous input
   jtag3_drain(pgm, 0);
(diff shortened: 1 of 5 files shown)


MCUdude:
Damnit! That merge conflict is there because I somehow forgot to update my dev branch before creating this PR. I'll open a new PR when I've figured it out...

dl8dtl:
> but the usb_libusb.c formatting was _completely broken_

It was using hard TABs each 8 columns (yet 2-column indentation levels) – something that once used to be *the* standard, back when terminals defaulted to 8-column TABs.

Here's how my Emacs displays it (with the hard TABs dimmed a bit, to show where they are).
screenshot-2024-08-11T231729

stefanrueger:
Yes, the trouble is no-one knows (or indeed can know) which file expects tab to be expanded to how many spaces. 8 used to be the default, but then people told their editors to use 4 spaces (or even 2).

My suggestion is to have a single PR before the release that re-formats the source to have two-space indentation. And no tabs. 

There are disadvantages to that. Weighing up the advantages my scales tip in favour of re-formatting.

MCUdude:
> It was using hard TABs each 8 columns

8 "spaces" per tab made the code readable. However, there are lots of spaces there as well. Why place the curly braces on the next line _with_ two additional spaces as well?

``` c
  for (bus = usb_get_busses(); bus; bus = bus->next)
    {
      for (dev = bus->devices; dev; dev = dev->next)
	{
	  if (dev->descriptor.idVendor == pinfo.usbinfo.vid &&
	      dev->descriptor.idProduct == pinfo.usbinfo.pid)
	    {
	      udev = usb_open(dev);
	      if (udev)
		{
```

> My suggestion is to have a single PR before the release that re-formats the source to have two-space indentation. And no tabs.

I'm all for this! A file with _just_ tabs is OK, I think. It's when they are mixed it becomes a problem.

PR #1258, opened 2023-01-02
Move static usbsn into serdev structure, remove serno access function

stefanrueger:
See [here](https://github.com/avrdudes/avrdude/issues/1254#issuecomment-1369091620) for details.

PR #1880, opened 2024-08-11
Make usbproduct available to programmers

MCUdude:
I messed up PR #1879, so I'm trying again.

I figured it would be nice to have the actual USB string available so we could print it using `pgm->display`. Especially now that PR #1863 supports multiple different programmers using `-c pickit5`.

@stefanrueger I know we generally shouldn't touch the existing code formatting, ~~but the usb_libusb.c formatting was _completely broken_, to a point where no text editor tabs/spaces setting could make the code acceptable. Indents were everywhere, and things were just a mess. I took the liberty to re-format the file as a separate commit.~~

I haven't done anything with the formatting yet, but I'm itching to do so! I can't believe how it could end up this bad. Intents are all over the place regardless of text editor settings.

### DIFF
--- a/src/libavrdude.h
+++ b/src/libavrdude.h
@@ -889,6 +889,7 @@ struct serial_device {
   int (*set_dtr_rts)(const union filedescriptor *fd, int is_on);
 
   const char *usbsn;
+  const char *usbproduct;
   int flags;
 #define SERDEV_FL_NONE         0x0000 /* no flags */
 #define SERDEV_FL_CANSETSPEED  0x0001 /* device can change speed */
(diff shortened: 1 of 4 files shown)


stefanrueger:
Code looks good. Have you tried and can evidence it works as intended?

MCUdude:
> Code looks good. Have you tried and can evidence it works as intended?

Yes, tried with both hidapi and libusb

mcuee:
@MCUdude 

Similar changes can be done for `jtag3.c` as well.

Reference:
* https://github.com/avrdudes/avrdude/pull/1258/files

mcuee:
BTW, other programmers may probably be added in the future as well. But let's wait after avrdude 8.0 release.

One potential example: USBASP.

PR #1895, opened 2024-08-22
Add detail to part and programmer list in avrdude.pdf

stefanrueger:
The -c \? and -p \? lists in the documentation have been moved to appendices and beefed up with programming modes and, for -p, with alternative names. Parts and programmers are now automatically put into the index.


stefanrueger:
@ndim I am guessing (but do not know) that the autotools run fails b/c the files `programmers.sed` and `parts.sed` are only in this PR but not (yet) in the repository itself (where they would be after merging). So reckon the failing check is a red herring. But what do I know?

ndim:
> @ndim I am guessing (but do not know) that the autotools run fails b/c the files `programmers.sed` and `parts.sed` are only in this PR but not (yet) in the repository itself (where they would be after merging). So reckon the failing check is a red herring. But what do I know?

See my review comments. Probably. I am on a mobile phone right now, cannot really check.

stefanrueger:
@ndim Thanks, I am having a hard time to find your review comments. No worries if they were lost somehow.

ndim:
> @ndim Thanks, I am having a hard time to find your review comments. No worries if they were lost somehow.

The make recipes need to refer to `$(srcdir)/parts.sed` and `$(srcdir)/programmers.sed`. Both in the dependency declaration and in the body of the recipe.

PR #1902, opened 2024-08-22
Document: use macOS to replace MacOS, no support of Parallel Port under macOS

mcuee:


mcuee:
@stefanrueger 

Please review. I think this PR should be safe to merge.

PR #1906, opened 2024-08-23
Remove known limitations comment from serprog.c

funkeleinhorn:
This PR removes obsolete comments from the serprog.c file header.

Connecting over tcp is already possible by specifying net:<host>:<port> as a port.
Performance is already good and further improvements would require custom AVR read/write logic to batch operations which seems not worth the maintenance effort for now.


stefanrueger:
Thanks for the PR! Looks good.

PR #1910, opened 2024-08-24
Add checks for zombie sourcecode problems

ndim:
This adds checks for zombie sourcecode problems, i.e. problems we have tried to eliminate but which might be or have been resurrected.

The example check is for the problematic

    #include "ac_cfg.h"

with double quotes instead of <> which were identified as a problem in https://github.com/avrdudes/avrdude/issues/1706 and then fixed.

### DIFF
--- a/.github/workflows/check-sourcecode.yml
+++ b/.github/workflows/check-sourcecode.yml
@@ -1,0 +1,24 @@
+name: "Check Sourcecode"
+
+on:
+  push:
+    branches-ignore:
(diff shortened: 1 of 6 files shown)


stefanrueger:
Great catch. We probably overlooked the `.l`, `.y` at the time of the original PR, and `serprog.c`/`libavrdude.i` wasn't spotted at review time (both will have worked with pre #1707 copies).

stefanrueger:
I think this is good to merge

PR #1916, opened 2024-08-29
Document elf2tag

ndim:
This aims to add documentation to the `elf2tag` script (installed to `$(bindir)` by default):

  * Add `-h` and `--help` argument handling to print the usage information.

  * Add a very basic man page `elf2tag.1`.

    Some distributions encourage having man pages (e.g. Fedora), some other distributions require a man page for each executable in `$(bindir)` (e.g. Debian).

    It makes sense to concentrate this documentation effort at the place common to all distributions, i.e. upstream, i.e. here.

TBD:

  * Build `elf2tag.1` from `elf2tag.1.adoc` or just commit `elf2tag.1` without `elf2tag.1.adoc` (like `avrdude.1`)?
  * More information for `elf2tag.1`, e.g. an example command using `avrdude disasm`?


stefanrueger:
Cool idea!

ndim:
Does this need any additions?

```
ELF2TAG(1)                     avrdude Manual                     ELF2TAG(1)

NAME
       elf2tag - output a tagfile for the avrdude disasm command

SYNOPSIS
       elf2tag <file.elf>

       elf2tag [-h | --help]

DESCRIPTION
       elf2tag generates a tagfile for use with the avrdude disasm command.

OPTIONS
       -h --help
           Prints the help message and exits.

EXAMPLES
           $ elf2tag file.elf > file.tag
           $ avrdude -t ...
           avrdude> ...
           avrdude> disasm -t=file.tag ...

AUTHORS
       elf2tag was written by Johannes Bauer with small modifications by
       Stefan Rueger.

SEE ALSO
       avrdude(1), avr-nm(1), avr-objdump(1)

avrdude                          2024-09-04                       ELF2TAG(1)
```

stefanrueger:
> Is the example command with `avrdude disasm` accurate?

@ndim Here a more extensive example using `blink.elf` compiled for ATmega328P as input:
```
$ elf2tag blink.elf  >blink.tag
$ avrdude -qq -c dryrun -p m328p -U blink.elf -t
avrdude> disasm -t=blink.tag flash 0 512
```

ndim:
I have updated the example in the elf2tag.1 man page.

Still TODO: Whether and possibly how to hook updating `src/elf2tag.1` from `src/elf2tag.1.adoc` into pre-commit, CI, builds, and whether to commit `src/elf2tag.1.adoc` at all.

``` man
ELF2TAG(1)                      avrdude Manual                     ELF2TAG(1)

NAME
       elf2tag - output a tagfile for the avrdude disasm command

SYNOPSIS
       elf2tag <file.elf>

       elf2tag [-h | --help]

DESCRIPTION
       elf2tag generates a tagfile for use with the avrdude disasm command.

OPTIONS
       -h --help
           Prints the help message and exits.

EXAMPLES
           $ elf2tag blink.elf > blink.tag
           $ avrdude -qq -c dryrun -p m328p -U blink.elf -t
           avrdude> disasm -t=file.tag flash 0 512

AUTHORS
       elf2tag was written by Johannes Bauer with small modifications by
       Stefan Rueger.

SEE ALSO
       avrdude(1), avr-nm(1), avr-objdump(1)

avrdude                           2024-11-20                       ELF2TAG(1)
```

ndim:
Still TODO: Whether and possibly how to hook updating `src/elf2tag.1` from `src/elf2tag.1.adoc` into pre-commit, CI, builds, and whether to commit `src/elf2tag.1.adoc` at all.

This is a matter of running the `src/update-elf2tag-manpage` script in a git clone (can be shallow) and then reacting if the `src/elf2tag.1` file now has changes.

ndim:
BTW, the "Build / macos-x86_64" CI fail is fixed in 8fc665fc4e5e7ec2da87060eca851c5ba956935c which is part of https://github.com/avrdudes/avrdude/pull/1934.

src/elf2tag.1
```diff
@@ -0,0 +1,62 @@
...
+elf2tag \- output a tagfile for the avrdude disasm command
+.SH "SYNOPSIS"
+.sp
+\fBelf2tag\fP <file.elf>
+.sp
+\fBelf2tag\fP [\fB\-h\fP | \fB\-\-help\fP]
+.SH "DESCRIPTION"
+.sp
+\fIelf2tag\fP generates a tagfile for use with the \fIavrdude disasm\fP command.
+.SH "OPTIONS"
```
stefanrueger, line 40: Maybe use OPTION instead of OPTIONS?

src/elf2tag.1
```diff
@@ -0,0 +1,62 @@
...
+.sp
+\fB\-h\fP \fB\-\-help\fP
+.RS 4
+Prints the help message and exits.
+.RE
+.SH "EXAMPLES"
```
stefanrueger, line 46: EXAMPLE instead of EXAMPLES?

src/elf2tag.1
```diff
@@ -0,0 +1,62 @@
...
+.sp
+.if n .RS 4
+.nf
+.fam C
+$ elf2tag blink.elf > blink.tag
+$ avrdude \-qq \-c dryrun \-p m328p \-U blink.elf \-t
+avrdude> disasm \-t=file.tag flash 0 512
```
stefanrueger, line 53: `\-t=blink.tag`

src/elf2tag.1
ndim: I followed the conventional section names in man pages according to `man-pages(7)` on my Linux system which is:

```
              NAME
              LIBRARY          [Normally only in Sections 2, 3]
              SYNOPSIS
              CONFIGURATION    [Normally only in Section 4]
              DESCRIPTION
              OPTIONS          [Normally only in Sections 1, 8]
              EXIT STATUS      [Normally only in Sections 1, 8]
              RETURN VALUE     [Normally only in Sections 2, 3]
              ERRORS           [Typically only in Sections 2, 3]
              ENVIRONMENT
              FILES
              ATTRIBUTES       [Normally only in Sections 2, 3]
              VERSIONS         [Normally only in Sections 2, 3]
              STANDARDS
              HISTORY
              NOTES
              CAVEATS
              BUGS
              EXAMPLES
              AUTHORS          [Discouraged]
              REPORTING BUGS   [Not used in man-pages]
              COPYRIGHT        [Not used in man-pages]
              SEE ALSO
```

FreeBSD man pages appear to use "The following options are available:" inside the `DESCRIPTION` section.

I would rather not invent a third way of having a differently named `OPTION` section.

src/elf2tag.1
ndim: See above on OPTIONS.

src/elf2tag.1
ndim: Fixed.

ndim:
The comment on the macos CI builds failing from still applies, but now for both the cmake *and* the automake build. Still both issues are fixed or worked around by https://github.com/avrdudes/avrdude/pull/1934/commits/b8594d1df1daae562af1f9d981f011db0a526972 from the https://github.com/avrdudes/avrdude/pull/1934 PR.

ndim:
@stefanrueger Do you have any preferences on hooking the `src/update-elf2tag-manpage` script into something? I can think of a few options to do that to help ensure that `src/elf2tag.1` and `src/elf2tag.1.adoc` are always committed in sync with each other:

  * CI rejecting pushes (looks the easiest and most comprehensive), either as part of `build.yml` or as a separate github workflow which would just clone avrdude, install `asciidoctor`, run the update script and then fail if that has changed `src/elf2tag.1`.
  * CI fixing commits on push (I would not want to implement giving CI jobs the power to change commits)
  * cmake/automake build running `src/update-elf2tag-manpage` on developer machine. Only runs on builds.
  * git pre-commit hook to run on developer machine. Depends on on developer setting that up.

ndim:
Haha. The CI check just caught me forgetting to run the update script after the quick fix changing `file.tag` to `blink.tag`.

src/elf2tag.1
stefanrueger: > I would rather not invent yet another way

Makes sense

src/elf2tag.1
stefanrueger: OK.

stefanrueger:
Cool work @ndim! I don't really have an opinion on how to deal with the workflows here (principally, b/c I have very little understanding about the ramifictions that the suggested CI/Cmake/automake changes have). So, yes, please feel free to implement a sensible solution.

ndim:
Checking this in CI adds no requirement to install asciidoctor to everybody who clones and works with the avrdude repo, just for people editing `src/elf2tag.1.adoc` which should be rare.

However, it still catches all pull requests which either

  * modify `src/elf2tag.1.adoc` without running `src/update-elf2tag-manpage` to propagate their changes to `src/elf2tag.1`
  * modify the generated `src/elf2tag.1` without editing the sources in `src/elf2tag.1.adoc`

therefore ensuring consistency between the partially redundant information in `src/elf2tag.1.adoc` and `src/elf2tag.1`.

This principle could be applied to other generated files as well.

ndim:
AFAICT, this is ready to merge.

src/elf2tag.1
ndim: And the CI has caught me forgetting to fix the generated `src/elf2tag.1` file. Fixed. This time, really.

PR #1918, opened 2024-08-31
Add AVR128DAxxS parts

stefanrueger:
Fixes #1885

@askn37 Would you like to review this PR?


askn37:
In the new DA-S parts, FUSE has increased from 9 bytes to 12 bytes. fusea==pdicfg (word size : default value 0x0003). This is the same as AVR-EB/DU.

So it seems a good idea to create a new DA-S parent-ID with the DA part as the parent, and redefine fuse to create a grandchild definition.

stefanrueger:
@askn37 Thanks for checking! I neglected to inject the memory parameters from the .atdf scraping. Now done.

I think this is ready now to be merged

askn37:
> @askn37 Thanks for checking! I neglected to inject the memory parameters from the .atdf scraping. Now done.
> 
> I think this is ready now to be merged

No problem. Thank you.

PR #1927, opened 2024-10-08
Add extra version information to avrdude output

ndim:
Some downstream projects build avrdude and want to add their own version information to the avrdude version.

This adds a clean way to do that without patching for both cmake and automake based builds:

    cmake -DEXTRA_VERSION:STRING=-arduino.1-rc1
    ../configure EXTRA_VERSION=-arduino.1-rc1

After building avrdude, the last line of "avrdude -?" will now look similar to the following, depending on whether you are building an avrdude git checkout or an avrdude release:

    avrdude version 8.0-20240901-arduino.1-rc1 (30e19f23), https://github.com/avrdudes/avrdude
    avrdude version 8.1-arduino.1-rc1 (30e19f23), https://github.com/avrdudes/avrdude

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ include(FindPackageMessage)
 include(GNUInstallDirs)
 
 set(CONFIG_DIR "${CMAKE_INSTALL_FULL_SYSCONFDIR}")
-set(AVRDUDE_FULL_VERSION ${CMAKE_PROJECT_VERSION})
+set(AVRDUDE_FULL_VERSION "${CMAKE_PROJECT_VERSION}${EXTRA_VERSION_SUFFIX}")
 
 # =====================================
 # Get Git commit info
@@ -91,7 +91,7 @@ if(Git_FOUND)
 
     # If the commit is not tagged, include the date and commit hash in the full version string.
     if(NOT GIT_COMMIT_HASH STREQUAL GIT_TAG_HASH)
-        set(AVRDUDE_FULL_VERSION "${CMAKE_PROJECT_VERSION}-${GIT_COMMIT_DATE} (${GIT_COMMIT_HASH})")
+        set(AVRDUDE_FULL_VERSION "${CMAKE_PROJECT_VERSION}-${GIT_COMMIT_DATE}${EXTRA_VERSION_SUFFIX} (${GIT_COMMIT_HASH})")
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


ndim:
This came up in https://github.com/arduino/avrdude-packing/issues/31

ndim:
`EXTRA_VERSION` might not be the best name for this. It just came to mind without me thinking much. Better ideas welcome.

mcuee:
I think `EXTRA_VERSION` is not bad. 

Maybe `CUSTOM_VERSION` is another option.

@MCUdude 

Which one do you prefer? 

Other than the name, this PR should be good to go.

umbynos (approved):

MCUdude:
@mcuee I'm fine with either. Another alternative is `VERSION_SUFFIX`, which will be `-arduino.1` in this case

mcuee:
> @mcuee I'm fine with either. Another alternative is `VERSION_SUFFIX`, which will be `-arduino.1` in this case

I like your proposal better.

@ndim 
How about you?

mcuee:
@MCUdude 

@umbynos has released Arduino flavor of avrdude 7.3 release and two RCs for avrdude 8.0. You may want to try it on your macOS machine. I will try it as well on my machines later this week.
https://github.com/arduino/avrdude-packing/releases

ndim:
Given `EXTRA_VERSION` and `VERSION_SUFFIX`, I would actually favour `EXTRA_VERSION_SUFFIX`.

That is a bit longer, but its three parts say all there is to know, making its name explain itself:

- `EXTRA` Optional extra information (not mandatory)...
- `VERSION` ...for the program version...
- `SUFFIX` ...added as a suffix.

ndim:
It should also be mentioned here that even though #1698 will always show git information like hash, branch , and tag  (if built from git checkout) in its new `--version` output, none of that information will be part of the shorter version string as shown at the end of `avrdude -?`.

So this PR will not be obsoleted by #1698.

mcuee:
> Given `EXTRA_VERSION` and `VERSION_SUFFIX`, I would actually favour `EXTRA_VERSION_SUFFIX`.
> 
> That is a bit longer, but its three parts say all there is to know, making its name explain itself:
> 
> * `EXTRA` Optional extra information (not mandatory)...
> * `VERSION` ...for the program version...
> * `SUFFIX` ...added as a suffix.

I am fine with this proposal.

ndim:
I have changed the name to `EXTRA_VERSION_SUFFIX`. I think this is ready to be merged.

PR #1934, opened 2024-11-20
CI: Ensure only supported configure options are used (--enable-option-checking=fatal)

ndim:
This ensures that only supported `configure` options are used in the CI builds, by adding `--enable-option-checking=fatal` to the `configure` command line (since Autoconf 2.62 from 2008 or something).

To make this work at all, this PR first fixes the issue of the MacOS "brew install" command suddenly failing while installing `pkg-config`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -218,10 +218,12 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v3
+      - name: Original PATH and build tools
+        run: |
+          echo "PATH=$PATH"
+          for b in bison flex pkg-config; do type "$b"; done
       - name: Install prerequisites
         run: >-
-          # brew update
-
           brew install
           cmake
           flex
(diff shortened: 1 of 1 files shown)


ndim:
Fixed commit message typo.

ndim:
Made the commit message slightly more verbose to force a CI re-run to check that the macos pkg-config issue is actually fixed for both macos builds, after the https://github.com/avrdudes/avrdude/pull/1916 PR CI runs have started to fail in *both* macos builds with the CI run for the https://github.com/avrdudes/avrdude/pull/1916/commits/5aaaf761ff30934377c5dbde704cbf81fd095140 commit instead of just the cmake macos build.

mcuee:
Looks good. I will merge this one so that CI runs will be okay.

mcuee:
@stefanrueger and @MCUdude 

I have not updated the NEWS file but we can add the two commits later.

* https://github.com/avrdudes/avrdude/pull/1927
* https://github.com/avrdudes/avrdude/pull/1934

stefanrueger:
> I have not updated the NEWS file but we can add the two commits later.

Please do - lest we forget. Best if the person carrying out the merge documents that in NEWS

mcuee:
> > I have not updated the NEWS file but we can add the two commits later.
> 
> Please do - lest we forget. Best if the person carrying out the merge documents that in NEWS

Done.

Issue #1922, opened 2024-09-08
Support options like `--config` `--part` and `--port`


MCUdude:
I created a post over at the Avrfreaks forum announcing that Avrdude v8.0 has just been released. The feedback was mixed (some pointing out that Avrdude is just a CLI tool and not a full-blown GUI, and thus not relevant in 2024), but @westfw mentioned that the option flags (`-C`, `-c`, `-p` etc) could be difficult to remember, and having full-blown `--flags` would be nice.

Avrdude currently uses [`getopt`](https://www.gnu.org/software/libc/manual/html_node/Getopt.html), and this doesn't support `--flags`. Instead, we'll have to use [`argp`](https://www.gnu.org/software/libc/manual/html_node/Argp.html).

Are there any downsides to using `agrp` instead of `getopt`?

ndim:
In PR #1698, I have replaced the `getopt` call with `getopt_long` in order to add `--version`.

We already have a Windows implementation of `getopt_long` in `src/msvc/`, so that was relatively easy to do.

That covers at least glibc/Linux, FreeBSD and MacOS, and Windows.

ndim:
In PR #1698, I have replaced the `getopt` call with `getopt_long` in order to add `--version`.

While at it, I also added`--help`.

We already have a Windows implementation of `getopt_long` in `src/msvc/`, so that was relatively easy to do.

That covers at least glibc/Linux, FreeBSD and MacOS, and Windows.

The `argp` API is a lot more powerful, of course, but it appears avrdude would need to ship an `argp` implementation for all non-glibc platforms in order to use the `argp` API with its advantages like autogenerated `--help` output.

ndim:
To actually answer the question: The downside to using argp is the work needed to ensure its availability on all systems other than Linux/glibc.

WestfW:
getopt_long() seems to drop in pretty painlessly, and is already present in the windows getopt used by avrdude, and works on both macos and linux builds as well, so it should be pretty safe.

(from the avrfreaks discussion):
I replaced in implicit include for getopt() from unistd.h with an explicit include of getopt.h (which is already present in src/msvc/)

Then it was just a matter of replacing the getopt() call with one to getopt_long() (and adding the structure with the long form commands...)

Everything else stayed the same!

```
 // Process command line arguments
 static struct option long_options[] =
   {
     { "baud",      required_argument, 0, 'b' },
     { "bitclock",  required_argument, 0, 'B'},
[...]

PR #1936, opened 2024-11-27
Add long options like --part

ndim:
Switch from getopt(3) to getopt_long(3) and add a few long options as aliases to existing short options, e.g. "--help" and "--part" being aliases for "-?" and "-p", respectively.

The getopt_long(3) function is available on GNU, BSD, and the existing msvc/getopt.[ch] already implements getopt_long() on Windows. This should cover all systems avrdude supports.

- [X] Use `getopt_long(3)` instead of `getopt(3)`
- [X] Add long options from @westfw https://github.com/avrdudes/avrdude/issues/1922#issuecomment-2337274055
- [X] Adapt the avrdude usage message shown by "-?" or "--help" to show the new long options.
- [x] Refine, review and finalize the set of new long options
- [x] Adapt the man page to reflect the long options
- [x] Adapt the texi manual to reflect the long options

Closes: #1922

### DIFF
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -377,7 +377,7 @@ to trigger programming mode for certain boards like Arduino Leonardo,
 Arduino Micro/Pro Micro and the Arduino Nano Every. Longer waits, and
 therefore multiple -r options, are sometimes needed for slower, less
 powerful hosts.
-.It Fl B \-bitclock
+.It Fl B \-bitclock Ar bitclock
 Specify the bit clock period for the JTAG, PDI, TPI, UPDI, or ISP
 interface. The value is a floating-point number in microseconds.
 Alternatively, the value might be suffixed with "Hz", "kHz" or
--- a/src/avrdude.1
+++ b/src/avrdude.1
@@ -25,33 +25,32 @@
 .Nd driver program for ``simple'' Atmel AVR MCU programmer
 .Sh SYNOPSIS
 .Nm
-.Fl p Ar partno
-.Op Fl b Ar baudrate
-.Op Fl r
-.Op Fl B Ar bitclock
-.Op Fl c Ar programmer-id
-.Op Fl C Ar config-file
-.Op Fl N
+.Op Fl p, \-part Ar partno
+.Op Fl b, \-baud Ar baudrate
+.Op Fl B, \-bitclock Ar bitclock
+.Op Fl c, \-programmer Ar programmer-id
+.Op Fl C, \-config Ar config-file
+.Op Fl N, \-noconfig
 .Op Fl A
-.Op Fl D
-.Op Fl e
+.Op Fl D, \-noerase
+.Op Fl e, \-erase
 .Oo Fl E Ar exitspec Ns
 .Op \&, Ns Ar exitspec
 .Oc
 .Op Fl F
 .Op Fl i Ar delay
-.Op Fl l Ar logfile
-.Op Fl n
-.Op Fl O
-.Op Fl P Ar port
-.Op Fl r
-.Op Fl q
+.Op Fl l, \-logfile Ar logfile
+.Op Fl n, \-test
+.Op Fl O, \-osccal 
+.Op Fl P, \-port Ar port
+.Op Fl r, \-reconnect
+.Op Fl q, \-quell
 .Op Fl T Ar cmd
-.Op Fl t
-.Op Fl U Ar memory:op:filename:filefmt
-.Op Fl v
+.Op Fl t, \-terminal
+.Op Fl U, \-memory Ar memory:op:filename:filefmt
+.Op Fl v, \-verbose
 .Op Fl x Ar extended_param
-.Op Fl V
+.Op Fl V, \-noverify
 .Sh DESCRIPTION
 .Nm Avrdude
 is a program for downloading and uploading on-chip memories of Atmel AVR
@@ -326,7 +325,7 @@ In order to control all the different operation modi, a number of options
 need to be specified to
 .Nm avrdude .
 .Bl -tag -offset indent -width indent
-.It Fl p Ar partno
+.It Fl p \-part Ar partno
 This option specifies the MCU connected to the programmer. The MCU
 descriptions are read from the config file. To see a list of currently
 supported MCUs use ? as partno, which will print the part ids and official
@@ -351,19 +350,18 @@ programmers talking the STK500v2 protocol.
 .It "AT90S2343"
 The AT90S2323 and ATtiny22 use the same algorithm.
 .It "ATmega2560, ATmega2561"
-Flash addressing above 128 KB is not supported by all
-programming hardware.  Known to work are jtag2, stk500v2,
-and bit-bang programmers.
+Flash addressing above 128 KB may not be supported by all
+programming hardware.
 .It "ATtiny11"
 The ATtiny11 can only be
 programmed in high-voltage serial mode.
 .El
-.It Fl p Ar wildcard/flags
+.It Fl p \-part Ar wildcard/flags
 Run developer options for MCUs that are matched by wildcard. Whilst
 their main use is for developers some flags can be of utility for users, e.g.,
 avrdude -p m328p/S outputs AVRDUDE's understanding of ATmega328P MCU properties;
 for more information run avrdude -p x/h.
-.It Fl b Ar baudrate
+.It Fl b \-baud Ar baudrate
 Override the RS-232 connection baud rate specified in the respective
 programmer's entry of the configuration file or defined by the default_baudrate
 entry in your user configuration file
@@ -371,7 +369,7 @@ entry in your user configuration file
 or
 .Pa ${HOME}/.avrduderc
 if no baudrate was defined for this programmer.
-.It Fl r
+.It Fl r \-reconnect
 Opens the serial port at 1200 baud and immediately closes it, waits 400 ms
 for each -r on the command line and then establishes communication with
 the programmer. This is commonly known as a "1200bps touch", and is used
@@ -379,7 +377,7 @@ to trigger programming mode for certain boards like Arduino Leonardo,
 Arduino Micro/Pro Micro and the Arduino Nano Every. Longer waits, and
 therefore multiple -r options, are sometimes needed for slower, less
 powerful hosts.
-.It Fl B Ar bitclock
+.It Fl B \-bitclock
 Specify the bit clock period for the JTAG, PDI, TPI, UPDI, or ISP
 interface. The value is a floating-point number in microseconds.
 Alternatively, the value might be suffixed with "Hz", "kHz" or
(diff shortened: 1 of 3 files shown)


src/main.c
```diff
@@ -814,7 +817,32 @@ int main(int argc, char *argv[]) {
...
+    {"logfile",    required_argument, NULL, 'l'},
+    {"test",       no_argument,       NULL, 'n'},
+    {"noconfig",   no_argument,       NULL, 'N'},
+    {"part",       required_argument, NULL, 'p'},
+    {"chip",       required_argument, NULL, 'p'},
+    {"port",       required_argument, NULL, 'P'},
+    {"quiet",      no_argument,       NULL, 'q'},
+    {"reconnect",  no_argument,       NULL, 'r'},
+    {"terminal",   no_argument,       NULL, 't'},
+    {"tty",        no_argument,       NULL, 't'},
```
stefanrueger, line 837: Now that I look at the list carefully, I don't think `tty` is a good long option name to signify `-t` for the principal reason that tty describes a communication method (teletype), which is orthogonal to what `-t` does. Terminal input can come from a file, from a pipe, via a USB connection etc and not necessarily from a tty. I'd prefer to either have no second long option for `-t`, but *if* there is an appetite for a second option, I'd much rather have one of `--interactive`, `--console`, `--avr-console`, `--shell` or similar. We normally use the wording terminal or interactive terminal in the documentation for `-t`, so `--terminal` might be sufficient.

Could we add `--terminal-command` for `-T` (note capital T), which like `-U` is not interactive?

src/main.c
```diff
@@ -814,7 +817,32 @@ int main(int argc, char *argv[]) {
...
+    {"memory",     required_argument, NULL, 'U'},
```
stefanrueger, line 838: `--memory` is a good long option name b/c it puts the focus correctly: the operations `r`, `w` and `v` all refer to memory and, crucially, not the corresponding file; originally `-U` got its name from the generic update operations it might describe, so I would also allow the mnemonic `--update` or, alternatively, `--update-memory`. I have an ever so slight  preference for the latter: personally, I think long options are good in shell scripts to improve readability, so I am less worried about the length of long options. If we go with the latter, we wouldn't need `--memory`.

src/main.c
ndim: "shell" implies turning completeness. "interactive" captures the spirit well.

src/main.c
```diff
@@ -814,7 +817,32 @@ int main(int argc, char *argv[]) {
...
+  struct option longopts[] = {
+    {"help",       no_argument,       NULL, '?'},
+    {"baud",       required_argument, NULL, 'b'},
+    {"bitclock",   required_argument, NULL, 'B'},
+    {"programmer", required_argument, NULL, 'c'},
+    {"config",     required_argument, NULL, 'C'},
+    {"noerase",    no_argument,       NULL, 'D'},
+    {"erase",      no_argument,       NULL, 'e'},
+    {"logfile",    required_argument, NULL, 'l'},
+    {"test",       no_argument,       NULL, 'n'},
```
stefanrueger, line 829: `-n` meant originally don't write to the AVR part for `-U` operations, where `-n` would write the file to stdout instead of the AVR memory. When other ways to write to the AVR were implemented (`-t`, `-T`, ...) there was no way to test these in a dryrun as it were until the `-c` programmer `dryrun` was implemented. I now wonder whether `--test` should be `--test-U` or rather `--test-update` to clarify the role of that option.

stefanrueger:
Cool improvement! I like this PR

src/main.c
MCUdude: I would argue that `--memory` would be sufficient. The help test will get awfully long if we have tons of long options that to the same thing. And the -U flag doesn't necessary _updates_ the memory.

IMO, `--memory flash:r:myfile.hex` makes way more sense than `--update-memory flash:r:myfile.hex`.

ndim:
I wonder whether `-U` should even get a direct 1:1 replacement long option.

A set of `--read-memory` and `--write-memory` long opts with suitably adapted arguments could make more sense. And get rid of the colon problems on Windows.

MCUdude:
> I wonder whether -U should even get a direct 1:1 replacement long option.
A set of ``--read-memory`` and ``--verify-memory`` and ``--write-memory`` long opts with suitably adapted arguments could make more sense. And get rid of the colon problems on Windows while the argument is redefined.

Good idea! But how would we let the user specify the format ( :i, :r, :m etc.)?

``-read-memory flash:myfile.hex:i``

ndim:
> > I wonder whether -U should even get a direct 1:1 replacement long option.
> A set of ``--read-memory`` and ``--verify-memory`` and ``--write-memory`` long opts with suitably adapted arguments could make more sense. And get rid of the colon problems on Windows while the argument is redefined.
> 
> Good idea! But how would we let the user specify the format ( :i, :r, :m etc.)?
> 
> ``--read-memory flash:myfile.hex:i``

There are other characters which can be used as separators. Or `--read-memory (I|hex)? flash:c:/foobar/myfile.hex`. or...

stefanrueger:
> I guess `--(read|verify|write)-(eeprom|flash)` is not flexible enough to accommodate for complex parts' memory layouts?

Correct. Avrdude knows three operations, some 48 memories and 12 file formats. Is the suggestion for main.c to entertain some 144 long options for all (op, memory) combos? Or 1728 long options to expand the space to all (op, memory, file format) combos? There is the added difficulty that a new part in a custom config file may well introduce a new memory, which requires new long options to automatically spring up in that case.

> Or `--read-from flash` `--write myfile.hex` `--format hex`

Projecting the `-U` parameter space onto these axes  is a neat idea but complicates parsing, which would need to be aware that a list of more than one `-U` operations can be specified. Given that the format used to be optional (there are defaults depending on r/w/v) which file has the default format and which one the octal format in `--read-from flash` `--write file1.hex` `--format oct` `--write file2.hex` `--read-from eeprom`?

And how should the command line express `-U sketch`, which updates flash from the file `sketch`?

> `(i|hex)`

That's another pitfall: `i` signifies Intel Hex files, and `h` a simple comma-separated ASCII list of hexadecimal numbers in the format `0x..`, so `hex` as a long name for the format can be misleading. As would be `binary` or `bin` which could either imply raw data or a list of `0b....` binary numbers.

Also note that allowing a different way than appending `:x` to the file name express the format of the file will invariably draw demands on doing that for files in the terminal syntax, too (`verify flash sketch:d`). So, even simple ideas will require a lot of implementation work.

All of above *can* be done and done well, but should it? I believe the best thing for now is to have a one-to-one long replacement option for `-U` and deal with novel ways to express a series of `-U` memory manipulations in a separate PR that can be independently implemented, critiqued and tested...

MCUdude:
I think we should wrap it up and merge this PR.
We got long options, and if we can print a small help message if the user got the `-U` / `--memory` flag wrong, that could resolve most issues I think. If we also can mention the long options in error messages that mentions short options (like "clock speed too high, use -B 125kHz"), that would be really neat!

src/main.c
stefanrueger: Agreed, `--memory` is sufficient

src/main.c
stefanrueger: I still think `--test` is the wrong mnemonic. Given we have `--memory` for `-U` we should rename `--test` to `--test-memory` as its scope is confined to `-U/--memory` operations.

stefanrueger:
It would be great to wrap up this PR.

I support all of @MCUdude's suggestions [here](https://github.com/avrdudes/avrdude/pull/1936#issuecomment-2506659897).

Suggesting a new syntax for `-U` using long options or otherwise is probably best suggested in a different PR (as it's complex and can have a few uninteded ramifications as above discussion shows).

Over to you, @ndim to continue the great work here!

MCUdude:
@ndim sorry for pushing directly to your branch. I was trying to push to my own fork or your development branch, but I messed up somewhere on the way. If you look at the last commit, what I did was to add a few missing long opts to the help text, and remove duplicate log opts.

MCUdude:
I think the implementation itself is OK now. I've done the manpage documentation but still have the texinfo docs left.

MCUdude:
@stefanrueger I think this PR is mostly done now, so please review! The only longopt name I'm not so sure about is `--test`. Is this really the most accurate substitute for `-n`?

@ndim sorry for the noise. The reason why I messed up/couldn't fork your branch is because I already have a fork of the Avrdude project. And I misunderstood and accidentally pushed straight to your fork instead of to a "fork of your fork". But at least now the docs should be OK, and the missing longest names in the help text is now present.

src/avrdude.1
```diff
@@ -25,33 +25,32 @@
...
 .Op Fl T Ar cmd
-.Op Fl t
-.Op Fl U Ar memory:op:filename:filefmt
-.Op Fl v
+.Op Fl t, \-terminal
+.Op Fl U, \-memory Ar memory:op:filename:filefmt
+.Op Fl v, \-verbose
 .Op Fl x Ar extended_param
-.Op Fl V
+.Op Fl V, \-noverify
```
stefanrueger, line 53: In the same way, I prefer `--noverify-memory` as only the `--memory` operations are affected (again, changes need to be made multiple times)

src/avrdude.1
```diff
@@ -351,35 +350,34 @@ programmers talking the STK500v2 protocol.
...
+.It Fl r \-reconnect
 Opens the serial port at 1200 baud and immediately closes it, waits 400 ms
 for each -r on the command line and then establishes communication with
 the programmer. This is commonly known as a "1200bps touch", and is used
 to trigger programming mode for certain boards like Arduino Leonardo,
 Arduino Micro/Pro Micro and the Arduino Nano Every. Longer waits, and
 therefore multiple -r options, are sometimes needed for slower, less
 powerful hosts.
-.It Fl B Ar bitclock
+.It Fl B \-bitclock
```
stefanrueger, line 380: Missing `Ar bitclock` here

src/doc/avrdude.texi
```diff
@@ -509,8 +509,9 @@ following options are recognized:
 
 @table @code
 @item -p @var{partno}
-@cindex Option @code{-p} @var{partno}
-@cindex @code{-p} @var{partno}
+@item --part @var{partno}
+@cindex Option @code{-p ,--part} @var{partno}
```
stefanrueger, line 513: Should these not be two index lines?
```
@cindex Option @code{-p} @var{partname}
@cindex Option @code{--part} @var{partname}
```
(also note it's better to use partname instead of partno, as the part name is not really a number)

Multiple times. It's not so bad here, but when it's `--noerase` the reader wouldn't easily find that option in the alphabetic index as it's not sorted at n but rather at d (it trails `-D`)

MCUdude: Do you instead want it to be like this?

```
@item -p @var{partname}
@item --part @var{partname}
@cindex Option @code{-p} @var{partname}
@cindex Option @code{--part} @var{partname}
@cindex @code{-p} @var{partname}
@cindex @code{--part} @var{partname}
```

stefanrueger: Would be brill. Many places.

MCUdude: @stefanrueger done!

PR #1949, opened 2025-02-12
Correct PICkit5 programmer type entries

stefanrueger:
The programming type of a programmer is a label that maps to a `_initpgm()` initialisation routine in the driver via a table in `pgm_type.c`. The pickit5 programmer has one initialisation routine, so should only have one entry in `pgm_type.c`. This is now corrected. Please could you test whether the `-c pickit5_xxx` programmers still work as expected @MX682X


MX682X:
I have run test8 on this PR with ISP, old JTAG, TPI and UPDI without problems. I guess the rest will be fine too, or @MCUdude could run a quick test on xJTAG and PDI just to make sure.

MCUdude:
I just tested this PR with`-c pickit5_jtag -p atxmega256a3bu` and `-c pickit5_pdi -p atxmega256a3bu`, and it works perfectly fine.

PR #1950, opened 2025-02-12
Shorten ID of ATA parts

stefanrueger:
Traditionally, the id of a part is a shorter version of the full name, eg, `m328p` for `ATmega328P`. This PR shortens the ID of ATA parts, for which the ID has been the lowercase version of the full name. A similar exercise was earlier for the AVRnnXYmm parts that shortened their ID to nnxymm

### DIFF
--- a/src/avrdude.conf.in
+++ b/src/avrdude.conf.in
@@ -14926,9 +14926,9 @@ part parent ".classic" # m406
 # ATA6612C
 #------------------------------------------------------------
 
-part parent ".classic" # ata6612c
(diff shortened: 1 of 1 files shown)


dl8dtl:
For these parts, I'm not really sure. They are merely niche products anyway, and the short form is IMHO not easier to remember than the long one.

stefanrueger:
There is no utility for an ID that is just the lower-case partname, as AVRDUDE accepts the case-insensitive part name.

PR #1951, opened 2025-02-12
Shorten ID of LGT parts

stefanrueger:
As with AVRnnXYmm and ATAnnn parts, this PR shortens the ID of LGT parts.


dl8dtl:
Fine with me

PR #1958, opened 2025-03-13
Fix variant errors introduced in recent .atdf versions

stefanrueger:
Commit https://github.com/avrdudes/avrdude/commit/2229be97779d4f8234dd9fd0f566efdb88e2b6a5 incorporated changes from recent .atdf updates. Unfortunately, some .atdf updates erronously listed wrong variant names. This PR corrects the errors introduced in the recent .atdf files.


MCUdude:
Good catch! Looks good to me

PR #1964, opened 2025-03-22
Remove spurious s from AVR32SDnn part id

stefanrueger:
Somehow a trailing s crept into the AVR32SDnn part id definitions... There is no reason for that, and this PR corrects this.


mcuee:
This should be safe to merge.

PR #1992, opened 2025-05-19
Prevent (most) output of >= in Windows

### LINKED ISSUE #1991: Display issues under Windows
@stefanrueger 

This is a minor issue. The workaround may be simple, just need to add a space.
```
config <property> = 
```

Windows command prompt and Windows Terminal (Power Shell).

Image

No issues under MSYS2 prompt.

Image

stefanrueger:
Fixes #1991

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -1733,7 +1733,7 @@ static int cmd_config(const PROGRAMMER *pgm, const AVRPART *p, int argc, const c
       "to show that of <property>. Wildcards or initial strings are permitted (but\n"
       "not both), in which case all settings of matching properties are displayed.\n"
       "\n"
-      "avrdude> config <property>= # note the trailing = as in config eesave=\n"
+      "avrdude> config <property>=      # Note the trailing = as in config eesave=\n"
       "\n"
       "shows all possible values that <property> can take on with the currently\n"
       "set one being the only that is not commented out. Assignments\n"
--- a/src/term.c
+++ b/src/term.c
@@ -1733,7 +1733,7 @@ static int cmd_config(const PROGRAMMER *pgm, const AVRPART *p, int argc, const c
       "to show that of <property>. Wildcards or initial strings are permitted (but\n"
       "not both), in which case all settings of matching properties are displayed.\n"
       "\n"
-      "avrdude> config <property>=\n"
+      "avrdude> config <property>= # note the trailing = as in config eesave=\n"
       "\n"
       "shows all possible values that <property> can take on with the currently\n"
       "set one being the only that is not commented out. Assignments\n"
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -1428,9 +1428,9 @@ static int pickit5_read_dev_id(const PROGRAMMER *pgm, const AVRPART *p) {
     pickit5_send_script_cmd(pgm, scr, scr_len, NULL, 0);
     pickit5_program_enable(pgm, p);
     if(my.rxBuf[17] == 0x0E) {  // Errors figured out during 6 hours of failing to get it to work
-      if(my.rxBuf[16] == 0x10 || my.rxBuf[16] == 58) { // with the serial/bootloader auto-reset circuit on Arduino board
-        pmsg_error("debugWIRE transmission error, aborting");
-         msg_error("(make sure there are no caps and a pullup >= 10kOhm on the Reset line)\n");
+      if(my.rxBuf[16] == 0x10 || my.rxBuf[16] == 58) { // Serial/bootloader auto-reset circuit on Arduino board
+        pmsg_error("debugWIRE transmission error, aborting"
+                   " (ensure reset has a pullup >= 10 kOhm and no cap)\n");
       } else {
         pmsg_error("%d\n", my.rxBuf[16]);
       }
--- a/src/stk500v2.c
+++ b/src/stk500v2.c
@@ -2035,7 +2035,7 @@ static int stk500v2_jtag3_parseextparms(const PROGRAMMER *pgm, const LISTID extp
     msg_error("%s -c %s extended options:\n", progname, pgmid);
     if(pgm->extra_features & HAS_SUFFER) {
       msg_error("  -x suffer              Read SUFFER register value\n");
-      msg_error("  -x suffer=<n>          Set SUFFER register value to <n>=0x.., 0... or decimal\n");
+      msg_error("  -x suffer=<n>          Set SUFFER register value to <n>\n");
     }
     if(pgm->extra_features & HAS_VTARG_SWITCH) {
       msg_error("  -x vtarg_switch        Read on-board target voltage switch state\n");
--- a/src/usbasp.c
+++ b/src/usbasp.c
@@ -165,10 +165,10 @@ static int usbasp_transmit(const PROGRAMMER *pgm, unsigned char receive,
   unsigned char functionid, const unsigned char *send, unsigned char *buffer, int buffersize);
 
 #ifdef USE_LIBUSB_1_0
-static int usbOpenDevice(const PROGRAMMER *pgm, libusb_device_handle ** device, int vendor,
+static int usbOpenDevice(const PROGRAMMER *pgm, libusb_device_handle **device, int vendor,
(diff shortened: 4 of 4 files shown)


mcuee:
Somehow the issue is still there.

Windows CMD command prompt (same for Powershell)
Screenshot 2025-05-20 102833

src/term.c
```diff
@@ -1733,12 +1735,12 @@ static int cmd_config(const PROGRAMMER *pgm, const AVRPART *p, int argc, const c
       "to show that of <property>. Wildcards or initial strings are permitted (but\n"
       "not both), in which case all settings of matching properties are displayed.\n"
       "\n"
-      "avrdude> config <property>=\n"
+      "avrdude> config <property>" SPBS "=\n"
```
dl8dtl, line 1738: I think I would have preferred a space around the equation sign
`avrdude> config <property> = …\n`

mcuee: That is my preference as well, but @stefanrueger seems to say it does not work.
* https://github.com/avrdudes/avrdude/issues/1991#issuecomment-2890725093

dl8dtl: I would have preferred it in the syntax as well. ;-)

Of course, I think it is stupid of any shell to (mis)interpret the characters that have been printed by an application in their own way, but that's a different story.

stefanrueger: > I would have preferred it in the syntax as well. ;-)

:rofl: That's a whole different kettle of fish. Changing the syntax of AVRDUDE in order to modify the help text is like the tail wagging the dog, but, *really*, changing the syntax makes the parsing a bit involved. I'd rather get to the bottom of this.

> shell to (mis)interpret the characters

I suspect, but don't know, it's the display or window manager and not actually the shell.

@mcuee How does the output *look* when you compile and run the following C program?
```
#include <stdio.h>

int main() {
  printf(" -xsleep=<n> where <n>=1...666 is the sleep duration in ms\n");
}
```

(Bracing myself to see 666 being replaced with a devil emoji ...) Do you get the original ASCII when you cut the output from the window and paste it into the comment box here? Does any of this change when you set the locale to C as in, eg,
```
$ LC_ALL=C a.out
```

Is there an option for your shell or window manager for ASCII rather than UTF-8? An option for fixed-width teletype fonts? I mean, there *must* be a way for above simple program to print ASCII, not?

stefanrueger: ... and what else get's replaced in your environment? `<=`? `!=`? `<>`? `><`? `...`? `2^3`? ... There *must* be documentation for that kind of stuff, not?

stefanrueger:
Seeing that the space-backspace sequence doesn't help, this PR has now replaced the last commit with a verbal description of the trailing `=`. It's a fitting nod to equality for the Windows-impaired population:

```
$ avrdude -qq -p m328p -c dryrun -T "config -h" 2>&1 | grep trail -A3
avrdude> config <property>=      # Note the trailing = as in config eesave=

shows all possible values that <property> can take on with the currently
set one being the only that is not commented out. Assignments
```

src/term.c
mcuee: > Do you get the original ASCII when you cut the output from the window and paste it into the comment box here?

Yes, that is the intersting part. Looks like it has more to do with the display and not related to the Windows Powershell or CMD command processor here.

mcuee:
> More tests under Windows 11 23H2 (upgraded from Windows 10).
> 
> **Got issues**
> 
> 1. Windows Power Shell 1.0  (direct invocation or insider Windows Terminal)
> 2. Windows Power Shell 7.5.1 on top of Windows Terminal
> 3. Windows Command Prompt (CMD) (direct invocation or insider Windows Terminal)
> 
> **Got no issues --> 3) 4) and 5) are kind of interesting.** The following results seem to say both the shell and terminal play a part.
> 
> 1. Git Bash, as expcted
> 2. MSYS2 Bash shell (Windows Terminal or Mintty), as expected
>    https://www.msys2.org/docs/terminals/
> 3. Alacritty --> running on top of Windows Power Shell 1.0
> 4. Hypr --> running on top of Windows Command Prompt (CMD)
> 5. ConEmu

@stefanrueger 

From the above, it seems to say the issue is more with the display and not with the shell (Powershell or CMD).

More over, I cannot reproduce the issue with my laptop running Windows 11 Home 24H2 using git main. And this PR is good too.

I will check further on the laptop which has the issue (Windows 11 Enterprise 23H2).

mcuee:
@stefanrueger 

This PR is good. Tested with both PowerShell and CMD under Windows 11 23H2 laptop which had the issue before.

Screenshot 2025-05-21 103756

src/term.c
mcuee: Under PowerShell of the problematic Windows 11 23H2 laptop.

Screenshot -- not okay.
Screenshot 2025-05-21 104614

Copy/paste the output -- Okay
```
C:\work\avr\avrdude_test> cat .\test_geq.c
#include <stdio.h>

int main() {
  printf(" -xsleep=<n> where <n>=1...666 is the sleep duration in ms\n");
}

C:\work\avr\avrdude_test> .\test_geq.exe
 -xsleep=<n> where <n>=1...666 is the sleep duration in ms
C:\work\avr\avrdude_test>
```

mcuee: System Locale is already US English.
```
C:\work\avr\avrdude_test> Get-WinSystemLocale

LCID             Name             DisplayName
----             ----             -----------
1033             en-US            English (United States)
```

mcuee: Test C code.
```
#include <stdio.h>

int main() {
  printf(" <n> where n >=1 \n");
  printf(" <n> where n <=1 \n");
  printf(" <n> where n !=1 \n");
  printf(" <n> where n !=1 \n");
  printf(" <n> where n <>1 \n");
  printf(" <n> where n ><=1 \n");
  printf(" <n> where n =1...3 \n");
  printf(" <n> where n ='2^3' \n");
}
```

Powershell display of the above code and execution result.
Screenshot 2025-05-21 111442

Comparison of MSYS2 output
Screenshot 2025-05-21 111522

stefanrueger:
@mcuee Thanks! Given the many different combinations that get misrepresented, we cannot make AVRDUDE avoid all these. Will merge this PR

mcuee:
> @mcuee Thanks! Given the many different combinations that get misrepresented, we cannot make AVRDUDE avoid all these. Will merge this PR

Yes, this PR is good to go.

mcuee:
> More tests under Windows 11 23H2 (upgraded from Windows 10).
> 
> **Got issues**
> 
> 1. Windows Power Shell 1.0  (direct invocation or insider Windows Terminal)
> 2. Windows Power Shell 7.5.1 on top of Windows Terminal
> 3. Windows Command Prompt (CMD) (direct invocation or insider Windows Terminal)
> 
> **Got no issues --> 3) 4) and 5) are kind of interesting.** The following results seem to say both the shell and terminal play a part.
> 
> 1. Git Bash by itself, as expected
> 2. MSYS2 Bash shell (Windows Terminal or Mintty), as expected
>    https://www.msys2.org/docs/terminals/
> 3. Alacritty --> running on top of Windows Power Shell 1.0 -- Good
>
> 4. Hypr --> running on top of Windows Command Prompt (CMD) -- Good
>
> 5. ConEmu -- Good

mcuee:
Windows CMD.exe is actually okay, it is just that I choose to use `Windows Terminal` as the default terminal app to handle command line applications. 

Screenshot 2025-05-26 103353

Once I change that to `Windows Console Host`, it works.
Screenshot 2025-05-26 103624

mcuee:
@stefanrueger and @dl8dtl 

Wiki FAQ added.
https://github.com/avrdudes/avrdude/wiki/FAQ#i-have-some-strange-display-related-problems-with-windows-terminal-what-should-i-do

mcuee:
@stefanrueger 

I agree it is not really an avrdude issue.

Microsoft may have already fixed the issue -- it could be that Windows Terminal is not the root cause but some other Window system component.

As mentioned, I do not have such issue in another laptop running Windows 11 Home 24H2.
Screenshot 2025-05-26 185756

Windows 23H2 is still widely used -- need to see if other users still have this issue or not on Windows 11 23H2/24H2.

Issue #433, opened 2021-12-10
[bug #49082] avrdude 6.3 fails with AT90S8515


avrs-admin:
dg1afg <None>
Tue 13 Sep 2016 05:46:32 PM UTC

I have problems programming old AT90S8515 with avrdude 6.3. It's impossible to program the controller. The output dialog as below.
Reading the controller after programming shows every 2nd bit is $FF! avrdude 6.1 did work properly.
```
avrdude: Version 6.3
Copyright (c) 2000-2005 Brian Dean, [http://www.bdmicro.com/](https://savannah.nongnu.orghttp://www.bdmicro.com/)
Copyright (c) 2007-2014 Joerg Wunsch

System wide configuration file is "/etc/avrdude.conf"
User configuration file is "/home/shack/.avrduderc"
User configuration file does not exist or is not a regular file, skipping

avrdude: AVR device initialized and ready to accept instructions

Reading | ################################################## | 100% 0.01s

avrdude: Device signature = 0x1e9301 (probably 8515)
avrdude: stk500isp_read_byte(): invalid operation AVR_OP_READ on fuse memory
avrdude: stk500isp_read_byte(): invalid operation AVR_OP_READ on fuse memory
avrdude: safemode: Fuse reading not support by programmer.
Safemode disabled.
avrdude: NOTE: "flash" memory has been specified, an erase cycle will be performed
To disable this feature, specify the -D option.
avrdude: erasing chip
avrdude: reading input file "DCF77.hex"
avrdude: input file DCF77.hex auto detected as Intel Hex
avrdude: writing flash (4932 bytes):

Writing | ################################################## | 100% 112.85s

avrdude: 4932 bytes of flash written
avrdude: verifying flash memory against DCF77.hex:
avrdude: load data flash data from input file DCF77.hex:
avrdude: input file DCF77.hex auto detected as Intel Hex
avrdude: input file DCF77.hex contains 4932 bytes
avrdude: reading on-chip flash data:

Reading | ################################################## | 100% 14.81s

avrdude: verifying ...
avrdude: verification error, first mismatch at byte 0x0001
0xff != 0xc0
```
This issue was migrated from https://savannah.nongnu.org/bugs/?49082

avrdude-admin:
Joerg Wunsch <joerg_wunsch>
Tue 13 Sep 2016 06:33:33 PM UTC

This appears to have been broken with the elimination of
the old "ISP multi" command in SVN r1336.

[...]

PR #2016, opened 2025-06-14
Maybe fix Attiny15 fuse reads?

MX682X:
There is a second file change, src/config.h. It's probably from the expansion fix, so not by me.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -872,8 +872,8 @@ char *cfg_escapen(const char *s, size_t n) {
     case 0:
       *d++ = '\\';
       *d++ = '0';
-      // Expand \0 to \000 if UTF-8 chars or digits might *look* like an octal number
-      if(n > 1 && ((s[1] & 0x80) || (s[1] >= 0 && s[1] <= '9')))
+      // Expand \0 to \000 if followed by UTF-8 chars or digits that *look* like octal digits
+      if(n > 1 && ((s[1] & 0x80) || (s[1] >= '0' && s[1] <= '9')))
         *d++ = '0', *d++ = '0';
       break;
     case '\n':
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -1651,7 +1651,7 @@ static int pickit5_isp_read_fuse(const PROGRAMMER *pgm, const AVRMEM *mem, unsig
   unsigned char read_fuse_isp [] = {    // As we pull the command from avrdude's conf file, this isn't limited to fuses
     0x90, 0x00, 0x32, 0x00, 0x00, 0x00, // load 0x32 to r00
     0x1E, 0x37, 0x00,                   // Enable Programming?
-    0x9F,                               // Send status from temp_reg to host
+    //0x9F,                               // Send status from temp_reg to host
     0xA8, 0x00, 0x00, 0x00, 0x00,       // ???
     0x90, 0x01, (cmd >> 24), (cmd >> 16), (cmd >> 8), cmd, // load programming command to r01 (swapped bitorder)
     0x9B, 0x02, 0x03,                   // Load 0x03 to r02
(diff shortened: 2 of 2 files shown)


mcuee:
Related old discussions for ATtiny15.

* https://github.com/avrdudes/avrdude/pull/1386

mcuee:
@stefanrueger, @MCUdude and @dl8dtl 

BTW, the Windows CI build (MSVC and MinGW) failed due to the failure with Python binding. This is probably due to some very recnt changes in github, not so sure how to fix it.

By right it should be  'python313.lib'

```
LINK : fatal error LNK1104: cannot open file 'python313t.lib' [D:\a\avrdude\avrdude\build\src\swig_avrdude.vcxproj]
```

mcuee:
Reference for old parts
* https://github.com/avrdudes/avrdude/issues/433

avr1
mcu = attiny11, attiny12, attiny15, attiny28, at90s1200 (no SRAM)

avr2
mcu = attiny22, attiny26, at90s2313, at90s2323, at90s2333, at90s2343, at90s4414, 
at90s4433, at90s4434, at90c8534, at90s8515, at90s8535

Interestingly avr1 is no longer mentioned in latest gcc documentation.
https://gcc.gnu.org/onlinedocs/gcc/AVR-Options.html

I believe all these old parts have been obsoleted by Microchip.

BTW, Microchip has listed all ATmega, ATxmega and ATtiny parts as **Legacy Devices.**
https://www.microchip.com/en-us/about/corporate-overview/acquisitions/atmel

stefanrueger:
avr1 describes the opcode set. If those parts don't have SRAM then it's a little hard to compile for them (no variables beyond those that fit in the registers, hardware stack of three levels severely limits subroutines). Avr-gcc probably never catered for these?

stefanrueger:
For me, the fact that a part is out of production never implied that the AVRDUDE project should stop supporting it: boards with discontinued parts will still be around! However, if the programming tools that AVRDUDE interacts with do not cater for out-of-production parts there is little that we can do.

MCUdude:
Programming the ATtiny15 using the PICkit5 (PICkit4 in MPLAB mode) used to work without any errors, but I've recently updated the firmware to the latest version through MPLAB, in order to support the new AVR-SD chips. This update may have broken ATtiny15 support. It would be nice to sort this (and the Xmega issue) out, but I don't think we should halt the 8.1 release just because of this.

mcuee:
> Programming the ATtiny15 using the PICkit5 (PICkit4 in MPLAB mode) used to work without any errors, but I've recently updated the firmware to the latest version through MPLAB, in order to support the new AVR-SD chips. This update may have broken ATtiny15 support. It would be nice to sort this (and the Xmega issue) out, but I don't think we should halt the 8.1 release just because of this.

@MCUdude 

For the ATxmega issue, is it also related to the FW updates? Or is it a known issue?
* https://github.com/avrdudes/avrdude/discussions/1956#discussioncomment-13464165

MX682X:
@MCUdude could you tap in the ISP connection with an Oscilloscope / Logic Analyser and look what is actually happening?

MCUdude:
@MX682X here is a screenshot of my scope when probing MOSI, MISO, SCK and RST. I've also attached the decoded output.

This is the command I ran:

```
$ avrdude -cpickit4_mplab_isp -xvtarg=5 -pattiny15 -B32 -Ufuse:w:0x5c:m
Reading 1 byte for fuse from input file 0x5c
Writing 1 byte (0x5C) to fuse, 1 byte written
Warning: fuse verification mismatch
  device 0xff != input 0x5c at addr 0x0000 (error)
Error: fuse verification mismatch

Avrdude done.  Thank you.
```

SDS1204X_HD_PNG_1

[ATtiny15_decode.csv](https://github.com/user-attachments/files/20784906/ATtiny15_decode.csv)

I haven't used the scope for SPI decode before, so it took some time to verify that the decoded data were actually correct.
I verified this by usilizing the terminal `send` command. This doesn't work on the MPLAB programmers, so I used a USBasp to send four aritrary numbers, and tweaked my decoding settings so they would properly decode.

MX682X:
ok so I think we are just too fast when reading the data back. Assuming the Fuse Memory is an EEPROM, it would take 8.2ms to rewrite it. But the read back happens in about 2ms. according to the CSV file.
The Datasheet specifies data Polling - if an address is still being written, it will be read back as 0xFF. 
Intrestingly, The Pickit is trying to Poll the Ready flag, which does not exist on the Attiny15, but on the 328p for example. Or at least they didn't document the command (0xF0) in the Attiny15. 

I'll add a delay in the isp_fuse_write function with the value "min_write_delay" from the .config file

Issue #2017, opened 2025-06-16
Windows CI Build Python Binding Issue


mcuee:
@stefanrueger, @MCUdude and @dl8dtl 

Windows CI build (MSVC and MinGW) failed due to the failure with Python binding. This is probably due to some very recnt changes in github, not so sure how to fix it.

By right it should be  'python313.lib'

```
LINK : fatal error LNK1104: cannot open file 'python313t.lib' [D:\a\avrdude\avrdude\build\src\swig_avrdude.vcxproj]
```

Very recent build for avrdude 8.1RC branch is okay.
https://github.com/avrdudes/avrdude/actions/runs/15566912448

mcuee:
This may prevent us from building the Windows binaries for 8.1 Release.

mcuee:
One possbility is to disable the build of Python bindings.

Or I can manually build the Windows binary for 8.1 release.

mcuee:
Hopefully this is just a temp issue.

My latest avrdude fork github action builds are okay.
https://github.com/mcuee/avrdude/actions/runs/15679794795
https://github.com/mcuee/avrdude/actions/runs/15679912329

mcuee:
@dl8dtl 

Just wondering if you have a quick fix here. If not, any objection to disable the Windows Python binding for 8.1 release?

mcuee:
@stefanrueger and @dl8dtl 

We anyway need to find a solution for #1946.
* https://github.com/avrdudes/avrdude/issues/1946

mcuee:
Windows MSVC 32 bit Python binding error: the installer Python is 64bit only and cannot be used.
```
-- Found SWIG: C:/ProgramData/chocolatey/bin/swig.exe (found suitable version "4.1.1", minimum required is "4.0") found components: python
-- Could NOT find Python3 (missing: Python3_EXECUTABLE Interpreter) (found version "3.9.13")
    Reason given by package: 
        Interpreter: Wrong architecture for the interpreter "C:/hostedtoolcache/windows/Python/3.9.13/x64/python.exe"

-- Found SWIG but no Python3 header/library; cannot use SWIG
```

mcuee:
I believe Python version detection is not really correct in avrdude CMake file. So this is a real bug.

mcuee:
@stefanrueger 
I do not think I am able to fix the issue any time soon. My suggestion is to live with the current workaround and fix it later.

@ndim 
I am not so sure if you can help here. Thanks.

[...]

PR #2018, opened 2025-06-18
Update build.yml to disable Python Binding build for Windows

mcuee:
This is a workaround for Issue #2017.


mcuee:
@stefanrueger and @dl8dtl 

I will suggest we merge this PR for avrdude 8.1 release to workaround Issue #2017.
* #2017

PR #2021, opened 2025-06-18
Add delay after fuse writes to fix Attiny15 readback

MX682X:
as discussed in #2016


MCUdude:
Thanks! I'll give it a try tomorrow morning. But what's the deal with the changes in config.c? Doesn't this just revert what @stefanrueger commited earlier?

Issue #1797, opened 2024-04-28
CMake improvement -- using pkg-find to find the libraries


mcuee:
From the comment by @Youw here.
* https://github.com/avrdudes/avrdude/issues/1793#issuecomment-2081429919

++++++++++++++++++
>  just wondering if you have some ideas why libusb.h was not found properly

Yeap, this is the reason:

> For libusb project, we usually recommend using pkg-config to find libusb.h and not to use libusb-1.0/libusb.h in any case.
https://github.com/avrdudes/avrdude/blob/main/src/avrftdi_private.h
> ---
> still happening even if I add pkg-config as a macOS dependancy

It is not enough to install pkg-config. One must explicitly use it to find a package.

The propper way would be to remove things like `find_library(HAVE_LIBUSB_1_0 ...)` and similar nonsense with a propper use of cmake/pkg-config, e.g. in a similar way like HIDAPI [does it](https://github.com/libusb/hidapi/blob/bda710178dfe9650bdf76aa02e3671e17e3a8a27/libusb/CMakeLists.txt#L14-L16).

mcuee:
@ndim 

In case you continue the work on the CMake improvement along with autotools improvement, maybe you want to take a look at the above recommendations. Thanks.

ndim:
Now that the gpiod support requires pkg-config logic anyway, we can jump on board completely, both on the cmake and on the automake side.

Both for finding our own dependencies and for installing a `libavrdude-2.pc` alongside libavrdude.

mcuee:
This may also help for the MSVC build with vcpkg.

For example, we have the following two long standing issues and the current recommended work-around is to use MSYS2 mingw build.
* https://github.com/avrdudes/avrdude/issues/1253
* https://github.com/avrdudes/avrdude/issues/968

It may also help the following issue.
* https://github.com/avrdudes/avrdude/issues/1552

mcuee:
In the end we may be able to deprecate the following repo.
https://github.com/avrdudes/libhidapi

The following repo is not used anyway -- we can probably archive it.
https://github.com/avrdudes/libreadline

The following is still needed.
https://github.com/avrdudes/libelf

We may still need the following (to add CMake support).
https://github.com/avrdudes/libserialport

The following two can be optional -- we should probably let the user choose whether to use them or the generic library.
https://github.com/avrdudes/libftdi --> this is the root cause of Issue 1253
https://github.com/avrdudes/libusb --> this is the root cause of Issue #968.

[...]

PR #2026, opened 2025-06-20
Adding pkg-config as a Linux build dependency

mcuee:
This will fix the warnings in the container based cross build for i386, armhf and arm64 Linux.
```
  linux:
    runs-on: ubuntu-latest
    container: debian:11
    strategy:
      matrix:
        include:
          - { arch: i386, processor: i686, prefix: i686-linux-gnu, inc-lib: i386-linux-gnu }
          - { arch: armhf, processor: armhf, prefix: arm-linux-gnueabihf, inc-lib: arm-linux-gnueabihf }
          - { arch: arm64, processor: aarch64, prefix: aarch64-linux-gnu, inc-lib: aarch64-linux-gnu }
```

Current git main:
https://github.com/avrdudes/avrdude/actions/runs/15734446611/job/44343291592
```
-- Could NOT find PkgConfig (missing: PKG_CONFIG_EXECUTABLE) 
CMake Warning at CMakeLists.txt:279 (message):
  For using libgpiod, pkg-config would be required which is not available.
```

With this PR, it can find pkg-config. But it does not really enable the libgpiod build, probably a cross-build limitation.
```
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.2") 
-- Checking for module 'libgpiod'...
-- Could not find module 'libgpiod', proceeding without.
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,7 @@ jobs:
           build-essential
           automake
           libtool
+          pkg-config
           flex
           bison
           libelf-dev
@@ -95,6 +96,7 @@ jobs:
           sudo apt-get install -y
           build-essential
           cmake
+          pkg-config
           flex
           bison
           libelf-dev
@@ -167,6 +169,7 @@ jobs:
           apt-get install -y
           git
           cmake
+          pkg-config
           flex
           bison
           crossbuild-essential-${{matrix.arch}}


mcuee:
This may be a preparation for #1797 as well.
* https://github.com/avrdudes/avrdude/issues/1797

mcuee:
@stefanrueger 

I tend to think this PR is safe to merge, even though it does not really enable libgpiod cross build.

PR #2033, opened 2025-06-21
Update build.yml to rename macOS CI build to arm64

### LINKED ISSUE #2032: avrdude github build for macOS is no longer x86_64 version but rather arm64 version
github macOS runner is now of arm64 version and so is avrdude github CI build.
```
mcuee@Macmini github_bin/avrdude-macos-x86_64  > ls
avrdude		avrdude.conf
mcuee@Macmini github_bin/avrdude-macos-x86_64  > file avrdude
avrdude: Mach-O 64-bit executable arm64
```

mcuee:
This is to fix #2032.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -218,7 +218,7 @@ jobs:
             build/src/avrdude
             build/src/avrdude.conf
 
-  macos-x86_64:
+  macos-arm64:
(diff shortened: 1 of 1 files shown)


mcuee:
@stefanrueger and @MCUdude 

I think this is also good to have for avrdude 8.1 release.

PR #2035, opened 2025-06-23
DNM, testing Only: Python Binding CMake Improvement 

mcuee:


mcuee:
Close this for now as it is not ready and I have not figured out the way to go.

I may create a seperate PR just for the second portion.

PR #2036, opened 2025-06-23
To disable MSys2 Windows Python Binding in CI

mcuee:


mcuee:
@stefanrueger and @dl8dtl 

As of now, two issues are present and I cannot get Python GUI to work under MSYS2 mingw. Therefore I think it is better to disable the build in CI for avrdude 8.1 release.

1) It does not detect the right MSYS2 mingw Python.
2) Even if I fix the first issue and get the Python binding working, there is a PySide6 packaging issue which prevents the Python GUI to work properly.

Reference:
* https://github.com/avrdudes/avrdude/issues/1946#issuecomment-2994948635

PR #2051, opened 2025-08-29
CMakeLists.txt: allow disabling Python support

tpetazzoni:
In some cases, even if Swig is found and Python3 is found, it may not be desirable to build Python support in avrdude, so this commit adds an ENABLE_PYTHON_SUPPORT option to be able to explicitly disable using Python support (unfortunately CMake doesn't allow passing arguments that would prevent it from finding Swig/Python 3 if available).

To preserve existing behavior, this option defaults to enabled (ON).

PR #2054, opened 2025-09-15
Remove cmake from macOS build prerequisites installation

mcuee:


mcuee:
Refer to comments here.
* https://github.com/avrdudes/avrdude/pull/2051#issuecomment-3290230817

PR #2074, opened 2026-02-11
Fix buspirate fallback to byte write

alexwarg:
The buspirate logic tries to detect different versions of supported features by the buspirate. One case is that the buspirate does not support paged writes, in this case the pgm->paged_write function pointer is set to NULL. However, the avr_write code unconditionally calls pgm->paged_write and crashes the avrdude.

This change checks if the paged_write pointer is still set and falls back to byte-by-byte write if not. This allows for byte-by-byte write, while still using paged_read for verification.

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -1122,12 +1122,14 @@ int avr_write_mem(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m, int 
 
         if(auto_erase && pgm->page_erase && !mem_is_eeprom(cm))
           rc = pgm->page_erase(pgm, p, cm, pageaddr);
-        if(rc >= 0)
+        if(pgm->paged_write && (rc >= 0))
           rc = pgm->paged_write(pgm, p, cm, cm->page_size, pageaddr, cm->page_size);
-        if(rc < 0)
+        if(!pgm->paged_write || (rc < 0))
           failure = 1;          // Paged write failed, fall back to byte-at-a-time write below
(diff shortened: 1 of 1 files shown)


stefanrueger:
@alexwarg Thanks for reporting and for suggesting a PR!

In this case the problem was that `paged_load` exists but `paged_write` doesn't, while AVRDUDE (wrongly!) relies on either both or none existing. The patch below should also work to fix the reported bug:
``` diff
diff --git a/src/avr.c b/src/avr.c
index 79ee3c63..c3b8c042 100644
--- a/src/avr.c
+++ b/src/avr.c
@@ -1024,7 +1024,7 @@ int avr_write_mem(const PROGRAMMER *pgm, const AVRPART *p, const AVRMEM *m, int
     return i;
   }
   // HW programmers need a page size > 1, bootloader typ only offer paged r/w
-  if((pgm->paged_load && m->page_size > 1 && m->size%m->page_size == 0) ||
+  if((pgm->paged_write && m->page_size > 1 && m->size%m->page_size == 0) ||
     (is_spm(pgm) && avr_has_paged_access(pgm, p, m))) {
 
     // The programmer supports a paged mode write
```

I think your PR works (and thanks for submitting), but I thought it is better for me to review the code base for other cases where `paged_write` or `paged_load` might be used without checking.

I will submit a PR that supersedes yours.

stefanrueger:
Superseded by PR #2078

Issue #2079, opened 2026-03-08
Issues building under Tribblix/illumos


mcuee:
It is almost there but there is an issue.

Ignore the missing libftdi warnings.

```
jack@tribblix:~/build/avrdude$ uname -a
SunOS tribblix 5.11 tribblix-m38 i86pc i386 i86pc illumos

jack@tribblix:~/build/avrdude$ ./build.sh
-- The C compiler identification is GNU 10.3.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/gcc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Found Git: /usr/bin/git (found version "2.53.0")
-- Found FLEX: /usr/bin/flex (found version "2.6.4")
-- Found BISON: /usr/bin/bison (found version "3.8.2")
-- Looking for libelf.h
-- Looking for libelf.h - found
-- Looking for libelf/libelf.h
-- Looking for libelf/libelf.h - not found
-- Looking for usb.h
-- Looking for usb.h - found
-- Looking for lusb0_usb.h
-- Looking for lusb0_usb.h - not found
-- Looking for libusb.h
-- Looking for libusb.h - not found
-- Looking for libusb-1.0/libusb.h
-- Looking for libusb-1.0/libusb.h - found
-- Could NOT find SWIG (missing: SWIG_EXECUTABLE SWIG_DIR python) (Required is at least version "4.0")
-- Configuration summary:
-- ----------------------
-- DO HAVE    libelf
-- DO HAVE    libusb
-- DO HAVE    libusb_1_0
-- DON'T HAVE libhidapi
-- DON'T HAVE libftdi
-- DON'T HAVE libftdi1
-- DON'T HAVE libreadline
-- DON'T HAVE libserialport
-- DISABLED   doc
-- DISABLED   parport
-- DISABLED   linuxgpio
-- DISABLED   linuxspi
-- DON'T HAVE swig+Python3
-- ----------------------
-- Configuring done (1.7s)
-- Generating done (0.1s)
-- Build files have been written to: /jack/build/avrdude/build_sunos
[  1%] Generating avrdude.conf
[  1%] Built target conf
[  2%] [FLEX][Parser] Building scanner with flex 2.6.4
[  3%] [BISON][Parser] Building parser with bison 3.8.2
[  4%] Building C object src/CMakeFiles/libavrdude.dir/arduino.c.o
[...]

PR #2087, opened 2026-03-26
Simplify `buspirate_expect_bin_byte()` function

stefanrueger:
The original code contained an uneccesary `alloca()` function call, going another step towards #2079.

If buspirate isn't available for testing, perhaps someone competent in C can check the change and review it OK (@MCUdude?)

### DIFF
--- a/src/buspirate.c
+++ b/src/buspirate.c
@@ -137,24 +137,18 @@ static int buspirate_recv_bin(const PROGRAMMER *pgm, unsigned char *buf, size_t 
   return len;
 }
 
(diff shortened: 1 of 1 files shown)


stefanrueger:
> What do you want me to test?

Cool. Thanks for offering! Flashing some part using buspirate's binary mode (I don't know what that means, but it's where the changed function sits).

stefanrueger:
Ahh, yes. Reading buspirate_bb_enable() suggests BP prints which mode it operates in, sth like starting BP in xxx mode.

mcuee (approved):

Issue #1980, opened 2025-05-07
terminal mode "d" command is ambiguous


dl8dtl:
```
avrdude> d flash
Error: (cmd) command d is ambiguous: dump, disasm
```
The "d" command used to be a valid abbreviation for "dump". I think it makes sense to add it explicitly as such.

stefanrueger:
Well, it's the nature of an expanding command set.

It's been like that since 8.0.

It's effort to document `d`.

There are other, equivalent cases: `r` and `re` are no longer possible owing to the new `restore`. So, terminal `help` would need to explain `d`, `r`, `re`, `dump` and `read` are the same thing. And there would *still* be the rule that the input can use a shorter non-ambiguous name except this now is from the expanded set `d`, `r`, `re`, `dump`, `read`. It all gets messy.

Summarising, it is not great practice from a documentation and maintenance perspective to introduce these special rules. 

I'd rather remove the special lex `q` than add more single-letter special cases

mcuee:
@dl8dtl 

The recommendation seems to leave as it is. Any objection to close this issue as not planned?

PR #2095, opened 2026-04-07
Terminal mode: re-add "d" as alias for "dump"

dl8dtl:
"d" has been a valid alias for "dump" for a long time.  This was disturbed by adding the "disasm" command, as a single "d" was no longer unambiguous then.

Thus, restore "d" as an alias for "dump" now.

### DIFF
--- a/src/term.c
+++ b/src/term.c
@@ -95,6 +95,7 @@ static int cmd_quell(const PROGRAMMER *pgm, const AVRPART *p, int argc, const ch
 // List of commands; don't add a command starting with e: main.c relies on e expanding to erase
 static const struct command cmd[] = {
   {"dump", cmd_dump, _fo(read_byte_cached), "display a memory section as hex dump"},
+  {"d", cmd_dump, _fo(read_byte_cached), "abbreviation for dump"},
   {"read", cmd_dump, _fo(read_byte_cached), "alias for dump"},
   {"disasm", cmd_disasm, _fo(read_byte_cached), "disassemble a memory section"},
   {"write", cmd_write, _fo(write_byte_cached), "write data to memory; flash and EEPROM are cached"},


stefanrueger:
Oh well, it's a duplicate of #1980. I am still not a fan of `d` but if you *really* want it, OK. Would be good if that was documented, though. The `.texi` file has a `help` example that would need rerunning and pasting in and there are a couple of other places where `d` should be documented.

MCUdude:
> "d" has been a valid alias for "dump" for a long time

It has been a valid alias indeed, but never documented as such. I don't see why it is necessary to bring back `d` other than for those with `d` in their muscle memory.

mcuee:
> It has been a valid alias indeed, but never documented as such. I don't see why it is necessary to bring back d other than for those with d in their muscle memory.

At work we are dealing with product familiies which have been in the market for 25 years or more. I would say that often "undocumented features" become an important features for the customers.

In this case, I will support this PR from @dl8dtl.

stefanrueger:
1. The current PR only brings only one undocumented shot-hand command back. What about r and re?

2. It clutters the help message with two short-hand lines (q and d). If we continue using the same technique used for q and d also for r and re, it would be 4 cluttered lines.

3. The current PR does not document behaviour.

4. Current documentation (an example showing the help output in the manual is no longer correct.

Suggested improvements: make comment field for commands q and d empty (""). Add commands r and re with empty comment field. Change the code outputting the help message by *not* outputting help lines with empty comments. Add one line at the end of the help output saying that dump, quit and read can be abbreviated with d, q, r and re. Document this in .1 and .texi. Correct the help example in the .pdf

I am still not sure this necessary or super useful but at least addressing above would make for a neater solution to the intended behaviour

mcuee:
@dl8dtl 

Please help to address the comments by @stefanrueger if you want to proceed with this PR. Thanks.

dl8dtl:
> @dl8dtl
> 
> Please help to address the comments by @stefanrueger if you want to proceed with this PR. Thanks.

Oops, sorry, I missed them before.

dl8dtl:
> > "d" has been a valid alias for "dump" for a long time
> 
> It has been a valid alias indeed, but never documented as such.

Actually, it *was* documented. ;-) The possibility to abbreviate commands to the shortest unambiguous form has been documented for ages already.

It's only that the addition of the `disasm` command has made a simple `d` ambiguous later. (Likewise with `quell` that made a simple `q` no longer unambiguous.)

> I don't see why it is necessary to bring back `d` other than for those with `d` in their muscle memory.

I can't judge for others, but for me, `d` is actually *the* most frequently used command inside terminal mode.

> 1. The current PR only brings only one undocumented short-hand command back. What about r and re?

See above, it was documented. :-)

No personal opinion about `r`. Without looking, I didn't even remember it's there. :) But well, Brian added it 25 years ago …

> 2. It clutters the help message with two short-hand lines (q and d).

One completely different option would be to modify the search algorithm so that any shortened command matches the first entry within the table (even if ambiguous), and then re-order the table so `quit` comes before `quell`. (`dump` already comes before `disasm` anyway.) Of course, and then document the behaviour respectively.

How would that sound?

stefanrueger:
The original naming of commands as read, dump, quit etc together with the (documented) accepting unambiguous shorter names always ran the risk that new commands in future versions might render abbreviations ambiguous!

So, by original design, the use of d for dump has never been future-proof.

Introducing new, additional rules (eg, d means dump because d is no longer unambiguous) need to be documented.

Out of symmetry, any PR should afford read and dump the same logic: if dump is seen as important work-horse that deserves a special one-letter abbreviation then so should read. Cone to think of it, also write.

The table order is used for the help text and reflects the logic groupings of commands. Hence, reordering the table entails other problems. I recommend avoiding that.

Fun fact 1: I never liked the way q for quit was introduced and I always considered that a hack. Before anyone mentions it I realise *who* introduced that ;). I would prefer a better, systematic way to make d, r, w and q future-proof commands.

Fun fact 2: I quite liked d being a free unpaid advertisement for the existence of disasm :)

But yes, if someone really wants to future-proof d, r, w and q, that's fine with me.

stefanrueger:
Closed in favour of PR #2122

PR #2104, opened 2026-05-11
Add consistent support for `-P USB`

### LINKED ISSUE #2103: `-P USB` not working since Avrdude 7.3
Not sure if this is a bug or a feature, but I copy/pasted an Avrdude command I found online, and Avrdude didn't find my USBasp programmer, because the port was specified as `-p USB` with USB being in capital letters. I tested a few older Avrdude versions, and `-P USB` did work with Avrdude 7.2, but not in the more recent versions. Is this intensional or not? I suggest we support `USB` for the sake of compatibility. Any thoughts?

MCUdude:
It seems like support for upper case `USB` could vary from programmer implementation to programmer implementation in earlier versions of Avrdude.

Resolves #2103

### DIFF
--- a/src/avrftdi.c
+++ b/src/avrftdi.c
@@ -642,7 +642,7 @@ static int avrftdi_open(PROGRAMMER *pgm, const char *port) {
   Avrftdi_data *pdata = to_pdata(pgm);
 
   pmsg_debug("%s(\"%s\")\n", __func__, port);
-  if(!str_eq(port, "usb"))
+  if(!str_caseeq(port, "usb"))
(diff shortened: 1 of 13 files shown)


mcuee (approved):

mcuee:
Nice cleanup!

stefanrueger:
Thanks @MCUdude! Note line 826 in ft245.c misses sscanf, otherwise all looks great

MCUdude:
> Thanks [@MCUdude](https://github.com/MCUdude)! Note line 826 in ft245.c misses sscanf, otherwise all looks great

Oops, fixed in 50ce36f

PR #2126, opened 2026-06-16
Libavrdude library

mcuee:
Patch created by Claude free chatbot (Sonnet 4.6 Low).


mcuee:
This is not working yet.

I want to document the experiment to see if the free tier of Claude can address this more complex issue.

mcuee:
# Patch Proposal: Fix `libavrdude` shared library build on Windows (issue #1414)

## Root cause

`libavrdude` calls `avrdude_message2()` and references global variables
(`verbose`, `progname`, `ovsigck`, `partdesc`, `pgmid`, `quell_progress`, `cx`)
that are currently **defined in `main.c`** (the executable).

On Linux/macOS this accidentally works via `-Wl,-undefined,dynamic_lookup` or
flat-namespace linking, but **Windows DLLs require all symbols to be fully
resolved at library link time**. This causes `BUILD_SHARED_LIBS=ON` to fail
with hundreds of `undefined reference` errors on both MinGW and MSVC.

## Fix

Move all affected definitions into a new file `src/avrdude_msg.c`, compiled
into `libavrdude`. The changes to `main.c` are **purely deletions** — all
existing assignments (`verbose = 0`, `pgmid = optarg`, etc.) remain unchanged,
they simply now write into variables owned by the library.

Moving `cx` (`libavrdude_context *`) here is consistent with the existing
comment in `libavrdude.h`:
> *"Global and static variables should go here; the only remaining static
> variables ought to be read-only tables."*

**No behaviour change. Zero logic modifications.**

---

## Files changed

| File | Change |
|---|---|
| `src/avrdude_msg.c` | **New file** — receives globals + `avrdude_message2()` cut from `main.c` |
| `src/main.c` | Deletions only |
| `src/CMakeLists.txt` | Add `avrdude_msg.c` to `SOURCES`; add `LIBAVRDUDE_BUILD_DLL` for Windows shared builds |
| `src/avrdude.h` | Add `LIBAVRDUDE_API` macro; annotate the six `extern` declarations and `avrdude_message2()` |
| `src/libavrdude.h` | Fix stale comment on `avrdude_message2()` |

Root `CMakeLists.txt` is **not changed**.

---

## 1. New file: `src/avrdude_msg.c`

```c
/*
 * avrdude - A Downloader/Uploader for AVR device programmers
 * Copyright (C) 2000-2005 Brian S. Dean <bsd@bdmicro.com>
 * Copyright (C) Joerg Wunsch <j@uriah.heep.sax.de>
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */

/*
 * Global variables and avrdude_message2() moved here from main.c so that
 * libavrdude is fully self-contained and can be built as a shared library
 * (.dll / .so / .dylib) on all platforms including Windows.
 *
 * Previously these symbols were defined in main.c (the executable) and
 * referenced by libavrdude, creating a circular dependency that Windows
 * DLL linking strictly forbids.  Moving them into the library eliminates
 * that inversion: the executable now depends on the library only, not
 * vice versa.
 *
 * This also moves cx (libavrdude_context *) here from main.c, consistent
 * with the stated direction in libavrdude.h that cx should eventually be
 * the only remaining global variable.
 *
 * See https://github.com/avrdudes/avrdude/issues/1414
 */

#include <stdio.h>
#include <stdlib.h>
#include <stdarg.h>
#include <string.h>
#include <ctype.h>

#include "avrdude.h"
#include "libavrdude.h"

// ---------------------------------------------------------------------------
// Global variables formerly defined in main.c
// ---------------------------------------------------------------------------

libavrdude_context *cx;         // Context pointer, eventually the only global variable

char       *progname       = "avrdude";
int         verbose        = 0;
int         quell_progress = 0;
int         ovsigck        = 0;
const char *partdesc       = NULL;
const char *pgmid          = NULL;

// ---------------------------------------------------------------------------
// avrdude_message2() and its private helper — body identical to main.c
// ---------------------------------------------------------------------------

static const char *avrdude_message_type(int msglvl) {
  switch(msglvl) {
  case MSG_EXT_ERROR: return "OS error";
  case MSG_ERROR:     return "error";
  case MSG_WARNING:   return "warning";
  case MSG_INFO:      return "info";
  case MSG_NOTICE:    return "notice";
  case MSG_NOTICE2:   return "notice2";
  case MSG_DEBUG:     return "debug";
  case MSG_TRACE:     return "trace";
  case MSG_TRACE2:    return "trace2";
  default:            return "unknown msglvl";
  }
}

/*
 * Core messaging routine for msg_xyz(), [pli]msg_xyz() and term_out()
 * See #define lines in avrdude.h for how it is normally called.
 *
 * Named that way as there used to be a now gone different avrdude_message()
 */
int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
  int msgmode, int msglvl, const char *format, ...) {

  int rc = 0;
  va_list ap;

  static struct {               // Memorise whether last print ended at beginning of line
    FILE *fp;
    int bol;                    // Are we at the beginning of a line for this fp stream?
  } bols[5 + 1];                // Cater for up to 5 different FILE pointers plus one catch-all

  size_t bi = 0;

  for(bi = 0; bi < sizeof bols/sizeof *bols - 1; bi++) {
    if(!bols[bi].fp) {
      bols[bi].fp = fp;
      bols[bi].bol = 1;
    }
    if(bols[bi].fp == fp)
      break;
  }

  if(msglvl <= MSG_ERROR)       // Serious error? Free progress bars (if any)
    report_progress(1, -1, NULL);

  if(msgmode & MSG2_FLUSH) {
    fflush(stdout);
    fflush(stderr);
  }

  // Reduce effective verbosity level by number of -q above one when printing to stderr
  if((fp == stderr? verblevel: verbose) >= msglvl) {
    if(msgmode & MSG2_LEFT_MARGIN && !bols[bi].bol) {
      fprintf(fp, "\n");
      bols[bi].bol = 1;
    }
    // Keep vertical tab at start of format string as conditional new line
    if(*format == '\v') {
      format++;
      if(!bols[bi].bol) {
        fprintf(fp, "\n");
        bols[bi].bol = 1;
      }
    }

    if(msgmode & (MSG2_PROGNAME | MSG2_TYPE)) {
      if(msgmode & MSG2_PROGNAME) {
        fprintf(fp, "%s", progname);
        bols[bi].bol = 0;
      }
      if(msgmode & MSG2_TYPE) {
        const char *mt = avrdude_message_type(msglvl);

        if(bols[bi].bol)
          fprintf(fp, "%c%s", msgmode & (MSG2_UCFIRST)? toupper(*mt & 0xff): *mt, mt + 1);
        else
          fprintf(fp, " %s", mt);
        bols[bi].bol = 0;
      }
      if(verblevel >= MSG_NOTICE2) {
        const char *bfname = strrchr(file, '/');

#if defined(WIN32)
        if(!bfname)
          bfname = strrchr(file, '\\');
#endif

        bfname = bfname? bfname + 1: file;
        if(msgmode & MSG2_FUNCTION)
          fprintf(fp, " %s()", func);
        if(msgmode & MSG2_FILELINE)
          fprintf(fp, " %s %d", bfname, lno);
      }
      fprintf(fp, ": ");
    } else if(msgmode & MSG2_INDENT1) {
      fprintf(fp, "%*s", (int) strlen(progname) + 1, "");
      bols[bi].bol = 0;
    } else if(msgmode & MSG2_INDENT2) {
      fprintf(fp, "%*s", (int) strlen(progname) + 2, "");
      bols[bi].bol = 0;
    }

    va_start(ap, format);
    rc = vsnprintf(NULL, 0, format, ap);
    va_end(ap);

    if(rc < 0)
      return 0;

    rc++;
    char *p = mmt_malloc(rc);

    va_start(ap, format);
    rc = vsnprintf(p, rc, format, ap);
    va_end(ap);

    if(rc < 0) {
      mmt_free(p);
      return 0;
    }

    if(*p) {
      if(bols[bi].bol && (msgmode & MSG2_UCFIRST))
        fprintf(fp, "%c%s", toupper(*p & 0xff), p + 1);
      else
        fprintf(fp, "%s", p);
      bols[bi].bol = p[strlen(p) - 1] == '\n';
    }
    mmt_free(p);
  }

  if(msgmode & MSG2_FLUSH)
    fflush(fp);

  return rc;
}
```

---

## 2. `src/main.c` — deletions only

Remove the following items. Nothing else in `main.c` changes.

```c
// DELETE: file-scope definition (line ~53)
char *progname = "avrdude";

// DELETE: entire static helper function
static const char *avrdude_message_type(int msglvl) { ... }

// DELETE: entire function
int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
  int msgmode, int msglvl, const char *format, ...) { ... }

// DELETE: near struct list_walk_cookie
libavrdude_context *cx;

// DELETE: Global options block
int verbose;
int quell_progress;
int ovsigck;
const char *partdesc;
const char *pgmid;
```

---

## 3. `src/CMakeLists.txt` — two additions

**Addition 1:** add `avrdude_msg.c` to the `SOURCES` list (alphabetical position between `arduino.c` and `avr.c`):

```diff
 set(SOURCES
     ${CMAKE_CURRENT_BINARY_DIR}/ac_cfg.h
     arduino.h
     arduino.c
+    avrdude_msg.c
     avr.c
```

**Addition 2:** add after the `set_target_properties(libavrdude ...)` block:

```diff
 set_target_properties(libavrdude PROPERTIES
     PREFIX ""
     PUBLIC_HEADER "libavrdude.h;libavrdude-avrintel.h"
     VERSION 4.0.0
     SOVERSION 4
     )
+
+# Export DLL symbols on Windows when building as a shared library.
+# avrdude_message2(), progname, verbose etc. formerly lived in the
+# executable (main.c), causing undefined-reference link failures on
+# Windows DLLs. They now live in libavrdude via avrdude_msg.c.
+if(BUILD_SHARED_LIBS AND WIN32)
+    target_compile_definitions(libavrdude PRIVATE LIBAVRDUDE_BUILD_DLL)
+endif()
```

Note: the `swig_avrdude` target uses `${SOURCES}` directly and will pick up
`avrdude_msg.c` automatically — no further changes needed there.

---

## 4. `src/avrdude.h` — add `LIBAVRDUDE_API` macro and annotate declarations

Add the macro block after the `#include <stdlib.h>` line and before
`#define SYSTEM_CONF_FILE`:

```c
// DLL import/export annotation for Windows shared library builds.
// When compiling libavrdude itself (LIBAVRDUDE_BUILD_DLL set by CMake),
// symbols are exported. When compiling a consumer (main.c, external app),
// symbols are imported. On non-Windows the macro expands to nothing,
// except on GCC >= 4 where it sets default visibility explicitly.
#if defined(_WIN32) && defined(LIBAVRDUDE_BUILD_DLL)
#  define LIBAVRDUDE_API __declspec(dllexport)
#elif defined(_WIN32)
#  define LIBAVRDUDE_API __declspec(dllimport)
#else
#  if defined(__GNUC__) && __GNUC__ >= 4
#    define LIBAVRDUDE_API __attribute__((visibility("default")))
#  else
#    define LIBAVRDUDE_API
#  endif
#endif
```

Then annotate the existing `extern` declarations and `avrdude_message2`:

```diff
-extern char *progname;
-extern int ovsigck;
-extern int verbose;
-extern int quell_progress;
-extern const char *partdesc;
-extern const char *pgmid;
+LIBAVRDUDE_API extern char       *progname;
+LIBAVRDUDE_API extern int         ovsigck;
+LIBAVRDUDE_API extern int         verbose;
+LIBAVRDUDE_API extern int         quell_progress;
+LIBAVRDUDE_API extern const char *partdesc;
+LIBAVRDUDE_API extern const char *pgmid;
```

```diff
-int avrdude_message2(FILE *fp, int lno, const char *file, const char *func, int msgmode, int msglvl, const char *format, ...)
+LIBAVRDUDE_API int avrdude_message2(FILE *fp, int lno, const char *file, const char *func, int msgmode, int msglvl, const char *format, ...)
 #if defined(__GNUC__)
   __attribute__((format(printf, 7, 8)))
 #endif
   ;
```

---

## 5. `src/libavrdude.h` — fix stale comment

```diff
-// This functions is supposed to be supplied by the application
-int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
-  int msgmode, int msglvl, const char *format, ...);
+// Core messaging function, defined in avrdude_msg.c (compiled into libavrdude)
+int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
+  int msgmode, int msglvl, const char *format, ...);
```

---

## Verification

To test the shared build on Linux (sanity check before Windows):

```sh
cmake -S . -B build/shared -DBUILD_SHARED_LIBS=ON
cmake --build build/shared
# Confirm no dangling undefined symbols in the .so:
nm -u build/shared/src/libavrdude.so | grep -v UND
```

To test on Windows (MinGW):

```sh
cmake -S . -B build/shared -DBUILD_SHARED_LIBS=ON
cmake --build build/shared
# Should produce libavrdude.dll and libavrdude.dll.a with no link errors
```

Previously this would fail with hundreds of `undefined reference to avrdude_message2`
and `undefined reference to verbose` etc. errors at the DLL link step.

mcuee:
Problem with the proposed fix.
1) Not dealing with Python binding build properly
2) MSVC build failure (even without Python binding)
3) MinGW build failure (no Python binding)

stefanrueger:
Just a quick note that `avrdude_message2()` should *not* be part of libavrdude. The idea is that each project using libavrdude provides its own messaging system. AVRDUDE's `avrdude_message2()` function in main.c (which is not part of libavrdude) prints to the console, but that's not suitable for a GUI, which might want to show a video of dancing hamsters holding a banner with the corresponding notices.

mcuee:
New proposal from Claude.

# Revised Proposal for Issue #1414 — `libavrdude` Shared Library on Windows

## Acknowledging the Developer's Design Intent

stefanrueger's comment is exactly right:

> *"avrdude_message2() should not be part of libavrdude. The idea is that each
> project using libavrdude provides its own messaging system."*

The original patch moved `avrdude_message2()` into `libavrdude`, which
contradicts this design. A GUI should be able to intercept messages (log to a
widget, filter, suppress, redirect) — not be forced to accept the CLI's
console-printing implementation.

The correct fix is therefore the **opposite**: keep `avrdude_message2()` in the
application (`main.c`) as the developer intends, but move only the **global
state variables** into `libavrdude`, since those are what prevents the library
from linking as a DLL.

---

## Root Cause (unchanged)

`libavrdude` references these symbols defined in `main.c`:

- `int verbose`, `int quell_progress`, `int ovsigck`
- `const char *partdesc`, `const char *pgmid`
- `char *progname`
- `libavrdude_context *cx`

Windows DLLs must resolve all symbols at link time. These backwards references
from the library into the executable make `BUILD_SHARED_LIBS=ON` fail on both
MinGW and MSVC.

`avrdude_message2()` is called inside `libavrdude` through the `msg_xyz()` and
`pmsg_xyz()` macros, but its **definition** is intentionally left to the
application. This is correct design — the fix must not break it.

---

## Revised Fix: Move only state globals into `libavrdude`

The right approach:

1. Move `verbose`, `quell_progress`, `ovsigck`, `partdesc`, `pgmid`, `progname`,
   and `cx` into a new `src/avrdude_globals.c` compiled into `libavrdude`.
2. Keep `avrdude_message2()` in `main.c` exactly where it is.
3. The `libavrdude.h` comment *"This function is supposed to be supplied by the
   application"* remains correct and unchanged.

This also fixes the SWIG (`swig_avrdude`) build failure from the first patch:
since `avrdude_message2()` stays in `main.c` and is not in `${SOURCES}`, the
SWIG target (which uses `${SOURCES}`) never sees a duplicate definition.

---

## Files Changed

| File | Change |
|---|---|
| `src/avrdude_globals.c` | **New file** — state globals only, no `avrdude_message2()` |
| `src/main.c` | Remove 6 global definitions and `cx`; keep `avrdude_message2()` intact |
| `src/CMakeLists.txt` | Add `avrdude_globals.c` to `SOURCES`; add `LIBAVRDUDE_BUILD_DLL` for Windows |
| `src/avrdude.h` | Add `LIBAVRDUDE_API` macro; annotate the `extern` variable declarations only |
| `src/libavrdude.h` | No change needed — comment remains correct |

---

## 1. New file: `src/avrdude_globals.c`

```c
/*
 * avrdude - A Downloader/Uploader for AVR device programmers
 * Copyright (C) 2000-2005 Brian S. Dean <bsd@bdmicro.com>
 * Copyright (C) Joerg Wunsch <j@uriah.heep.sax.de>
 *
 * This program is free software; you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation; either version 2 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program. If not, see <http://www.gnu.org/licenses/>.
 */

/*
 * Global state variables moved here from main.c so that libavrdude is
 * self-contained and can be built as a shared library on all platforms
 * including Windows.
 *
 * Windows DLLs cannot have unresolved symbol references at link time.
 * Previously these variables were defined in main.c (the executable) and
 * referenced by libavrdude, which is a circular dependency that Windows
 * DLL linking forbids. Moving the definitions into the library eliminates
 * that inversion: the executable now depends only on the library.
 *
 * Note: avrdude_message2() deliberately remains in main.c (not here).
 * Each application using libavrdude supplies its own implementation, as
 * documented in libavrdude.h. A GUI, for instance, may route messages to
 * a widget rather than printing to the console.
 *
 * Moving cx (libavrdude_context *) here is consistent with the comment
 * in libavrdude.h: "Global and static variables should go here; the only
 * remaining static variables ought to be read-only tables."
 *
 * See https://github.com/avrdudes/avrdude/issues/1414
 */

#include "avrdude.h"
#include "libavrdude.h"

libavrdude_context *cx;         // Context pointer, eventually the only global variable

char       *progname       = "avrdude";
int         verbose        = 0;
int         quell_progress = 0;
int         ovsigck        = 0;
const char *partdesc       = NULL;
const char *pgmid          = NULL;
```

---

## 2. `src/main.c` — targeted deletions only

Remove exactly the following lines. `avrdude_message2()` and its helper
`avrdude_message_type()` are **not** touched.

```diff
-char *progname = "avrdude";
```

```diff
-libavrdude_context *cx;         // Context pointer, eventually the only global variable
```

```diff
-// Global options
-int verbose;                    // Verbose output
-int quell_progress;             // Quell progress report and un-verbose output
-int ovsigck;                    // 1 = override sig check, 0 = don't
-const char *partdesc;           // Part -p string
-const char *pgmid;              // Programmer -c string
```

Everything else in `main.c` — including the full `avrdude_message2()` function
— remains exactly as it is.

---

## 3. `src/CMakeLists.txt` — two additions

**Addition 1:** Add `avrdude_globals.c` to `SOURCES` (between `arduino.c` and `avr.c`):

```diff
 set(SOURCES
     ${CMAKE_CURRENT_BINARY_DIR}/ac_cfg.h
     arduino.h
     arduino.c
+    avrdude_globals.c
     avr.c
```

**Addition 2:** After the `set_target_properties(libavrdude ...)` block:

```diff
 set_target_properties(libavrdude PROPERTIES
     PREFIX ""
     PUBLIC_HEADER "libavrdude.h;libavrdude-avrintel.h"
     VERSION 4.0.0
     SOVERSION 4
     )
+
+# On Windows, export the global state variables from the DLL.
+# These were formerly defined in main.c (the executable), which prevented
+# libavrdude from linking as a shared library. They now live in
+# avrdude_globals.c (compiled into libavrdude).
+# Note: avrdude_message2() is intentionally NOT exported from the DLL;
+# each application supplies its own implementation per libavrdude.h.
+if(BUILD_SHARED_LIBS AND WIN32)
+    target_compile_definitions(libavrdude PRIVATE LIBAVRDUDE_BUILD_DLL)
+endif()
```

The `swig_avrdude` target uses `${SOURCES}` and picks up `avrdude_globals.c`
automatically. Since `avrdude_message2()` is **not** in `SOURCES`, there is no
duplicate definition issue with `swig_avrdude` — each SWIG-based application
supplies its own messaging implementation as intended.

---

## 4. `src/avrdude.h` — add `LIBAVRDUDE_API`, annotate variable externs only

Add the macro block after `#include <stdlib.h>` and before
`#define SYSTEM_CONF_FILE`:

```c
// DLL import/export annotation for Windows shared library builds.
// Applied only to global state variables exported from libavrdude.
// avrdude_message2() is intentionally not annotated — it is supplied
// by each application, not exported from the library.
#if defined(_WIN32) && defined(LIBAVRDUDE_BUILD_DLL)
#  define LIBAVRDUDE_API __declspec(dllexport)
#elif defined(_WIN32)
#  define LIBAVRDUDE_API __declspec(dllimport)
#else
#  if defined(__GNUC__) && __GNUC__ >= 4
#    define LIBAVRDUDE_API __attribute__((visibility("default")))
#  else
#    define LIBAVRDUDE_API
#  endif
#endif
```

Annotate **only the variable declarations** — not `avrdude_message2()`:

```diff
-extern char *progname;
-extern int ovsigck;
-extern int verbose;
-extern int quell_progress;
-extern const char *partdesc;
-extern const char *pgmid;
+LIBAVRDUDE_API extern char       *progname;
+LIBAVRDUDE_API extern int         ovsigck;
+LIBAVRDUDE_API extern int         verbose;
+LIBAVRDUDE_API extern int         quell_progress;
+LIBAVRDUDE_API extern const char *partdesc;
+LIBAVRDUDE_API extern const char *pgmid;
```

`avrdude_message2()` declaration is left exactly as-is — no `LIBAVRDUDE_API`.

---

## 5. `src/libavrdude.h` — no change

The existing comment:

```c
// This functions is supposed to be supplied by the application
int avrdude_message2(...);
```

...is correct and should stay exactly as written. This patch does not change it.

---

## Why this also fixes the SWIG/Python build failure

The first patch put `avrdude_message2()` into `avrdude_msg.c`, which was part
of `${SOURCES}`. Since `swig_avrdude` also uses `${SOURCES}`, it compiled
`avrdude_message2()` into the SWIG module — but each SWIG-based application is
supposed to supply its own implementation, causing a conflict.

This revised patch puts **only the state globals** into `avrdude_globals.c`.
The SWIG module compiles those globals in (correct — it needs them), but
`avrdude_message2()` remains the responsibility of the application layer
(correct — the Python bindings can supply a Python-friendly implementation).

---

## Why this also fixes the MSVC build failure

The first patch annotated `avrdude_message2()` with `LIBAVRDUDE_API`, which
caused MSVC to expect a `dllexport`/`dllimport` on a function that is
intentionally defined outside the DLL. This revision does not annotate
`avrdude_message2()` at all, removing that conflict.

---

## Summary

| | First patch (broken) | This revision |
|---|---|---|
| `avrdude_message2()` location | Moved into `libavrdude` | Stays in `main.c` ✓ |
| State globals location | Moved into `libavrdude` | Moved into `libavrdude` ✓ |
| SWIG build | Duplicate `avrdude_message2` ✗ | No conflict ✓ |
| MSVC build | `dllexport` on app-supplied fn ✗ | Only state vars exported ✓ |
| Developer's design intent | Violated | Respected ✓ |

mcuee:
Claude found some comments from PR #1874.
* https://github.com/avrdudes/avrdude/pull/1874

From @ndim 
> > `avrdude_message2()` could be moved [to] libavrdude  [...] would need to be made available to a user in some way

From @stefanrueger 
> I like this idea. `avrdude_message2()` could be a function pointer that can be overwritten by the user of that library. @dl8dtl?

From @dl8dtl 
> I think I could live with that.

mcuee:
## Fix `libavrdude` shared library (DLL) build on Windows — revised

Fixes #1414

### Background

`libavrdude` calls `avrdude_message2()` and references global variables
(`verbose`, `progname`, `ovsigck`, `partdesc`, `pgmid`, `quell_progress`,
`cx`) that are currently defined in `main.c` (the executable).  Windows DLLs
require all symbols to be resolved at library link time, so
`BUILD_SHARED_LIBS=ON` fails with hundreds of `undefined reference` errors on
both MinGW and MSVC.  On Linux/macOS this only "works" via
`-Wl,-undefined,dynamic_lookup`, which is a workaround, not a real fix.

### Design (agreed in PR #1874 by stefanrueger and dl8dtl)

> *"avrdude_message2() could be a function pointer that can be overwritten by
> the user of that library."* — stefanrueger
>
> *"I think I could live with that."* — dl8dtl

`avrdude_message2()` becomes a thin dispatcher in `libavrdude` that calls
through a function pointer registered by the application at startup via
`avrdude_set_message_handler()`.  Each application supplies its own handler:

- **avrdude CLI**: registers `cli_message_handler()` in `main.c` — the
  original `avrdude_message2()` body, now with a `va_list` parameter instead
  of `...` (full BOL tracking, progname prefix, verbosity gating, etc.)
- **A GUI**: registers a handler writing to a log widget instead of stdout
- **SWIG/Python binding**: registers a handler routing to Python logging

A minimal built-in fallback (plain `vfprintf` to `fp`) is used when no handler
has been registered, so `libavrdude` is usable standalone without an explicit
registration call.

The global state variables (`verbose`, `progname`, `cx`, etc.) are moved into
the same new file so `libavrdude` owns all the symbols it references.

### Why the previous attempt (PR #2126 v1) failed

| Failure | Cause | Fix in this revision |
|---|---|---|
| SWIG/Python (Linux + macOS) | `avrdude_msg.c` was in `${SOURCES}`, so `swig_avrdude` compiled `avrdude_message2()` in — duplicate definition | `avrdude_message2()` now stays in `libavrdude`; SWIG consumers supply their own handler via `avrdude_set_message_handler()` |
| MSVC x86 | `LIBAVRDUDE_API` was applied to `avrdude_message2()` which confused MSVC about a symbol it expected the app to supply | `avrdude_message2()` is now defined in the library; no special annotation needed |
| MinGW | Same root cause as MSVC | Same fix |

### Files changed

| File | Change |
|---|---|
| `src/avrdude_globals.c` | **New** — globals + `avrdude_message2()` dispatcher + fallback handler + `set/get_message_handler()` |
| `src/main.c` | Remove 7 global defs; rename handler body to `cli_message_handler(va_list)`; register it in `main()` |
| `src/CMakeLists.txt` | Add `avrdude_globals.c` to `SOURCES`; add `LIBAVRDUDE_BUILD_DLL` for Windows shared builds |
| `src/avrdude.h` | Add `LIBAVRDUDE_API` macro; annotate variable `extern` declarations |
| `src/libavrdude.h` | Add `<stdarg.h>`; update `avrdude_message2()` comment; add `avrdude_message_handler_t` typedef and two new function declarations |

Root `CMakeLists.txt` is not changed.

---

### `src/avrdude_globals.c` (new file — full content)

See attached `avrdude_globals.c`.

---

### `src/main.c` diff

See attached `main_c.diff`.

Key points:
- `avrdude_message_type()` static helper: unchanged, stays in `main.c`
- `avrdude_message2()` body: renamed to `static int cli_message_handler()`
  with `va_list ap` replacing `...`; two `va_start`/`va_end` pairs become
  one `va_copy`/`va_end` for sizing + direct use of `ap` for final print
- `report_progress()` and the global `MSG2_FLUSH` calls are moved to the
  dispatcher in `avrdude_globals.c` so they fire regardless of handler
- `libavrdude_context *cx` and the five `int`/`const char *` globals removed
- `char *progname = "avrdude"` removed
- In `main()`, add immediately after `init_cx(NULL);`:
  ```c
  avrdude_set_message_handler(cli_message_handler);
  ```

---

### `src/CMakeLists.txt`, `src/avrdude.h`, `src/libavrdude.h` diffs

See attached `other_files.diff`.

mcuee:
```
--- a/src/main.c
+++ b/src/main.c
@@ -51,11 +51,6 @@ char *progname = "avrdude";
-char *progname = "avrdude";
-
-static const char *avrdude_message_type(int msglvl) {
-  switch(msglvl) {
-  case MSG_EXT_ERROR:
-    return "OS error";
-  case MSG_ERROR:
-    return "error";
-  case MSG_WARNING:
-    return "warning";
-  case MSG_INFO:
-    return "info";
-  case MSG_NOTICE:
-    return "notice";
-  case MSG_NOTICE2:
-    return "notice2";
-  case MSG_DEBUG:
-    return "debug";
-  case MSG_TRACE:
-    return "trace";
-  case MSG_TRACE2:
-    return "trace2";
-  default:
-    return "unknown msglvl";
-  }
-}
-
-/*
- * Core messaging routine for msg_xyz(), [pli]msg_xyz() and term_out()
- * See #define lines in avrdude.h of how it is normally called
- *
- * Named that way as there used to be a now gone different avrdude_message()
- */
-int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
-  int msgmode, int msglvl, const char *format, ...) {
-
-  int rc = 0;
-  va_list ap;
-
-  static struct {               // Memorise whether last print ended at beginning of line
-    FILE *fp;
-    int bol;                    // Are we at the beginning of a line for this fp stream?
-  } bols[5 + 1];                // Cater for up to 5 different FILE pointers plus one catch-all
-
-  size_t bi = 0;
-
-  for(bi = 0; bi < sizeof bols/sizeof *bols - 1; bi++) {      // Note the -1, so bi is valid after loop
-    if(!bols[bi].fp) {          // First free space
-      bols[bi].fp = fp;         // Insert fp in first free space
-      bols[bi].bol = 1;         // Assume beginning of line on first use
-    }
-    if(bols[bi].fp == fp)
-      break;
-  }
-
-  if(msglvl <= MSG_ERROR)       // Serious error? Free progress bars (if any)
-    report_progress(1, -1, NULL);
-
-  if(msgmode & MSG2_FLUSH) {
-    fflush(stdout);
-    fflush(stderr);
-  }
-  // Reduce effective verbosity level by number of -q above one when printing to stderr
-  if((fp == stderr? verblevel: verbose) >= msglvl) {
-    if(msgmode & MSG2_LEFT_MARGIN && !bols[bi].bol) {
-      fprintf(fp, "\n");
-      bols[bi].bol = 1;
-    }
-    // Keep vertical tab at start of format string as conditional new line
-    if(*format == '\v') {
-      format++;
-      if(!bols[bi].bol) {
-        fprintf(fp, "\n");
-        bols[bi].bol = 1;
-      }
-    }
-
-    if(msgmode & (MSG2_PROGNAME | MSG2_TYPE)) {
-      if(msgmode & MSG2_PROGNAME) {
-        fprintf(fp, "%s", progname);
-        bols[bi].bol = 0;
-      }
-      if(msgmode & MSG2_TYPE) {
-        const char *mt = avrdude_message_type(msglvl);
-
-        if(bols[bi].bol)
-          fprintf(fp, "%c%s", msgmode & (MSG2_UCFIRST)? toupper(*mt & 0xff): *mt, mt + 1);
-        else
-          fprintf(fp, " %s", mt);
-        bols[bi].bol = 0;
-      }
-      if(verblevel >= MSG_NOTICE2) {
-        const char *bfname = strrchr(file, '/');        // Only print basename
-
-#if defined (WIN32)
-        if(!bfname)
-          bfname = strrchr(file, '\\');
-#endif
-
-        bfname = bfname? bfname + 1: file;
-        if(msgmode & MSG2_FUNCTION)
-          fprintf(fp, " %s()", func);
-        if(msgmode & MSG2_FILELINE)
-          fprintf(fp, " %s %d", bfname, lno);
-      }
-      fprintf(fp, ": ");
-    } else if(msgmode & MSG2_INDENT1) {
-      fprintf(fp, "%*s", (int) strlen(progname) + 1, "");
-      bols[bi].bol = 0;
-    } else if(msgmode & MSG2_INDENT2) {
-      fprintf(fp, "%*s", (int) strlen(progname) + 2, "");
-      bols[bi].bol = 0;
-    }
-    // Figure out whether this print will leave us at beginning of line
-
-    // Determine required size first
-    va_start(ap, format);
-    rc = vsnprintf(NULL, 0, format, ap);
-    va_end(ap);
-
-    if(rc < 0)                  // Some errror?
-      return 0;
-
-    rc++;                       // Accommodate terminating nul
-    char *p = mmt_malloc(rc);
-
-    va_start(ap, format);
-    rc = vsnprintf(p, rc, format, ap);
-    va_end(ap);
-
-    if(rc < 0) {
-      mmt_free(p);
-      return 0;
-    }
-
-    if(*p) {                    // Finally: print!
-      if(bols[bi].bol && (msgmode & MSG2_UCFIRST))
-        fprintf(fp, "%c%s", toupper(*p & 0xff), p + 1);
-      else
-        fprintf(fp, "%s", p);
-      bols[bi].bol = p[strlen(p) - 1] == '\n';
-    }
-    mmt_free(p);
-  }
-
-  if(msgmode & MSG2_FLUSH)
-    fflush(fp);
-
-  return rc;
-}
+static const char *avrdude_message_type(int msglvl) {
+  switch(msglvl) {
+  case MSG_EXT_ERROR: return "OS error";
+  case MSG_ERROR:     return "error";
+  case MSG_WARNING:   return "warning";
+  case MSG_INFO:      return "info";
+  case MSG_NOTICE:    return "notice";
+  case MSG_NOTICE2:   return "notice2";
+  case MSG_DEBUG:     return "debug";
+  case MSG_TRACE:     return "trace";
+  case MSG_TRACE2:    return "trace2";
+  default:            return "unknown msglvl";
+  }
+}
+
+/*
+ * Full-featured CLI message handler registered with libavrdude at startup.
+ * This is the original avrdude_message2() body from main.c, adapted to
+ * receive a va_list instead of variadic arguments so it can be called
+ * through the avrdude_message_handler_t callback registered via
+ * avrdude_set_message_handler().
+ *
+ * A GUI or other application would register a different handler here,
+ * e.g. one that writes to a log widget instead of printing to the console.
+ * See avrdude_set_message_handler() in libavrdude.h.
+ */
+static int cli_message_handler(FILE *fp, int lno, const char *file,
+  const char *func, int msgmode, int msglvl, const char *format, va_list ap) {
+
+  int rc = 0;
+
+  static struct {               // Memorise whether last print ended at beginning of line
+    FILE *fp;
+    int bol;                    // Are we at the beginning of a line for this fp stream?
+  } bols[5 + 1];                // Cater for up to 5 different FILE pointers plus one catch-all
+
+  size_t bi = 0;
+
+  for(bi = 0; bi < sizeof bols/sizeof *bols - 1; bi++) {      // Note the -1, so bi is valid after loop
+    if(!bols[bi].fp) {          // First free space
+      bols[bi].fp = fp;         // Insert fp in first free space
+      bols[bi].bol = 1;         // Assume beginning of line on first use
+    }
+    if(bols[bi].fp == fp)
+      break;
+  }
+
+  // Reduce effective verbosity level by number of -q above one when printing to stderr
+  if((fp == stderr? verblevel: verbose) >= msglvl) {
+    if(msgmode & MSG2_LEFT_MARGIN && !bols[bi].bol) {
+      fprintf(fp, "\n");
+      bols[bi].bol = 1;
+    }
+    // Keep vertical tab at start of format string as conditional new line
+    if(*format == '\v') {
+      format++;
+      if(!bols[bi].bol) {
+        fprintf(fp, "\n");
+        bols[bi].bol = 1;
+      }
+    }
+
+    if(msgmode & (MSG2_PROGNAME | MSG2_TYPE)) {
+      if(msgmode & MSG2_PROGNAME) {
+        fprintf(fp, "%s", progname);
+        bols[bi].bol = 0;
+      }
+      if(msgmode & MSG2_TYPE) {
+        const char *mt = avrdude_message_type(msglvl);
+
+        if(bols[bi].bol)
+          fprintf(fp, "%c%s", msgmode & (MSG2_UCFIRST)? toupper(*mt & 0xff): *mt, mt + 1);
+        else
+          fprintf(fp, " %s", mt);
+        bols[bi].bol = 0;
+      }
+      if(verblevel >= MSG_NOTICE2) {
+        const char *bfname = strrchr(file, '/');        // Only print basename
+
+#if defined (WIN32)
+        if(!bfname)
+          bfname = strrchr(file, '\\');
+#endif
+
+        bfname = bfname? bfname + 1: file;
+        if(msgmode & MSG2_FUNCTION)
+          fprintf(fp, " %s()", func);
+        if(msgmode & MSG2_FILELINE)
+          fprintf(fp, " %s %d", bfname, lno);
+      }
+      fprintf(fp, ": ");
+    } else if(msgmode & MSG2_INDENT1) {
+      fprintf(fp, "%*s", (int) strlen(progname) + 1, "");
+      bols[bi].bol = 0;
+    } else if(msgmode & MSG2_INDENT2) {
+      fprintf(fp, "%*s", (int) strlen(progname) + 2, "");
+      bols[bi].bol = 0;
+    }
+
+    // Determine required size first, using a copy of ap
+    va_list ap2;
+    va_copy(ap2, ap);
+    rc = vsnprintf(NULL, 0, format, ap2);
+    va_end(ap2);
+
+    if(rc < 0)                  // Some error?
+      return 0;
+
+    rc++;                       // Accommodate terminating nul
+    char *p = mmt_malloc(rc);
+
+    rc = vsnprintf(p, rc, format, ap);
+
+    if(rc < 0) {
+      mmt_free(p);
+      return 0;
+    }
+
+    if(*p) {                    // Finally: print!
+      if(bols[bi].bol && (msgmode & MSG2_UCFIRST))
+        fprintf(fp, "%c%s", toupper(*p & 0xff), p + 1);
+      else
+        fprintf(fp, "%s", p);
+      bols[bi].bol = p[strlen(p) - 1] == '\n';
+    }
+    mmt_free(p);
+  }
+
+  if(msgmode & MSG2_FLUSH)
+    fflush(fp);
+
+  return rc;
+}

 struct list_walk_cookie {
   FILE *f;
   const char *prefix;
 };

-libavrdude_context *cx;         // Context pointer, eventually the only global variable
-
 static LISTID updates = NULL;

@@ Global options block in main.c @@
-// Global options
-int verbose;                    // Verbose output
-int quell_progress;             // Quell progress report and un-verbose output
-int ovsigck;                    // 1 = override sig check, 0 = don't
-const char *partdesc;           // Part -p string
-const char *pgmid;              // Programmer -c string

@@ In main(), right after init_cx(NULL); @@
   init_cx(NULL);
+  avrdude_set_message_handler(cli_message_handler);
```

mcuee:
```
===========================================================================
src/CMakeLists.txt
===========================================================================
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ set(SOURCES @@
 set(SOURCES
     ${CMAKE_CURRENT_BINARY_DIR}/ac_cfg.h
     arduino.h
     arduino.c
+    avrdude_globals.c
     avr.c

@@ after set_target_properties(libavrdude ...) @@
 set_target_properties(libavrdude PROPERTIES
     PREFIX ""
     PUBLIC_HEADER "libavrdude.h;libavrdude-avrintel.h"
     VERSION 4.0.0
     SOVERSION 4
     )
+
+# On Windows shared builds, export the symbols defined in avrdude_globals.c
+# (state variables, avrdude_message2 dispatcher, set/get_message_handler).
+# These were formerly in main.c, making libavrdude unlink-able as a DLL.
+if(BUILD_SHARED_LIBS AND WIN32)
+    target_compile_definitions(libavrdude PRIVATE LIBAVRDUDE_BUILD_DLL)
+endif()

===========================================================================
src/avrdude.h
===========================================================================
--- a/src/avrdude.h
+++ b/src/avrdude.h
@@ after #include <stdlib.h> @@
+// DLL import/export annotation for Windows shared library builds.
+// Applied to global state variables and new message handler API exported
+// from libavrdude.  avrdude_message2() does not need explicit annotation
+// because it is defined in the same libavrdude translation unit.
+#if defined(_WIN32) && defined(LIBAVRDUDE_BUILD_DLL)
+#  define LIBAVRDUDE_API __declspec(dllexport)
+#elif defined(_WIN32)
+#  define LIBAVRDUDE_API __declspec(dllimport)
+#else
+#  if defined(__GNUC__) && __GNUC__ >= 4
+#    define LIBAVRDUDE_API __attribute__((visibility("default")))
+#  else
+#    define LIBAVRDUDE_API
+#  endif
+#endif
+
@@ extern variable declarations @@
-extern char *progname;          // Name of program, for messages
-extern int ovsigck;             // Override signature check (-F)
-extern int verbose;             // Verbosity level (-v, -vv, ...)
-extern int quell_progress;      // Quell progress report -q, reduce effective verbosity level (-qq, -qqq)
-extern const char *partdesc;    // Part -p string
-extern const char *pgmid;       // Programmer -c string
+LIBAVRDUDE_API extern char       *progname;       // Name of program, for messages
+LIBAVRDUDE_API extern int         ovsigck;        // Override signature check (-F)
+LIBAVRDUDE_API extern int         verbose;        // Verbosity level (-v, -vv, ...)
+LIBAVRDUDE_API extern int         quell_progress; // Quell progress report -q, reduce effective verbosity level (-qq, -qqq)
+LIBAVRDUDE_API extern const char *partdesc;       // Part -p string
+LIBAVRDUDE_API extern const char *pgmid;          // Programmer -c string

===========================================================================
src/libavrdude.h
===========================================================================
--- a/src/libavrdude.h
+++ b/src/libavrdude.h
@@ #include <stdio.h> block at top @@
 #include <stdio.h>
 #include <limits.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
+#include <stdarg.h>

@@ Message system section @@
 // Message system
 
-// This functions is supposed to be supplied by the application
+// avrdude_message2() is defined in libavrdude (avrdude_globals.c) and
+// dispatches to a handler registered by the application via
+// avrdude_set_message_handler().  Applications supply a handler suited to
+// their output model: the CLI registers a console handler in main.c, a GUI
+// registers a widget handler, the SWIG/Python binding registers a Python
+// logging handler, etc.  If no handler is registered, a built-in fallback
+// (plain vfprintf to fp) is used so libavrdude is usable standalone.
+// See avrdude_set_message_handler() below.
 int avrdude_message2(FILE *fp, int lno, const char *file, const char *func,
   int msgmode, int msglvl, const char *format, ...);
+
+// Pluggable message handler type.
+// Matches the signature of avrdude_message2() but takes a va_list instead
+// of variadic arguments, so it can be called from the dispatcher.
+typedef int (*avrdude_message_handler_t)(FILE *fp, int lno, const char *file,
+  const char *func, int msgmode, int msglvl, const char *format, va_list ap);
+
+// Register an application-supplied message handler with libavrdude.
+// Call once at startup, before any library functions are used.
+// Pass NULL to revert to the built-in fallback handler.
+void avrdude_set_message_handler(avrdude_message_handler_t handler);
+
+// Retrieve the currently registered handler (or NULL if using fallback).
+avrdude_message_handler_t avrdude_get_message_handler(void);
```

mcuee:
Let me close this draft PR and I will test and create another PR.

PR #2101, opened 2026-04-18
Update artifact upload name and path in arduino_packing_release workflow

mcuee:
Fix https://github.com/avrdudes/avrdude/issues/2039.

Issue #2039, opened 2025-06-24
github avrdude_packing_release action failed


mcuee:
Somehow we did not have this issue previously for 8.0 release.

I have no ideas why this failed.

https://github.com/avrdudes/avrdude/actions/runs/15824523096/job/44645916386
```
Run actions/upload-artifact@v4
/usr/bin/docker exec  bee507e0733f4a79dd7690dd2a4c4fe7898de9[1](https://github.com/avrdudes/avrdude/actions/runs/15824523096/job/44646408235#step:7:1)44c6d60d7a7d9f69e4e6f282c sh -c "cat /etc/*release | grep ^ID"
With the provided path, there will be 4 files uploaded
Artifact name is valid!
Root directory input is valid!
Error: Failed to CreateArtifact: Received non-retryable error: Failed request: (409) Conflict: an artifact with this name already exists on the workflow run
```

mcuee:
@umbynos 

Sorry to disturb you. Last time you helped on this github action. Just wondering if you can take a look to see what is the potential problem. Thanks.

mcuee:
It seems to me that somehow each run create the same `dist.zip` file and that may be the issue.

On the other hand, the other routine github action seems to work fine. It creates 6 `dist_xxxOS_xxxarch.zip` files.
https://github.com/avrdudes/avrdude/actions/runs/15824598803

![Image](https://github.com/user-attachments/assets/32d7e996-0932-420d-b238-86fd8e714856)

umbynos:
Hey @mcuee, a little late but the problem should be that the `upload-artifact` uses an old version of the action. To upgrade to a newer version I recommend using this guide: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md

stefanrueger:
Is this still an active issue @mcuee?

mcuee:
> Is this still an active issue [@mcuee](https://github.com/mcuee)?

Yes. It will be good that we to fix this before 8.2 release.

mcuee:
PR #2101 has been raised.

Issue #1316, opened 2023-03-11
Abstract out transport -- libusb and hidapi


mcuee:
I am not so sure how difficult this is. I can see that the transport mechanisms are kind of abstracted out to a certain extent but not that clear. It would be good to improve the situation if possible.

Maybe serial port and parallel port are better asbstracted out but I did not look so deep into them. Parallel port is kind of dead anyway.

Example:
1) HIDAPI: the following code is there but it does not seem to be widely used by the programmers, rather the programmers are still using <hidapi/hidapi.h> directly.
https://github.com/avrdudes/avrdude/blob/main/src/usb_hidapi.c (probably this is not sufficient)

2) libusb-0,1 API: the following code is there but I am not that so sure how widely it is used.
https://github.com/avrdudes/avrdude/blob/main/src/usb_libusb.c

And then there are quite some fall back codes. I understand the needs to support as wide platforms as possible but I guess it does add to the complexity of the codes.
1) usbasp -- try to use libusb-1.0 API, if not present then fall back to libusb-0.1 API
2) HIDAPI based programmer -- use HIDAPI first, if not present then fall back to libusb
3) PICKit 2 -- use libusb or Win32 API, it would be good to use HIDAPI instead and only fall back when HIDAPI is not present.

mcuee:
Even though this is a nice to have enhancement, but it is not essential. So I will close this one for now.

Issue #974, opened 2022-05-22
Move away from libusb-0.1 and libusb-win32


mcuee:
It will be good to move away from libusb-0.1 (or libusb-compat-0.1) from Linux, macOS, FreeBSD, etc. It is also good to move away from libusb-win32 under Windows.

Main affected file.
dfu.h
usb_libusb.c
micronucleus.c
pickti2.c [better move to hidapi and not libusb-1.0]
usbasp.c
usbtiny.c

Depending on the approaches used, it could affect more files as well like jtag3.c, jtagmkII.c and stk500v2.c.

Earlier attempts for stk500v2 and related programers.
* https://github.com/avrdudes/avrdude/issues/704#issuecomment-1135022810
> Thus, the code needs a bit of cleanup at first, and thorough testing against all affected devices (AVRISPmkII, JTAGICEmkII, JTAGICE3, AtmelICE, STK600, AVR Dragon – did I forget any?) afterwards, as it changes the underlying transport library for all these devices.

mcuee:
Right now avrdude under Windows is using avrdude-libwinusb which is good to support both libusb0.sys and WinUSB. It has a limitation though -- some issues with USB Composite Device support.

Ref: https://github.com/avrdudes/avrdude/issues/968#issuecomment-1132031771

mcuee:
It seems okay to use libusb0.sys and libusb-1.0.26 and libftdi-1.0 together with avrdude under Windows. I tested with Amontec JTAGKey-2.
https://github.com/avrdudes/avrdude/issues/975#issuecomment-1149539658

I will test out libusb0.sys and libusb-1.0.26 with usbasp later.

mcuee:
usbasp test with official Windows binary which is using avrdude-libwinusb and not libusb-1.0. It supports libusb0.sys and WinUSB. It actually supports libusbk.sys as well (since libusbk.sys supports most of the driver ioctl codes from libusb0.sys)

Run log using libusbk.sys.

```
PS C:\work\avr\avrdude_test\avrdude-v7.0-windows-x64> .\avrdude.exe -c usbasp -p m328p -U flash:v:.\Blink.ino.with_bootl
oader.mega_mod.hex:i -v

avrdude.exe: Version 7.0
             Copyright (c) Brian Dean, http://www.bdmicro.com/
             Copyright (c) Joerg Wunsch

             System wide configuration file is "C:/work/avr/avrdude_test/avrdude-v7.0-windows-x64/avrdude.conf"

[...]

PR #2128, opened 2026-06-17
Fix subtle usb issues including inconsistent parsing

### LINKED ISSUE #1221: PICKit 4, SNAP, Atmel ICE and Power Debugger only work with hidapi but not libusb (low priority)
From here:
* https://github.com/avrdudes/avrdude/issues/1038#issuecomment-1354195120

For this test, I build avrdude without hidapi support. And I make sure the hid drver is already detached and then libusb is used.

hidapi version works fine (with hidapi-libusb backend). So there is a bug here in git main.

```
mcuee@UbuntuSwift3:~/build/avr/avrdude_main$ cat build_nohidapi.sh 
#!/bin/sh
cmake -D CMAKE_BUILD_TYPE=RelWithDebInfo -D HAVE_LIBHIDAPI=OFF  -B build_linux
cmake --build build_linux

mcuee@UbuntuSwift3:~/build/avr/avrdude_bin$ ldd ./avrdude_git_nohidapi 
	linux-vdso.so.1 (0x00007ffc52b2f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fa6169f6000)
	libelf.so.1 => /lib/x86_64-linux-gnu/libelf.so.1 (0x00007fa6169da000)
	libusb-0.1.so.4 => /usr/local/lib/libusb-0.1.so.4 (0x00007fa6169d2000)
	libusb-1.0.so.0 => /usr/local/lib/libusb-1.0.so.0 (0x00007fa6169b2000)
	libftdi.so.1 => /lib/x86_64-linux-gnu/libftdi.so.1 (0x00007fa6169a8000)
	libftdi1.so.2 => /usr/local/lib/libftdi1.so.2 (0x00007fa616995000)
	libreadline.so.8 => /lib/x86_64-linux-gnu/libreadline.so.8 (0x00007fa616945000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fa616753000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fa616c5f000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007fa616737000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007fa616714000)
	libudev.so.1 => /lib/x86_64-linux-gnu/libudev.so.1 (0x00007fa6166e7000)
	libtinfo.so.6 => /lib/x86_64-linux-gnu/libtinfo.so.6 (0x00007fa6166b5000)

mcuee@UbuntuSwift3:~/build/avr/avrdude_bin$ ./avrdude_git_nohidapi -c pickit4_updi -p m4808 -v

avrdude_git_nohidapi: Version 7.0-20221214 (159d126)
                      Copyright the AVRDUDE authors;
                      see https://github.com/avrdudes/avrdude/blob/main/AUTHORS

                      System wide configuration file is /home/mcuee/build/avr/avrdude_bin/avrdude.conf
                      User configuration file is /home/mcuee/.avrduderc

                      Using Programmer              : pickit4_updi
avrdude_git_nohidapi: usbdev_open(): found MPLAB PICkit 4 CMSIS-DAP, serno: BUR2009xxxxx
avrdude_git_nohidapi usbdev_open() warning: unable to set configuration 1: Device or resource busy
avrdude_git_nohidapi: found CMSIS-DAP compliant device, using EDBG protocol
avrdude_git_nohidapi usbdev_send() error: wrote -5 out of 912 bytes, err = Input/output error
avrdude_git_nohidapi jtag3_edbg_prepare() error: unable to send command to serial port
avrdude_git_nohidapi main() error: unable to open programmer pickit4_updi on port usb

avrdude_git_nohidapi done.  Thank you.
```

### LINKED ISSUE #2127: Misleading messaging with `-P usb:<vid>:<pid>`
The syntax `-P usb:<vid>:<pid>` allows overriding the `avrdude.conf` defaults for (most) programmers. For example

```
$ lsusb | grep Snap
Bus 001 Device 027: ID 03eb:2180 Atmel Corp. MPLAB Snap ICD CMSIS-DAP

$ avrdude -p x32a4u -c snap_pdi -P usb:03eb:2180

Avrdude done.  Thank you.
```

So far, so expected. But if I specify a wrong pid all hell breaks loose.

```
$ avrdude -p x32a4u -c snap_pdi -P usb:03eb:2181
Error: no matching device found
Error: no USB HID devices found
Error: no USB HID devices found
Error: no device found matching VID 0x04d8 and PID list: 0x2180, 0x217f, 0x2181 with SN 03eb:2181
Error: unable to open port usb:03eb:2181 for programmer snap_pdi

Avrdude done.  Thank you.
```

There are two things that a user might object to:
  - Neither `avrdude.conf` nor the command line mention vid **0x04d8**, *but* the no-device-found message does! How come?
  - The same message claims the user wanted a match with serial number 03eb:2181, so that's wrong, too

And of course, the four intermediate error messages before AVRDUDE finally gives up are a bit much but I'll give that to AVRDUDE for its diligence.

I'll review the `-P usb:<vid>:<pid>` usage and parsing.

stefanrueger:
Fixes #2127 and #1221 

This is also an attempt to standardise `-P usb[:<vid>:<pid>][:<serno>]` parsing to a better degree for those programmers that allow  vid, pid and serno specification. (There are some that use the `-P usb[:<busdir>:<devicefile]` logic; these won't be affected).

Novelty: for the first time I've used a code fragment suggested by an AI program following @mcuee's enthusiastic use:
https://github.com/stefanrueger/avrdude/blob/68b970e2456083566cb8be1d2d935780b052bcad/src/usb_libusb.c#L207-L208
I might delete that attribution to Claude, as the used code fragment is only a small sub-expression, and I wonder how much that differs from using a tip that Ms Google had given. It's certainly been some work trying to make sense of the verbose AI review.

What is needed now is good testing, a) whether #2127 and #1221 are really gone and b) whether there are regressions. 

@MX682X This PR replaces pickit5.c's `-P usb[:<vid>:<pid>]` parsing with the more general (and stricter) parsing through `str_set_vid_pid_serno()`. Could you please test/try/review the changed parsing? How attached are you to the variant `-P usb::<pid>` where vid defaults to Microchip's vid? This element turns out to be a slight pain for the attempt to standardise `-P usb:...` parsing across different programmers in AVRDUDE, as it's not clear to me whether defaulting to Microchip's vid makes generally sense (I suspect, not). Dropping that otherwise neat `-P usb::<pid>` syntax would also make documentation easier.

I intend to use `-P usb:...` parsing for more programmers, eg, avrftdi, ch341a, pickit2, teensy, ... where `-P ...` is currently ignored.That's why this PR is a draft for now (though one that can be tested).

### DIFF
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -650,93 +650,47 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
   union pinfo pinfo;
   LNODEID usbpid;
   int rv = LIBAVRDUDE_GENERAL_FAILURE;
-  size_t serial_num_len = 0;
 
 #if !defined(HAVE_LIBUSB)
   pmsg_error("need to be compiled with USB or HIDAPI support\n");
   return LIBAVRDUDE_GENERAL_FAILURE;
 #endif
 
-  if(!str_casestarts(port, "usb:") && !str_caseeq(port, "usb")) {
-    pmsg_error("invalid -P %s; drop -P option or else use -P usb:<vid>:<pid> or -P usb:<serialno>\n", port);
+  unsigned short new_vid = 0, new_pid = 0;
+  char serno[64] = { 0 };
+  int numids = str_set_vid_pid_serno(port, &new_vid, &new_pid, USB_VENDOR_MICROCHIP, serno, sizeof serno);
+  if(numids < 0) {
+    pmsg_error("invalid -P %s; drop -P option or else use -P usb[:[<vid>]:<pid>][:<serno>]\n", port);
     return LIBAVRDUDE_GENERAL_FAILURE;
   }
-  unsigned int new_vid = 0, new_pid = 0, setids = 0;
-  const char *vidp, *pidp;
-
-  /*
-   * The syntax for usb devices is defined as:
-   *
-   * -P usb:vid:pid
-   * -P usb::pid
-   * -P usb:serialnumber
-   * -P usb
-   *
-   * First we check if we have two colons.
-   * Then check the filed between the two colons is empty
-   * Parse VID as hex number
-   * If it is empty, assume Microchip VID
-   * The PID is handled similary but can not be empty
-   *
-   * If there are fewer than two colons nothing is changed
-   */
-
-  vidp = strchr(port, ':');
-  if(vidp != NULL) {
-    vidp += 1;
-    pidp = strchr(vidp, ':');
-    if(pidp != NULL) {
-      setids = 1;
-      if(vidp != pidp) {        // User specified an VID
-        // First: Handle VID input
-        if(sscanf(vidp, "%x", &new_vid) != 1) {
-          pmsg_error("failed to parse -P VID input %s: expected hexadecimal number\n", vidp);
-          return LIBAVRDUDE_GENERAL_FAILURE;
-        }
-      } else {                  // VID space empty: default to Microchip
-        new_vid = USB_VENDOR_MICROCHIP;
-      }
-
-      // Now handle PID input
-      if(sscanf(pidp + 1, "%x", &new_pid) != 1) {
-        pmsg_error("failed to parse -P PID input %s: expected hexadecimal number\n", pidp+1);
-        return LIBAVRDUDE_GENERAL_FAILURE;
-      }
-
-      pmsg_notice("overwriting VID:PID to %04x:%04x\n", new_vid, new_pid);
-      port = "usb";             // Overwrite the string to avoid confusing the libusb
-    } else {                    // pidp == NULL means vidp points to serial number
-      serial_num_len = strlen(vidp);
-    }
-  }                             // vidp == NULL means dropped -P option or -P usb
 
   // If the config entry did not specify a USB PID, insert the default one
   if(lfirst(pgm->usbpid) == NULL)
     ladd(pgm->usbpid, (void *) USB_DEVICE_PICKIT5);
 
-  pinfo.usbinfo.vid = pgm->usbvid? pgm->usbvid: USB_VENDOR_MICROCHIP;
-
   // PICkit 5 does not have support for HID, so no need to support it
   serdev = &usb_serdev;
-  if(setids) {                  // In case a specific VID/PID was specified
+  pinfo.usbinfo.vid = pgm->usbvid? pgm->usbvid: USB_VENDOR_MICROCHIP;
+  pinfo.usbinfo.flags = PINFO_FL_SILENT;
+  if(numids >= 2) {             // Valid VID:PID was specified in -P port
+    if(pinfo.usbinfo.vid != new_vid || pinfo.usbinfo.pid != new_pid)
+      pmsg_notice("overwriting VID:PID to %04x:%04x\n", new_vid, new_pid);
     pinfo.usbinfo.vid = new_vid;
     pinfo.usbinfo.pid = new_pid;
-    pinfo.usbinfo.flags = PINFO_FL_SILENT;
     pgm->fd.usb.max_xfer = USB_PK5_MAX_XFER;
     pgm->fd.usb.rep = USB_PK5_CMD_READ_EP;  // Command read
     pgm->fd.usb.wep = USB_PK5_CMD_WRITE_EP; // Command write
     pgm->fd.usb.eep = 0x00;
+    port = str_sprintf("usb:%s", serno); // Avoids passing on usb::<pid>:<serno>
     pgm->port = port;
     rv = serial_open(port, pinfo, &pgm->fd);
   } else {                      // Otherwise walk the list of config file PIDs
     for(usbpid = lfirst(pgm->usbpid); rv < 0 && usbpid != NULL; usbpid = lnext(usbpid)) {
-      pinfo.usbinfo.flags = PINFO_FL_SILENT;
       pinfo.usbinfo.pid = *(int *) ldata(usbpid);
       pgm->fd.usb.max_xfer = USB_PK5_MAX_XFER;
       pgm->fd.usb.rep = USB_PK5_CMD_READ_EP;  // Command read
       pgm->fd.usb.wep = USB_PK5_CMD_WRITE_EP; // Command write
       pgm->fd.usb.eep = 0x00;
-
       pgm->port = port;
       rv = serial_open(port, pinfo, &pgm->fd);
     }
@@ -808,8 +762,8 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
       serial_close(&pgm->fd);
       return LIBAVRDUDE_EXIT_FAIL;
     }
-    if(serial_num_len) {
-      pmsg_error("no device found matching the specified serial number %s", vidp);
+    if(*serno) {
+      pmsg_error("no device found matching the specified serial number %s", serno);
       return LIBAVRDUDE_GENERAL_FAILURE;
     }
 
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -681,7 +681,6 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
     pgm->fd.usb.rep = USB_PK5_CMD_READ_EP;  // Command read
     pgm->fd.usb.wep = USB_PK5_CMD_WRITE_EP; // Command write
     pgm->fd.usb.eep = 0x00;
-    port = str_sprintf("usb:%s", serno); // Avoids passing on usb::<pid>:<serno>
     pgm->port = port;
     rv = serial_open(port, pinfo, &pgm->fd);
   } else {                      // Otherwise walk the list of config file PIDs
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -658,9 +658,9 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
 
   unsigned short new_vid = 0, new_pid = 0;
   char serno[64] = { 0 };
-  int numids = str_set_vid_pid_serno(port, &new_vid, &new_pid, USB_VENDOR_MICROCHIP, serno, sizeof serno);
+  int numids = str_set_vid_pid_serno(port, &new_vid, &new_pid, serno, sizeof serno);
   if(numids < 0) {
-    pmsg_error("invalid -P %s; drop -P option or else use -P usb[:[<vid>]:<pid>][:<serno>]\n", port);
+    pmsg_error("invalid -P %s; drop -P option or use -P usb[:<vid>:<pid>][:<serno>]\n", port);
     return LIBAVRDUDE_GENERAL_FAILURE;
   }
 
--- a/src/pickit5.c
+++ b/src/pickit5.c
@@ -661,7 +661,7 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
   int numids = str_set_vid_pid_serno(port, &new_vid, &new_pid, serno, sizeof serno);
   if(numids < 0) {
     pmsg_error("invalid -P %s; drop -P option or use -P usb[:<vid>:<pid>][:<serno>]\n", port);
-    return LIBAVRDUDE_GENERAL_FAILURE;
+    return LIBAVRDUDE_EXIT_FAIL;
   }
 
   // If the config entry did not specify a USB PID, insert the default one
--- a/src/strutil.c
+++ b/src/strutil.c
@@ -1231,6 +1231,94 @@ finished:
 }
 
 /*
+ * The syntax of the -P port string is -P usb[[:[<vid>]:<pid>]:<serno>]
+ *
+ * Set *vidp and *pidp if <vid> and <pid> are hex strings in [0, 0xffff];
+ * an empty <vid> string is allowed if defvid is given as non-negative
+ * number, in which case *vidp is set to defvid. In case there is no valid
+ * <vid>/<pid> pair, the full string after usb: is treated as serial
+ * number, as is any trailing string a after a valid usb:<vid>:<pid>:
+ * sequence. Serial numbers, if detected in the port string, are stripped
+ * of colon characters, while a maximum of n-1 serial number characters
+ * are stored in the space pointed to by serno, followed by a terminating
+ * nul character.
+ *
+ * Returns
+ *  -1 on failed parsing
+ *   0 if no paramters could be set
+ *   1 if only serno could be set
+ *   2 if only vid and pid could be set
+ *   3 if vid, pid and serno could be set
+ *
+ * *vidp and *pidp are unchanged for return values smaller than 2. The
+ * buffer serno may be set with the initial serial number if n turned out
+ * to be too small to hold the full serial number: in this case the return
+ * value is 0 or 2 indicating the serial number could not be completely
+ * copied over from the port string. On a failed parsing or on -P usb,
+ * serno is not set, otherwise it will contain a nul-terminated string
+ * within the n bytes size associated with serno.
+ */
+
+int str_set_vid_pid_serno(const char *port, unsigned short *vidp, unsigned short *pidp,
+  int defvid, char *serno, size_t n) {
+
+  unsigned long vid, pid;
+  const char *pidstr, *sn = port + 4;
+  char *endhex;
+  int ret = 0;
+
+  if(!str_casestarts(port, "usb"))
+    return -1;
+
+  if(port[3] != ':')            // -P usb returns 0, wrong character after -p usb returns -1
+    return port[3] == 0? 0: -1;
+
+  if(port[4] == ':' && defvid >= 0) { // -P usb::<pid> <vid> sets vid to defvid if given
+    vid = defvid;
+    pidstr = port + 5;
+  } else {                      // Scan hex vid and ensure it's terminated by :
+    errno = 0;
+    vid = strtoul(port + 4, &endhex, 16);
+    if(endhex == port + 4 || *endhex != ':' || errno || vid > 0xffff)
+      goto setserno;
+    pidstr = endhex + 1;
+  }
+
+  errno = 0;                    // Scan hex pid and ensure it's terminated by nul or :
+  pid = strtoul(pidstr, &endhex, 16);
+  if(endhex == pidstr || (*endhex != 0 && *endhex != ':') || errno || pid > 0xffff)
+    goto setserno;
+
+  if(vidp)
+    *vidp = (int) vid;
+  if(pidp)
+    *pidp = (int) pid;
+
+  if(*endhex == 0)              // No :<serno> after <pid>? Finished
+    return 2;
+
+  sn = endhex + 1;
+  ret = 2;
+setserno:
+
+  if(serno && !n)
+    return ret;
+
+  // Copy the serial number into given buffer but without colons
+  for(; *sn; sn++)
+    if(*sn != ':') {
+      if(!--n)
+        break;
+      if(serno)
+        *serno++ = *sn;
+     }
+  if(serno)
+    *serno = 0;
+
+  return ret + !*sn;            // Return 0 iff serno buffer was too short
+}
+
+/*
  * Returns the next space separated token in buf (terminating it) and places
  * start of next token into pointer pointed to by next. Keeps single or double
  * quoted strings together and changes backslash-space sequences to space
--- a/src/strutil.c
+++ b/src/strutil.c
@@ -1231,41 +1231,34 @@ finished:
 }
 
 /*
- * The syntax of the -P port string is -P usb[[:[<vid>]:<pid>]:<serno>]
+ * The syntax of the -P port string is -P usb[:<vid>:<pid>]:[<serno>]
  *
- * Set *vidp and *pidp if <vid> and <pid> are hex strings in [0, 0xffff];
- * an empty <vid> string is allowed if defvid is given as non-negative
- * number, in which case *vidp is set to defvid. In case there is no valid
- * <vid>/<pid> pair, the full string after usb: is treated as serial
- * number, as is any trailing string a after a valid usb:<vid>:<pid>:
- * sequence. Serial numbers, if detected in the port string, are stripped
- * of colon characters, while a maximum of n-1 serial number characters
- * are stored in the space pointed to by serno, followed by a terminating
- * nul character.
+ * Set *vidp and *pidp iff <vid> and <pid> are hex strings in [0, 0xffff]. If
+ * there is no valid :<vid>:<pid> pair the full string after usb: is treated as
+ * serial number, as is any trailing string after a valid usb:<vid>:<pid>:
+ * sequence. Serial numbers, if detected in port, are stripped of colons, while
+ * a maximum of n-1 serial-number characters are stored in the serno buffer
+ * followed by a terminating nul character. If the serno buffer is too small
+ * serno remains unset and the port argument is considered invalid.
  *
  * Returns
- *  -1 on failed parsing
- *   0 if no paramters could be set
- *   1 if only serno could be set
- *   2 if only vid and pid could be set
- *   3 if vid, pid and serno could be set
+ *  -1 on failed parsing (-P string does not start with usb[:], n too small)
+ *   0 if no paramters was set (-P usb)
+ *   1 if only serno was set or serno could have been set but was NULL
+ *   2 if only vid and pid were or could have been set
+ *   3 if vid, pid and serno were or could have been set
  *
- * *vidp and *pidp are unchanged for return values smaller than 2. The
- * buffer serno may be set with the initial serial number if n turned out
- * to be too small to hold the full serial number: in this case the return
- * value is 0 or 2 indicating the serial number could not be completely
- * copied over from the port string. On a failed parsing or on -P usb,
- * serno is not set, otherwise it will contain a nul-terminated string
- * within the n bytes size associated with serno.
+ * *vidp and *pidp remain unchanged for return values smaller than 2. The serno
+ * buffer remains unchanged on return values other than 1 or 3.
  */
 
 int str_set_vid_pid_serno(const char *port, unsigned short *vidp, unsigned short *pidp,
-  int defvid, char *serno, size_t n) {
+  char *serno, const size_t n) {
 
   unsigned long vid, pid;
-  const char *pidstr, *sn = port + 4;
-  char *endhex;
+  const char *sn = port + 4;
   int ret = 0;
+  char *pidstr, *endhex;
 
   if(!str_casestarts(port, "usb"))
     return -1;
@@ -1273,26 +1266,21 @@ int str_set_vid_pid_serno(const char *port, unsigned short *vidp, unsigned short
   if(port[3] != ':')            // -P usb returns 0, wrong character after -p usb returns -1
     return port[3] == 0? 0: -1;
 
-  if(port[4] == ':' && defvid >= 0) { // -P usb::<pid> <vid> sets vid to defvid if given
-    vid = defvid;
-    pidstr = port + 5;
-  } else {                      // Scan hex vid and ensure it's terminated by :
-    errno = 0;
-    vid = strtoul(port + 4, &endhex, 16);
-    if(endhex == port + 4 || *endhex != ':' || errno || vid > 0xffff)
-      goto setserno;
-    pidstr = endhex + 1;
-  }
+  errno = 0;                    // Read hex vid and assert it is followed by a colon
+  vid = strtoul(port + 4, &endhex, 16);
+  if(endhex == port + 4 || *endhex != ':' || errno || vid > 0xffff)
+    goto setserno;
 
-  errno = 0;                    // Scan hex pid and ensure it's terminated by nul or :
+  errno = 0;                    // Read hex pid and assert it is followed by nul or colon
+  pidstr = endhex + 1;
   pid = strtoul(pidstr, &endhex, 16);
   if(endhex == pidstr || (*endhex != 0 && *endhex != ':') || errno || pid > 0xffff)
     goto setserno;
 
   if(vidp)
-    *vidp = (int) vid;
+    *vidp = vid;
   if(pidp)
-    *pidp = (int) pid;
+    *pidp = pid;
 
   if(*endhex == 0)              // No :<serno> after <pid>? Finished
     return 2;
@@ -1301,21 +1289,24 @@ int str_set_vid_pid_serno(const char *port, unsigned short *vidp, unsigned short
   ret = 2;
 setserno:
 
-  if(serno && !n)
-    return ret;
+  if(!n)
+    return -1;
 
+  char *serbuf = mmt_malloc(n), *bp;
   // Copy the serial number into given buffer but without colons
-  for(; *sn; sn++)
+  for(bp = serbuf; *sn; sn++)
     if(*sn != ':') {
-      if(!--n)
+      if(bp >= serbuf + n-1)
         break;
-      if(serno)
-        *serno++ = *sn;
-     }
-  if(serno)
-    *serno = 0;
+      *bp++ = *sn;
+    }
+  *bp = 0;
 
-  return ret + !*sn;            // Return 0 iff serno buffer was too short
+  if(serno && !*sn)
+    memcpy(serno, serbuf, n);
+  mmt_free(serbuf);
+
+  return *sn? -1: ret+1;
 }
 
 /*
--- a/src/usb_libusb.c
+++ b/src/usb_libusb.c
@@ -145,11 +145,6 @@ static int usbdev_open(const char *port, union pinfo pinfo, union filedescriptor
             goto trynext;
           }
 
-          if(usb_set_configuration(udev, dev->config[0].bConfigurationValue)) {
-            pmsg_notice("(config %d) %s\n", dev->config[0].bConfigurationValue, usb_strerror());
-            // goto trynext;    // Let's hope it has already been configured
-          }
-
           int usehid = !!(pinfo.usbinfo.flags & PINFO_FL_USEHID);
           for(iface = 0; iface < dev->config[0].bNumInterfaces; iface++) {
             cx->usb_interface = dev->config[0].interface[iface].altsetting[0].bInterfaceNumber;
--- a/src/usb_libusb.c
+++ b/src/usb_libusb.c
@@ -67,7 +67,7 @@ static int usbdev_open(const char *port, union pinfo pinfo, union filedescriptor
   // Override/set pid, vid and/or serno from -P usb[:<vid>:<pid>][:<serno>]
   if(str_set_vid_pid_serno(port, &pinfo.usbinfo.vid, &pinfo.usbinfo.pid, serno, sizeof serno) < 0) {
     pmsg_error("invalid -P %s; drop -P option or use -P usb[:<vid>:<pid>][:<serno>]\n", port);
-    return -1;
+    return LIBAVRDUDE_EXIT_FAIL;
   }
 
   usb_init();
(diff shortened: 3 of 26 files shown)


mcuee:
Wonderful, I will for sure give this a test over the weekend.

mcuee:
Review from Claude free chatbot (Sonnet 4.6 Low).
+++++++++++++
<html>
<body>
<html><head></head><body><p>Good. Now let me fetch the actual diff / files changed:The files changed page doesn't show <code>usb_libusb.c</code> in full — let me fetch the key commit directly:Now I have the complete PR and the full new <code>usb_libusb.c</code>. Let me review it thoroughly.</p>
<p>The PR is a substantial rework across 6 files in 8 commits. Here is the full review:</p>
<hr>
<h2>PR #2128 Review</h2>
<h3>Overall architecture — well designed</h3>
<p>The central idea of extracting <code>str_set_vid_pid_serno()</code> into <code>strutil.c</code> and sharing it across <code>usb_hidapi.c</code>, <code>usb_libusb.c</code>, and <code>pickit5.c</code> is exactly the right approach. It eliminates the inconsistent per-file VID:PID parsing that was at the root of several prior issues and removes the fragile <code>vidp</code>/<code>serp</code> pointer aliasing we flagged earlier.</p>
<hr>
<h3><code>strutil.c</code> — <code>str_set_vid_pid_serno()</code> [Commit 1]</h3>
<p>The implementation is clean and well-documented. Specific observations:</p>
<p><strong>Good:</strong></p>
<ul>
<li>Uses <code>strtoul</code> with <code>endptr</code> and <code>errno</code> checking — robust against overflow and non-hex input, far better than <code>sscanf("%x")</code>.</li>
<li>Correctly validates that VID/PID are terminated by <code>:</code> or <code>\0</code>, preventing silent truncation.</li>
<li>The <code>defvid</code> mechanism (negative = disallow empty VID, non-negative = default to that value) handles the <code>usb::pid</code> syntax for pickit5 cleanly.</li>
<li>The <code>goto setserno</code> fallback correctly treats an unparseable VID:PID as a serial number, solving the hex-serial-ambiguity problem we identified.</li>
<li>Returns a numeric code (0–3) rather than a bool, giving callers precise information about what was parsed.</li>
<li>Buffer overflow on the serial number copy is handled gracefully (returns 0/2 indicating truncation rather than crashing).</li>
</ul>
<p><strong>One concern — <code>*vidp = (int) vid</code>:</strong></p>
<pre><code class="language-c">if(vidp)
  *vidp = (int) vid;
if(pidp)
  *pidp = (int) pid;
</code></pre>
<p><code>vidp</code> and <code>pidp</code> are declared <code>unsigned short *</code> in the signature, but the cast is <code>(int)</code>. This is harmless since <code>vid</code>/<code>pid</code> are already validated <code>&lt;= 0xffff</code>, but the cast to <code>int</code> before assignment to <code>unsigned short</code> is slightly misleading. <code>(unsigned short) vid</code> would be cleaner.</p>
<p><strong>Minor — return value semantics for truncated serno:</strong>
The comment says "return 0 iff serno buffer was too short" but the return expression <code>ret + !*sn</code> returns <code>ret</code> (0 or 2) plus 1 if the loop completed without truncation. If <code>n</code> ran out before all serno chars were copied, <code>*sn != 0</code> so <code>!*sn == 0</code> and the return is <code>ret + 0</code> = 0 or 2, signalling truncation. This is correct but the comment is slightly backwards — it should read "returns ret+1 if serno copied completely, ret if buffer was too short".</p>
<hr>
<h3><code>usb_hidapi.c</code> — Commit 2</h3>
<p>The old messy VID:PID parsing block (30 lines with <code>vidp</code>/<code>pidp</code>/<code>serp</code> pointer aliasing) is replaced by a single <code>str_set_vid_pid_serno()</code> call. This is a clear improvement.</p>
<p><strong>Note on <code>defvid = -1</code>:</strong> The hidapi path passes <code>defvid = -1</code> (no default VID), so <code>-P usb::pid</code> is not supported — the empty VID field causes <code>goto setserno</code> and the whole thing is treated as a serial number. This is correct behaviour for jtag3-family programmers.</p>
<p><strong><code>} else {</code> replacing <code>} else if(!dev) {</code>:</strong> The old code only opened by VID/PID if <code>dev == NULL</code> (i.e., the VID:PID branch hadn't already opened it). The new code always falls into the <code>else</code> branch when <code>*serno</code> is empty — which is the same condition since there's no prior <code>dev =</code> assignment in the new code. Correct.</p>
<hr>
<h3><code>usb_libusb.c</code> — Commits 3 and 5 [the core fix]</h3>
<p>The full new file is visible. Key changes:</p>
<p><strong>Port parsing — clean:</strong></p>
<pre><code class="language-c">if(str_set_vid_pid_serno(port, &amp;pinfo.usbinfo.vid, &amp;pinfo.usbinfo.pid, -1, serno, sizeof serno) &lt; 0) {
  pmsg_error("invalid -P %s; ...\n", port);
  return -1;
}
</code></pre>
<p>Replaces the original <code>strchr</code>-based serno-only extraction. Now VID:PID overrides work correctly here too.</p>
<p><strong><code>max_xfer = 64</code> on CMSIS-DAP detection:</strong> Present and correct.</p>
<p><strong>The bidirectional endpoint loop — the Claude-attributed line:</strong></p>
<pre><code class="language-c">// (usehid &amp;&amp; (mps == 64 || mps == 512)) suggested by Claude free chatbot (Sonnet 4.6 Low)
if((usehid &amp;&amp; (mps == 64 || mps == 512)) || mps &lt; fd-&gt;usb.max_xfer) {
  if(mps != fd-&gt;usb.max_xfer) {
    pmsg_notice("CMSIS-DAP max_xfer adjusted %d -&gt; %d owing to EP 0x%02x's wMaxPacketSize\n",
      fd-&gt;usb.max_xfer, mps, epa);
    fd-&gt;usb.max_xfer = mps;
  }
}
</code></pre>
<p>This is correct and elegantly combines both cases. Logic table:</p>

Device | Speed | mps | usehid | mps < max_xfer | condition | result
-- | -- | -- | -- | -- | -- | --
Atmel-ICE | HS | 512 | true | 512 < 64 = F | `true |  
Atmel-ICE | FS | 64 | true | 64 < 64 = F | `true |  
PICkit4/SNAP | FS | 64 | true | 64 < 64 = F | `true |  
AVRISP mkII | — | 64 | false | 64 < 64 = F | `F |  
AVRISP mkII | — | 64 | false | 64 < 912 = T | `F |  


<p>The PR is in good shape. The two items worth raising before merge: (1) the <code>pmsg_notice</code> message text for non-HID bulk EP adjustments, and (2) the <code>(int)</code> cast in <code>str_set_vid_pid_serno()</code> should be <code>(unsigned short)</code>. Everything else looks correct and the overall approach is significantly better than the earlier draft patch.</p></body></html>
</body>
</html>

mcuee:
@Youw 

Just wondering if you can use the proper Claude (or Github Copilot) to review this PR.

mcuee:
> Novelty: for the first time I've used a code fragment suggested by an AI program following @mcuee's enthusiastic use:
https://github.com/stefanrueger/avrdude/blob/68b970e2456083566cb8be1d2d935780b052bcad/src/usb_libusb.c#L207-L208
> I might delete that attribution to Claude, as the used code fragment is only a small sub-expression, and I wonder how much that differs from using a tip that Ms Google had given. It's certainly been some work trying to make sense of the verbose AI review.

@stefanrueger 

I think you can safely delete the attribution to Claude in this case.

mcuee:
> @MX682X This PR replaces pickit5.c's -P usb[:<vid>:<pid>] parsing with the more general (and stricter) parsing through str_set_vid_pid_serno(). Could you please test/try/review the changed parsing? How attached are you to the variant -P usb::<pid> where vid defaults to Microchip's vid? This element turns out to be a slight pain for the attempt to standardise -P usb:... parsing across different programmers in AVRDUDE, as it's not clear to me whether defaulting to Microchip's vid makes generally sense (I suspect, not). Dropping that otherwise neat -P usb::<pid> syntax would also make documentation easier.

My own opinion is that it is okay, the variant `-P usb::<pid> where vid defaults to Microchip's vid` can be removed and just always require vid, if that makes the code easier to maintain.

mcuee:
@stefanrueger 

My minor suggestion for usb_libusb.c (not from AI), from my libusb experiences.

```
          if(usb_set_configuration(udev, dev->config[0].bConfigurationValue)) {
            pmsg_notice("(config %d) %s\n", dev->config[0].bConfigurationValue, usb_strerror());
            // goto trynext;    // Let's hope it has already been configured
          }
```

The proper way is to skip this if the programmer has been configured. Typical programmers for avrdude do not have multiple configurations, even if they have the first configuration will be the proper one (eg: PICkit 2).

Reference:
1) https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#ga785ddea63a2b9bcb879a614ca4867bed

2) https://github.com/pyusb/pyusb/blob/master/docs/faq.rst#how-not-to-call-set_configuration-on-a-device-already-configured-with-the-selected-configuration

mcuee:
Of course I have another goal to port usb_libusb.c to use libusb-1.0 API but that will be another day.
* https://github.com/avrdudes/avrdude/issues/974

Or even to unify libusb and hidapi.
* https://github.com/avrdudes/avrdude/issues/1316

The other problem is that some programmers do not use the helper but rather use their own methods.

PICkit 2 is one major one --> need to move to hidapi with libusb fallback.
* https://github.com/avrdudes/avrdude/issues/883

src/pickit5.c
```diff
@@ -650,93 +650,47 @@ static int pickit5_open(PROGRAMMER *pgm, const char *port) {
...
+      pmsg_notice("overwriting VID:PID to %04x:%04x\n", new_vid, new_pid);
     pinfo.usbinfo.vid = new_vid;
     pinfo.usbinfo.pid = new_pid;
-    pinfo.usbinfo.flags = PINFO_FL_SILENT;
     pgm->fd.usb.max_xfer = USB_PK5_MAX_XFER;
     pgm->fd.usb.rep = USB_PK5_CMD_READ_EP;  // Command read
     pgm->fd.usb.wep = USB_PK5_CMD_WRITE_EP; // Command write
     pgm->fd.usb.eep = 0x00;
+    port = str_sprintf("usb:%s", serno); // Avoids passing on usb::<pid>:<serno>
     pgm->port = port;
```
Youw: **Memory leak (medium).** `str_sprintf()` returns `mmt_malloc()`-tracked memory, but `pgm->port` is a non-owned alias — `pgm_free()` (src/pgm.c) never frees `->port` — so this allocation leaks on every open with an explicit VID:PID (and on the error returns in this branch).

The reconstruction itself is needed, though: simply reusing the original `port` would feed `usb::<pid>:<serno>` back into `usbdev_open()`, which parses with `defvid = -1` and would fold the `<pid>` into the serial number. So keep the `usb:<serno>` string but make it teardown-managed via the existing `cache_string(str_ccprintf(...))` idiom (cf. strutil.c:1167):

```suggestion
    port = cache_string(str_ccprintf("usb:%s", serno)); // Avoids passing on usb::<pid>:<serno>
    pgm->port = port;
```

*— Claude (Claude Code)*

Youw (commented):
## Code review by Claude

Automated review requested by @Youw; the findings below are from **Claude (Opus 4.8)** via Claude Code. Please treat them as suggestions for your judgement, not blocking requests — and they are static-analysis only, so they don't replace the hardware testing you and @mcuee have lined up for #2127 / #1221.

Overall the shared `str_set_vid_pid_serno()` reads as a clean consolidation: `strtoul`+`errno` parsing is robust against overflow/garbage, the `goto setserno` fallback neatly resolves the hex-serial ambiguity, and routing `pickit5` / `usb_hidapi` / `usb_libusb` through one parser removes the old per-file `vidp`/`serp` aliasing.

One substantive item plus a few minor ones are left inline:

- **`src/pickit5.c` — memory leak (medium):** the `str_sprintf()` string stored in `pgm->port` is never freed (`pgm_free()` doesn't own `->port`).
- **`src/usb_libusb.c` — misleading message (low):** the reduce-path `pmsg_notice` now prints "CMSIS-DAP" even for non-HID devices that hit the `mps < max_xfer` branch.
- **`src/strutil.c` / `src/usb_libusb.c` — nits:** one inaccurate inline comment and two off-by-one indentations.

I did not find correctness problems in the `usb_hidapi.c` or `jtag3.c` changes (the `vidbak` save/restore there is correct and necessary, since the SNAP/PICkit4 probe loop overwrites `pinfo.usbinfo.vid`).

src/usb_libusb.c
```diff
@@ -215,20 +198,27 @@ static int usbdev_open(const char *port, union pinfo pinfo, union filedescriptor
...
+                fd->usb.max_xfer = mps;
+              }
             }
           }
-          if(pinfo.usbinfo.flags & PINFO_FL_USEHID) {
-            if(usb_control_msg(udev, 0x21, 0x0a /* SET_IDLE */ , 0, 0, NULL, 0, 100) < 0)
-              pmsg_warning("SET_IDLE failed\n");
+          if(fd->usb.max_xfer > USBDEV_MAX_XFER_3) {
+            pmsg_error("reducing too big max packet size %d to %d\n", fd->usb.max_xfer, USBDEV_MAX_XFER_3);
+           fd->usb.max_xfer = USBDEV_MAX_XFER_3;
```
Youw: Nit: this assignment is indented 11 spaces; the rest of the `if` body (the `pmsg_error` above) is at 12.

```suggestion
            fd->usb.max_xfer = USBDEV_MAX_XFER_3;
```

*— Claude (Claude Code)*

src/usb_libusb.c
```diff
@@ -215,20 +198,27 @@ static int usbdev_open(const char *port, union pinfo pinfo, union filedescriptor
...
+            int mps = dev->config[0].interface[iface].altsetting[0].endpoint[i].wMaxPacketSize,
+                epa = dev->config[0].interface[iface].altsetting[0].endpoint[i].bEndpointAddress;
+
+            if(epa != fd->usb.rep && epa != fd->usb.wep)
+              continue;
+
+            // (usehid && (mps == 64 || mps == 512)) suggested by Claude free chatbot (Sonnet 4.6 Low)
+            if((usehid && (mps == 64 || mps == 512)) || mps < fd->usb.max_xfer) {
+              if(mps != fd->usb.max_xfer) {
+                pmsg_notice("CMSIS-DAP max_xfer adjusted %d -> %d owing to EP 0x%02x's wMaxPacketSize\n",
```
Youw: Low / message-only: this branch also fires for **non-HID** devices via the `mps < fd->usb.max_xfer` disjunct on line 208 (as the pre-PR device-neutral message did). For e.g. JTAGICEmkII / AVRISP mkII / mEDBG that genuinely shrink `max_xfer`, the hard-coded "CMSIS-DAP" label is misleading. Either drop the prefix (below) or gate the wording on `usehid`:

```suggestion
                pmsg_notice("max_xfer adjusted %d -> %d owing to EP 0x%02x's wMaxPacketSize\n",
```

*— Claude (Claude Code)*

src/strutil.c
```diff
@@ -1230,6 +1230,94 @@ int str_membuf(const char *str, int type, unsigned char *buf, int size, const ch
...
+    if(*sn != ':') {
+      if(!--n)
+        break;
+      if(serno)
+        *serno++ = *sn;
+     }
+  if(serno)
+    *serno = 0;
+
+  return ret + !*sn;            // Return 0 iff serno buffer was too short
```
Youw: Two minor things in this block:

1. The closing `}` (line 1314) is indented one space too far (5 vs 4).
2. The inline comment on the `return` is inaccurate: when the buffer is too short and a valid `<vid>:<pid>:` prefix was parsed, `ret == 2`, so this returns **2**, not 0. (The doc block above already says it correctly — "0 or 2 indicating the serial number could not be completely copied".) As written, a caller testing `== 0` for truncation would miss the vid:pid case.

```suggestion
    }
  if(serno)
    *serno = 0;

  return ret + !*sn;            // ret (0 or 2): serno truncated; ret+1 (1 or 3): serno fully copied
```

*— Claude (Claude Code)*

src/pickit5.c
stefanrueger: I was aware of this (and thought I could get away with it :smile:). So am mildly impressed by the analysis, as it shows awareness of the different memory allocation methods used in AVRDUDE. I don't fully agree with wording (non-pwned alias, teardown-managed) and neither the suggestion but Claude appears to have got the gist of `str_sprintf()`, `cache_string()` and `str_ccprintf()`.

src/usb_libusb.c
stefanrueger: Good shout

src/usb_libusb.c
stefanrueger: Good catch. I should have seen this when merging the two messages.

src/strutil.c
stefanrueger: Neat analysis of the comments.

stefanrueger:
@Youw and @mcuee Thank your for the automated review and your own spiel on it. It's detected some things this PR overlooked and some I was aware of (str_sprintf() w/out mmt_free()) but thought an acceptable to solution to the necessary port string modification for the -P usb::<pid> syntax.

Having played with the PR I now think it is best to remove the otherwise clever -P usb::<pid> syntax. It saves programming headaches (see above) and holds less confusion for the user (the functions usb_libusb.c and usb_hidapi.c might print the modified port string which is not the one the user gave). Using the single parsing function actually now enables `-c pickit5` users to override vid and pid on the command line and still request serno match via `-P usb:<vid>:<pid>:<serno>`, which wasn't possible before.

MX682X:
@stefanrueger I'm not attached to the -P usb:: syntax, It was just meant as a nice to have.

I will try to test the PR tomorrow, or at least on the weekend.

stefanrueger:
This is now as good as it gets.

stefanrueger:
> I will try to test the PR tomorrow, or at least on the weekend.

@MX682X Brilliant, thank you! #2129 has a few thoughts as to testing

stefanrueger:
> The proper way is to skip this if the programmer has been configured. 

Thanks @mcuee. Do you mean deleting the code snippet? I know very little about USB...

mcuee:
> > The proper way is to skip this if the programmer has been configured.
> 
> Thanks @mcuee. Do you mean deleting the code snippet? I know very little about USB...

Yes, I think that is the way to go. The earlier code anyway already skipped unconfigured device.
https://github.com/stefanrueger/avrdude/blob/libusb-snap-etc/src/usb_libusb.c

```
          if(dev->config == NULL) {
            pmsg_warning("USB device has no configuration\n");
            goto trynext;
          }

          if(usb_set_configuration(udev, dev->config[0].bConfigurationValue)) {
            pmsg_notice("(config %d) %s\n", dev->config[0].bConfigurationValue, usb_strerror());
            // goto trynext;    // Let's hope it has already been configured
          }
```

stefanrueger:
@mcuee OK, deleting these lines
``` diff
diff --git a/src/usb_libusb.c b/src/usb_libusb.c
index af576775..737ec44f 100644
--- a/src/usb_libusb.c
+++ b/src/usb_libusb.c
@@ -145,11 +145,6 @@ static int usbdev_open(const char *port, union pinfo pinfo, union filedescriptor
             goto trynext;
           }
 
-          if(usb_set_configuration(udev, dev->config[0].bConfigurationValue)) {
-            pmsg_notice("(config %d) %s\n", dev->config[0].bConfigurationValue, usb_strerror());
-            // goto trynext;    // Let's hope it has already been configured
-          }
-
           int usehid = !!(pinfo.usbinfo.flags & PINFO_FL_USEHID);
           for(iface = 0; iface < dev->config[0].bNumInterfaces; iface++) {
             cx->usb_interface = dev->config[0].interface[iface].altsetting[0].bInterfaceNumber;
```

[edit: <s>They only ever would have produced notice messages.</s>]

mcuee:
BTW, this is one missing feature of libusb-0.1/libusb-win32 API.

libusb-1.0 API has this function: 
`int libusb_get_configuration(libusb_device_handle *  dev_handle, int * config)`
https://libusb.sourceforge.io/api-1.0/group__libusb__dev.html#gae921014b888b105471a31d54c77c1c4d

Technically it is also possible to use control request to get the info, but I remember some devices will chock this due to firmware bugs.

Reference: libusb-win32 feature request (closed as the idea is not to change the API any more but focus on the driver improvement).
* https://github.com/mcuee/libusb-win32/issues/1

stefanrueger:
@mcuee But why is it a bad idea to *set* the usb configuration?

stefanrueger:
I just changed the last commit message (no change in code) and will merge this PR to be used as release candidate for v8.2.

mcuee:
> I just changed the last commit message (no change in code) and will merge this PR to be used as release candidate for v8.2.
> 

Maybe it is a good idea to tag a RC release, say v8.2-rc1.

stefanrueger:
> Maybe it is a good idea to tag a RC release, say v8.2-rc1.

Did so, but this created a release. I don't want to put attention to this *before* any serious testing has been done, so removed the tag.

image

mcuee:
@stefanrueger 
Good, at least github actions prove that the github action scripts are working.

For example, the `deploy` and `arduino_packing_release` github actions will only run upon tagging a release.
https://github.com/avrdudes/avrdude/actions/runs/27821978881
https://github.com/avrdudes/avrdude/actions/runs/27821978886

Your screenshot shows the deploy github action works perfectly.

For our previous 8.1 release, arduino_packing_release was not working and I only fixed it post 8.1 release.
* https://github.com/avrdudes/avrdude/issues/2039
* https://github.com/avrdudes/avrdude/pull/2101

BTW, the practice from the other projects I am involved is usually to create the RC1 and then start the testing, the followed by RC2 or other RCs if needed, then to final release.

Example from libusb project.
* https://github.com/libusb/libusb/issues/1757
* https://github.com/libusb/libusb/issues/1824

stefanrueger:
> BTW, the practice from the other projects I am involved is usually to create the RC1 and then start the testing

As github advertised the RC as release, I'm just worried that some OS distributors might grab that, eg, Arch :astonished:

stefanrueger:
> Wonderful, I will for sure give this a test over the weekend.

@mcuee don't forget in all the excitement of release testing to *specifically* test whether #1221 has been dealt with. :wink:

mcuee:
@MCUdude and @stefanrueger 

One important test for avrdude 8.2 release is to test the recently merged PR #2128 which should fix #1221.
* #2128 
* #1221

I will test PICKit 4 in AVR mode under Windows.

@stefanrueger may have tested snap under Linux already.

I cannot test high speed USB HID device like Atmel ICE. @MCUdude has the PowerDebugger. I believe it is a high speed USB HID device, just like Atmel ICE.

@MCUdude  Please confirm with `lsusb -vvv` or USBProber under macOS).

The test can not easily done under macOS (need to force detach the driver using sudo), better do it under Windows or Linux.
1) Windows: Use Zadig to switch the driver from the default Windows HID driver to WinUSB, then test
2) Linux: remove hidapi and then rebuild avrdude, then test

Same for FreeBSD: remove hidapi and rebuild avrdude, then test. But I guess you two may not use FreeBSD though.

You can test PICKit 4 as well to make sure that works as well.

Under Linux using the following method should work as well -- to force detach kernel hid driver.
* https://github.com/avrdudes/avrdude/issues/1221#issuecomment-1546485691
* https://lwn.net/Articles/143397/

Or we can use Python and pyusb. Supposedly this may work under macOS as well.

```
import usb.core
import usb.util

# Locate your device using Hex IDs
dev = usb.core.find(idVendor=0x1234, idProduct=0x5678)

if dev is None:
    raise ValueError("Device not found")

interface = 0

# Check and detach the kernel driver
if dev.is_kernel_driver_active(interface):
    try:
        dev.detach_kernel_driver(interface)
        print("Kernel driver detached successfully.")
    except usb.core.USBError as e:
        print(block_txt := f"Could not detach driver: {e}")
```

mcuee:
Okay, tested by both @stefanrueger and I, this PR does not fix #1221.

MCUdude:
@mcuee here's the output of `lsusb -vvv` when the powerdebugger is connected to my mac.

<details>
<summary>lsusb -vvv</summary>

```
$ lsusb -vvv

Bus 001 Device 001: ID 03eb:2144  
Negotiated speed: High Speed (480Mbps)
Device Descriptor:
  bLength                18
  bDescriptorType         1
  bcdUSB               2.00
  bDeviceClass          239 [unknown]
  bDeviceSubClass         2 [unknown]
  bDeviceProtocol         1 
  bMaxPacketSize0        64
  idVendor           0x03eb 
  idProduct          0x2144 
  bcdDevice            1.01
  iManufacturer           1 
  iProduct                2 
  iSerial                 3 
  bNumConfigurations      1
  Configuration Descriptor:
    bLength                 9
    bDescriptorType         2
    wTotalLength       0x0082
    bNumInterfaces          4
    bConfigurationValue     1
    iConfiguration          0 
    bmAttributes         0x80
      (Bus Powered)
    MaxPower              500mA
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        0
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass         3 [unknown]
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              4 Power Debugger CMSIS-DAP
        HID Device Descriptor:
          bLength                 9
          bDescriptorType        33
          bcdHID               1.11
          bCountryCode            0 Not supported
          bNumDescriptors         1
          bDescriptorType        34 (null)
          wDescriptorLength      35
          Report Descriptors: 
            ** UNAVAILABLE **
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x01  EP 1 OUT
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval               1
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x82  EP 2 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval               1
    Interface Association:
      bLength                 8
      bDescriptorType        11
      bFirstInterface         1
      bInterfaceCount         2
      bFunctionClass          2 [unknown]
      bFunctionSubClass       2 [unknown]
      bFunctionProtocol       1 
      iFunction               6 Power Debugger Virtual COM Port
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        1
      bAlternateSetting       0
      bNumEndpoints           1
      bInterfaceClass         2 [unknown]
      bInterfaceSubClass      2 [unknown]
      bInterfaceProtocol      1 
      iInterface              0 
      CDC Header:
        bcdCDC               1.10
      CDC ACM:
        bmCapabilities       0x06
          sends break
          line coding and serial state
      CDC Union:
        bMasterInterface        1
        bSlaveInterface         2 
      CDC Call Management:
        bmCapabilities       0x03
          call management
          use DataInterface
        bDataInterface          2
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x83  EP 3 IN
        bmAttributes            3
          Transfer Type            Interrupt
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               8
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        2
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass        10 [unknown]
      bInterfaceSubClass      0 [unknown]
      bInterfaceProtocol      0 
      iInterface              0 
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x84  EP 4 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x05  EP 5 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0040  1x 64 bytes
        bInterval               0
    Interface Descriptor:
      bLength                 9
      bDescriptorType         4
      bInterfaceNumber        3
      bAlternateSetting       0
      bNumEndpoints           2
      bInterfaceClass       255 [unknown]
      bInterfaceSubClass    255 [unknown]
      bInterfaceProtocol    255 
      iInterface              5 Power Debugger Data Gateway
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x87  EP 7 IN
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval             255
      Endpoint Descriptor:
        bLength                 7
        bDescriptorType         5
        bEndpointAddress     0x06  EP 6 OUT
        bmAttributes            2
          Transfer Type            Bulk
          Synch Type               None
          Usage Type               Data
        wMaxPacketSize     0x0200  1x 512 bytes
        bInterval             255
Device Qualifier (for other device speed):
  bLength                10
  bDescriptorType         6
  bcdUSB               2.00
  bDeviceClass          239 [unknown]
  bDeviceSubClass         2 [unknown]
  bDeviceProtocol         1 
  bMaxPacketSize0        64
  bNumConfigurations      1
can't get debug descriptor: No such file or directory
Device Status:     0x0000
  (Bus Powered)

```

</details>

mcuee:
> @mcuee here's the output of `lsusb -vvv` when the powerdebugger is connected to my mac.
> 
> lsusb -vvv
> ```
> $ lsusb -vvv
> 
> Bus 001 Device 001: ID 03eb:2144  
> Negotiated speed: High Speed (480Mbps)
> Device Descriptor:
>   bLength                18
>   bDescriptorType         1
>   bcdUSB               2.00
>   bDeviceClass          239 [unknown]
>   bDeviceSubClass         2 [unknown]
>   bDeviceProtocol         1 
>   bMaxPacketSize0        64
>   idVendor           0x03eb 
>   idProduct          0x2144 
>   bcdDevice            1.01
>   iManufacturer           1 
>   iProduct                2 
>   iSerial                 3 
>   bNumConfigurations      1
>   Configuration Descriptor:
>     bLength                 9
>     bDescriptorType         2
>     wTotalLength       0x0082
>     bNumInterfaces          4
>     bConfigurationValue     1
>     iConfiguration          0 
>     bmAttributes         0x80
>       (Bus Powered)
>     MaxPower              500mA
>     Interface Descriptor:
>       bLength                 9
>       bDescriptorType         4
>       bInterfaceNumber        0
>       bAlternateSetting       0
>       bNumEndpoints           2
>       bInterfaceClass         3 [unknown]
>       bInterfaceSubClass      0 [unknown]
>       bInterfaceProtocol      0 
>       iInterface              4 Power Debugger CMSIS-DAP
>         HID Device Descriptor:
>           bLength                 9
>           bDescriptorType        33
>           bcdHID               1.11
>           bCountryCode            0 Not supported
>           bNumDescriptors         1
>           bDescriptorType        34 (null)
>           wDescriptorLength      35
>           Report Descriptors: 
>             ** UNAVAILABLE **
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x01  EP 1 OUT
>         bmAttributes            3
>           Transfer Type            Interrupt
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0200  1x 512 bytes
>         bInterval               1
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x82  EP 2 IN
>         bmAttributes            3
>           Transfer Type            Interrupt
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0200  1x 512 bytes
>         bInterval               1
>     Interface Association:
>       bLength                 8
>       bDescriptorType        11
>       bFirstInterface         1
>       bInterfaceCount         2
>       bFunctionClass          2 [unknown]
>       bFunctionSubClass       2 [unknown]
>       bFunctionProtocol       1 
>       iFunction               6 Power Debugger Virtual COM Port
>     Interface Descriptor:
>       bLength                 9
>       bDescriptorType         4
>       bInterfaceNumber        1
>       bAlternateSetting       0
>       bNumEndpoints           1
>       bInterfaceClass         2 [unknown]
>       bInterfaceSubClass      2 [unknown]
>       bInterfaceProtocol      1 
>       iInterface              0 
>       CDC Header:
>         bcdCDC               1.10
>       CDC ACM:
>         bmCapabilities       0x06
>           sends break
>           line coding and serial state
>       CDC Union:
>         bMasterInterface        1
>         bSlaveInterface         2 
>       CDC Call Management:
>         bmCapabilities       0x03
>           call management
>           use DataInterface
>         bDataInterface          2
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x83  EP 3 IN
>         bmAttributes            3
>           Transfer Type            Interrupt
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0040  1x 64 bytes
>         bInterval               8
>     Interface Descriptor:
>       bLength                 9
>       bDescriptorType         4
>       bInterfaceNumber        2
>       bAlternateSetting       0
>       bNumEndpoints           2
>       bInterfaceClass        10 [unknown]
>       bInterfaceSubClass      0 [unknown]
>       bInterfaceProtocol      0 
>       iInterface              0 
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x84  EP 4 IN
>         bmAttributes            2
>           Transfer Type            Bulk
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0040  1x 64 bytes
>         bInterval               0
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x05  EP 5 OUT
>         bmAttributes            2
>           Transfer Type            Bulk
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0040  1x 64 bytes
>         bInterval               0
>     Interface Descriptor:
>       bLength                 9
>       bDescriptorType         4
>       bInterfaceNumber        3
>       bAlternateSetting       0
>       bNumEndpoints           2
>       bInterfaceClass       255 [unknown]
>       bInterfaceSubClass    255 [unknown]
>       bInterfaceProtocol    255 
>       iInterface              5 Power Debugger Data Gateway
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x87  EP 7 IN
>         bmAttributes            2
>           Transfer Type            Bulk
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0200  1x 512 bytes
>         bInterval             255
>       Endpoint Descriptor:
>         bLength                 7
>         bDescriptorType         5
>         bEndpointAddress     0x06  EP 6 OUT
>         bmAttributes            2
>           Transfer Type            Bulk
>           Synch Type               None
>           Usage Type               Data
>         wMaxPacketSize     0x0200  1x 512 bytes
>         bInterval             255
> Device Qualifier (for other device speed):
>   bLength                10
>   bDescriptorType         6
>   bcdUSB               2.00
>   bDeviceClass          239 [unknown]
>   bDeviceSubClass         2 [unknown]
>   bDeviceProtocol         1 
>   bMaxPacketSize0        64
>   bNumConfigurations      1
> can't get debug descriptor: No such file or directory
> Device Status:     0x0000
>   (Bus Powered)
> ```

@MCUdude 

Great, that confirms Power Debugger is a high speed USB HID device just like Atmel ICE.

PR #2137, opened 2026-06-23
AVR dragon does not support TPI programming

MCUdude:
The AVR dragon user guide can be found here, and there's no mention of TPI programming anywhere.
https://ww1.microchip.com/downloads/en/devicedoc/atmel-42723-avr-dragon_userguide.pdf

Testing also shows that TPI programming does not work.


stefanrueger:
Nice one. Merged

Issue #2154, opened 2026-07-12
GUI should detect darkmode


dl8dtl:
On MacOS (maybe others, too), when the system is configured for a dark mode them, the GUI starts with a dark grey background, and black text on it, which is basically unreadable.

Hints:

https://stackoverflow.com/questions/75457687/detect-dark-application-style-theme-of-currently-used-desktop-in-qt

PySide equivalent:

```
>>> from PySide6.QtGui import *
>>> from PySide6.QtCore import *
>>> app=QGuiApplication()
>>> x=app.styleHints()
>>> x.colorScheme()
<ColorScheme.Dark: 2>
>>> z=x.colorScheme()
>>> z.value
2
>>> z.name
'Dark'
```

mcuee:
Same issue under Windows. I think I mentioned the issue before.

Best option is probably to have three choices to the user.
1) System 
2) Light mode
3) Dark mode 

But need to fix the dark mode first.

mcuee:
BTW, I have created a simple avrdude PySide 6 GUI wrapper with the help of AI (Claude Sonnet 4.6 Low, free chatbot version). It is feature incomplete now but kind of working.

https://github.com/mcuee/avrdude_gui_wrapper/blob/main/claude%2Favrdude_gui_qt.py

It has the support of System/Light/Dark mode.

PR #2156, opened 2026-07-14
Fix dark mode

dl8dtl:
These changes basically make dark mode usable / readable on MacOS.

People with other systems that offer dark mode schemes (that can be detected as such by Qt) might test that out as well. Things taken care so far:

* completely different color style for different log levels in main logging area
* style for text areas when reading signature, including "signature matches / does not match configured device"

Systems that do not explicitly provide a distinction for "dark" or "light" mode end up as "unknown" (by Qt), they will be treated the way as everything was done so far, i.e. same as "light" mode systems.


dl8dtl:
Suggested fix for #2154 

Btw. no attempt has been made to catch transitions between dark and light mode in a running application. Dark mode is only examined at startup.

PR #2157, opened 2026-07-15
Fix issue #2149

### LINKED ISSUE #2149: GUI puts fusea in odd position
`fusea` appears nowhere near the other fuses but after flash:

Image

Notwithstanding this, would it not be better to offer the mnemonic names such as
```
0/wdtcfg
1/bodcfg
2/osccfg
5/syscfg0
6/syscfg1
7/codesize
8/bootsize
a/pdicfg
```

dl8dtl:
Split the AVR-DD and AVR-DU common values into two separate config sections.,

That maintains "fusea" in the order right behind "fuse9" where it was expected to be.

Fixes #2149

### DIFF
--- a/src/config_gram.y
+++ b/src/config_gram.y
@@ -378,6 +378,10 @@ part_def :
           yywarning("revise fileio_mem_offset(), avrdude.conf entry or memory type assignment");
       }
 
+      // Sort memories in canonical order
+      if(current_part->mem)
+        lsort(current_part->mem, avr_mem_cmp);
+
       existing_part = locate_part(part_list, current_part->id);
       if(existing_part) {
         { /* temporarily set lineno to lineno of part start */
--- a/src/libavrdude.i
+++ b/src/libavrdude.i
@@ -627,6 +627,7 @@ AVRMEM * avr_locate_mem(const AVRPART *p, const char *desc);
 AVRPART * locate_part(const LISTID parts, const char *partdesc);
 AVRMEM_ALIAS * avr_locate_memalias(const AVRPART *p, const char *desc);
 AVRMEM_ALIAS * avr_find_memalias(const AVRPART *p, const AVRMEM *m_orig);
+const char *avr_mem_name(const AVRPART *p, const AVRMEM *mem);
 
 // Programming modes for parts and programmers: ensure it's copied from libavrdude.h
 #define PM_SPM                1 // Bootloaders, self-programming with SPM opcodes or NVM Controllers
--- a/src/main.c
+++ b/src/main.c
@@ -1184,11 +1184,6 @@ int main(int argc, char *argv[]) {
     }
   }
 
-  // Sort memories of all parts in canonical order
-  for(LNODEID ln1 = lfirst(part_list); ln1; ln1 = lnext(ln1))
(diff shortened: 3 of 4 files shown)


stefanrueger:
See [comment](https://github.com/avrdudes/avrdude/issues/2149#issuecomment-4988130859) on how AVRDUDE's main.c orders memory in a consistent way (no matter how they appear in `avrdude.conf`)

stefanrueger:
I thought of a better way of ensuring the right order of memories when displayed (both in terminal `part -m` and the GUI):
``` diff
diff --git a/src/config_gram.y b/src/config_gram.y
index 07024000..fe9ecbbb 100644
--- a/src/config_gram.y
+++ b/src/config_gram.y
@@ -378,6 +378,10 @@ part_def :
           yywarning("revise fileio_mem_offset(), avrdude.conf entry or memory type assignment");
       }
 
+      // Sort memories in canonical order
+      if(current_part->mem)
+        lsort(current_part->mem, avr_mem_cmp);
+
       existing_part = locate_part(part_list, current_part->id);
       if(existing_part) {
         { /* temporarily set lineno to lineno of part start */
diff --git a/src/main.c b/src/main.c
index d2a430c4..f239c129 100644
--- a/src/main.c
+++ b/src/main.c
@@ -1184,11 +1184,6 @@ int main(int argc, char *argv[]) {
     }
   }
 
-  // Sort memories of all parts in canonical order
-  for(LNODEID ln1 = lfirst(part_list); ln1; ln1 = lnext(ln1))
-    if((p = ldata(ln1))->mem)
-      lsort(p->mem, avr_mem_cmp);
-
   // Set bitclock from configuration files unless changed by command line
   if(default_bitclock > 0 && bitclock == 0.0) {
     bitclock = default_bitclock;
```

This sorts  the memories into the "right" order straight away when reading in the part definition:
 - It is the right thing to do, as the order of memories in `avrdude.conf` has no inherent meaning
 - Expecting maintainers of `avrdude.conf` to consider the order of memories is not good and hard to document
 - This patch makes sorting the memories after reading the configuration file unnecessary.
 - Above patch avoids extra work for anyone using libavrdude.
 - Above patch also  makes it unnecessary to change `avrdude.conf` rendering commit 62d5daf86bc458e241b54d6a472856f3cb2b9ac9 in this PR unnecessary

dl8dtl:
Something like that would have been my next suggestion, after you mentioned it is already arranged for in the standard `main.c` :-) So yes, I'm all for that!

dl8dtl:
To me, all this looks quite fine now. It also solves the "fusea" display for devices like the AVR-DA-S devices that add "fusea" on top of their non-S counterparts.

Displaying the more descriptive alias names instead of just "fuseN" is still left.

stefanrueger:
> Displaying the more descriptive alias names instead of just "fuseN" is still left.

Maybe use `avr_mem_name(p, mem)` as a name in the first instance. It returns sth like `flash` or `fuse7/codesize`. In future, this function could drop the mention of `fuseX` if we think this is no longer necessary.

dl8dtl:
OK, that still needed to be marked to be wrapped by SWIG, but otherwise, good idea.
screenshot-2026-07-17T233459

PR #2158, opened 2026-07-16
Fixing look up table logic for automotive ATTiny parts with pickit5

MX682X:
My suggestion for a fix for this problem: Just check if the description ends with "auto" and create a copy of the description without "auto" to map it correctly to the look up table. 
Making a copy seems to create an error on arm64 or msvc in general though. It does not like my code, so I made it a draft. Any suggestions on how to do it more elegant?
```
D:\a\avrdude\avrdude\src\pickit5.c(990,22): error C2057: expected constant expression [D:\a\avrdude\avrdude\build\src\libavrdude.vcxproj]
D:\a\avrdude\avrdude\src\pickit5.c(990,22): error C2466: cannot allocate an array of constant size 0 [D:\a\avrdude\avrdude\build\src\libavrdude.vcxproj]
D:\a\avrdude\avrdude\src\pickit5.c(990,12): error C2133: 'temp_desc': unknown size [D:\a\avrdude\avrdude\build\src\libavrdude.vcxproj]
```


stefanrueger:
> Making a copy seems to create an error on arm64 or msvc in general though.

Yes, these compilers don't like variable-length arrays at compile time. I pushed a commit that uses `strdup()` instead. Have a look and test whether the code is working.

MX682X:
Tested with `-pattiny1614auto` and `-cpickit5_updi` and a standard 1614 as actual chip. Works fine.

Also tested with removing the part from the LUT, part not found error is shown correctly.

Thanks for putting in the string helper functions. I really need to remember for the next time that those exist...

stefanrueger:
Great! I am happy with this PR and suggest merging at next mergefest.

PR #2160, opened 2026-07-19
Update usbasp.h

jayantarb22-oss:


src/usbasp.h
```diff
@@ -59,7 +59,7 @@
 #define USBASP_ISP_SCK_8      5 //   8 kHz
 #define USBASP_ISP_SCK_16     6 //  16 kHz
 #define USBASP_ISP_SCK_32     7 //  32 kHz
-#define USBASP_ISP_SCK_93_75  8 //  93.75 kHz
+#define USBASP_ISP_SCK_93_75  8 //  125 kHz
```
dl8dtl, line 62: The comment has a discrepancy to the name of the macro now. That makes no sense.

dl8dtl (changes requested):
First of all, a pull request with a completely empty description is never a good idea.

Second, see inline remark, having a code comment that does not match the name of the macro makes no sense. If the clock is really different from the claimed 93.75 kHz, either the macro needs to be renamed as well, or at the very least, the comment should explain something like "claims to be 93.75 kHz, but measured as 120 kHz" or such.

Finally, please explain *why* you think the 93.75 kHz are incorrect. They match the USBasp source code.

mcuee:
Closing this for now.

PR #2168, opened 2026-07-22
Fix `prog_enabled` clearance

stefanrueger:
`prog_enabled` should only be cleared once there is evidence that leave programming has actually worked.

See [comments](https://github.com/avrdudes/avrdude/pull/2162#issuecomment-5024736144) for details.


stefanrueger:
@dl8dtl @wexi please test

PR #2171, opened 2026-07-28
Provide `port` and `modify` configuration syntax

### LINKED ISSUE #2170: Add "port" as a token in the programmer section of the config file to supply -P content
For linuxspi programmers, currently avrdude 8.2 requires a -P argument with something like /dev/spidev0.0:/dev/gpiochip0:23.

In principle, there's no reason this can't be in the config file so that it doesn't have to be on _every_ command line. For a given setup, it's extremely unlikely to ever change.

While we're at it, for the Pi 5 / 500, it's necessary to always add "-x disable_no_cs". It'd be nice to be able to add that to the config as well.

stefanrueger:
Fixes #2170 

1. This PR provides a way to modify programmer and part descriptions using the new syntax
```
# Add my own spiel on the default port for existing linuxspi programmer
modify programmer "linuxspi"
    port                   = "/dev/spidev0.0:/dev/gpiochip0:23";
;

modify part "m328p"
    chip_erase_delay       = 100000; # Test longer delay
;
```
This is intended to be used in the per-user configuration file. The advantage of the `modify` syntax is that one does not need to create a new `id` as would be necessary with parenting off an existing instance:
```
programmer parent "linuxspi"
    id                     = "spi23";
    port                   = "/dev/spidev0.0:/dev/gpiochip0:23";
; 
```

2. This PR also adds a `port` string property to the programmer structure. The `port` of the chosen programmer supersedes the corresponding `default_serial`, `default_parallel`, `default_spi` or `default_linuxgpio` setting from `avrdude.conf`. The command-line `-P <port>` option in turn supersedes the `port` setting as it should be.

Note that this PR only deals with the avrdude command-line program, not with the avrdude-gui.


stefanrueger:
Just to be absolutely clear: `modify` is not needed to solve #2170. Instantiation of specific programmers through parenting with new id as above (as suggested by @dl8dtl in #2170)  is probably the best way forward. This PR implements `modify` as an independent syntax that I've wanted for a long time for quicker debugging.

stefanrueger:
After playing with the PR, I thought that the new syntax `port = "usb:2341:0043";` looks nicer and is more intuitive than `default_port =  "usb:2341:0043";` This PR now uses `port` instead of `default_port` in user-facing configuration fies. The corresponding PROGRAMMER component remains `default_port` for the simple reason that this C structure used to have a `port` component for internal use; the latter has now been renamed `chosen_port` as it is a copy of port string that AVRDUDE has settled on.

stefanrueger:
@nsayer Seeing that your avatar sports the FreeBSD art, could you advise on whether the `"unknown"` entries
https://github.com/avrdudes/avrdude/blob/1ecca1fdbe12c3cba6732fe99d9af13699799120/src/CMakeLists.txt#L95-L99
have better system defaults?

nsayer:
I'm ashamed to say that my avatar is... historical. I no longer have FreeBSD machines active. I'm now MacOS, Raspberry Pi OS or Ubuntu.

That said, I'm unaware of any SPI or GPIO support in FreeBSD at all. But I'm not the guy to ask at this point.

dl8dtl:
Too bad ;-)

PR #2204, opened 2026-09-07
Silence MSVC C4146 in the verify loop

### LINKED ISSUE #2203: Minor compiler warning for Windows MSVC build
From github action.
https://github.com/avrdudes/avrdude/actions/runs/33599728349/job/100150531169
```
D:\a\avrdude\avrdude\src\avr.c(1401,27): warning C4146: unary minus operator applied to unsigned type, result still unsigned [D:\a\avrdude\avrdude\build\src\swig_avrdude.vcxproj]
```

Same for my local build

swimmingbrain:
The MSVC build warns on the verify loop:

```
src\avr.c(1401,27): warning C4146: unary minus operator applied to unsigned type, result still unsigned
```

`bdiff & -bdiff` is the usual lowest-set-bit trick and `bdiff` is an `unsigned int`, so MSVC complains about the minus. `strutil.c` already works around the same warning with `~ret + 1`, so I used that spelling here too.

The value is the same for every input. I compared both versions over 0..0x1FFFF and a handful of edge cases including 0, 0x80000000 and 0xFFFFFFFF, and the loop walks the same bits in the same order.

Closes #2203

### DIFF
--- a/src/avr.c
+++ b/src/avr.c
@@ -1398,7 +1398,7 @@ int avr_verify_mem(const PROGRAMMER *pgm, const AVRPART *p, const AVRPART *v, co
       } else if((bdiff = (buf1[i] & bitmask) ^ (buf2[i] & bitmask))) {
         // Mismatch is not just in unused bits, loop over bit positions that differ
         for(unsigned int lbit; bdiff; bdiff ^= lbit) {
-           lbit = bdiff & -bdiff; // Lowest bit that differs
+           lbit = bdiff & (~bdiff + 1); // Lowest bit that differs (~x + 1 is -x without MSVC C4146)
            biterrs++;             // Number of bit mismatches
            bitsset += !!(lbit & buf1[i]); // The mismatched bit was set on device
         }


swimmingbrain:
The red armhf leg is not from this change. Its apt step fails with `Release file for http://deb.debian.org/debian-security/dists/bullseye-security/InRelease is expired`, so the job stops before anything is compiled, and the arm64 and i386 legs were cancelled by fail-fast right after. A rerun once the mirror is sane again should be green.

stefanrueger:
Thanks for the PR. Yes, this will silence MSVC C4146. (I think that is a grotesque warning, as if anyone would reasonably expect the unary minus operator to change the type of an unsigned expression; the compiler is wrong in issuing that warning).

swimmingbrain:
Thanks for the quick merge. Agreed on the warning, `-x` on an unsigned type is well defined and does exactly what the loop wants; it just happens to be the one spelling MSVC objects to.

Issue #2208, opened 2026-09-11
Deploy release with automated workflow fails


stefanrueger:
See [here](https://github.com/avrdudes/avrdude/actions/runs/34575323628) and [here](https://github.com/avrdudes/avrdude/actions/runs/34575323628/job/103192101780) for details.

I created the v8.3 release manually from the tag.

mcuee:
Error:
```
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
Run actions/create-release@v1
(node:2125) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead. CVEs are not issued for `url.parse()` vulnerabilities.
(Use `node --trace-deprecation ...` to show where the warning was created)
Error: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}
```

mcuee:
https://github.com/actions/create-release

This has been archived and seems to me no longer working. We can just delete it and use manual upload for the release binaries.

mcuee:
I just created a simple PR to remove the automatic upload of the MSVC binaries.
* https://github.com/avrdudes/avrdude/issues/2210

mcuee:
Future improvement options from Google AI Overview.

We can go with either Option 1 or Option 3. Maybe Option 3 is the easiest.

Requirement | Recommended Solution | Why
-- | -- | --
Uploading files + Creating release | softprops/action-gh-release | Combines release creation and asset uploading into one fast step.
No third-party actions | GitHub CLI (gh release) | Native tool managed directly by GitHub; no security supply-chain risks.
Minimal code changes | shogo82148/actions-create-release | Keeps the exact same inputs/outputs as the archived action.

Option 1 example
```
- name: Create GitHub Release
  uses: softprops/action-gh-release@v2
  with:
    tag_name: ${{ github.ref_name }}
    name: Release ${{ github.ref_name }}
    draft: false
    prerelease: false
    files: |
      dist/*.zip
      build/*.tar.gz
  env:
    GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

Option 2 example
```
- name: Create Release using GitHub CLI
[...]

PR #2210, opened 2026-09-12
Remove release and asset jobs from deploy.yml

mcuee:


mcuee:
This is just a work-around for Issue #2208.

In any case, the original deployment step does not do much either -- just upload MSVC binaries. I still need to carry out manually upload for other binaries.

